PR #13, opened 2024-09-20
Shader crate cargo features

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/12 to be merged first

Adds fields for default-features and features to SpirvBuilder [and] pipes those fields through to the cargo command that builds the shader crate.

[Original PR in embark's repo](https://github.com/EmbarkStudios/rust-gpu/pull/1124)

PR #12, opened 2024-09-19
updated toolchain

Firestar99:
cherry-picking some of eddyb's nightly updates

[For this one the PR's diff is fine](https://github.com/Rust-GPU/rust-gpu/pull/12/files)

PR #14, opened 2024-09-20
Subgroup intrinsics

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/12 https://github.com/Rust-GPU/rust-gpu/pull/13 to be merged first

Adds support for subgroup intrinsics, see [GL_KHR_shader_subgroup](https://github.com/KhronosGroup/GLSL/blob/main/extensions/khr/GL_KHR_shader_subgroup.txt) and [SPIRV spec 3.49.24. Non-Uniform Instructions](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_group_and_subgroup_instructions).

Note that the glsl extension only exposes spirv's `OpGroupNonUniform*` instructions, and NOT the `OpGroup*` instructions which can be found in [3.49.21. Group and Subgroup Instructions](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_group_and_subgroup_instructions). As many `OpGroup*` instructions are only documented with `Reserved.`, I assume during development they decided against these and switched to the `OpGroupNonUniform*` instructions.

Open Questions:
* How should the intrinsic functions be named? spirv calls them  `OpGroupNonUniform*`, glsl calls them `subgroup*`. I've decided to name them `subgroup_non_uniform*` but I'm wondering if the `non_uniform` should be dropped as they all are non uniform? **Resolved:** Dropped the `non_uniform`, all are now named `subgroup_*` like glsl.
* Many of the instructions take or return "a scalar or vector of floating-point type, integer type, or Boolean type", for which I've added the `trait VectorOrScalar`. But glam's boolean vectors seem to behave weirdly and not work, not even the compiletest `arch/any` uses them. **Resolved:** For now I've just left them unsupported.


LegNeato:
I thought we had a scalorOr vector trait already but looks like it never landed:

https://github.com/EmbarkStudios/rust-gpu/pull/1030/

Firestar99:
I've rebased and made the `VectorOrScalar` interface match the discussions in https://github.com/EmbarkStudios/rust-gpu/pull/1030

crates/spirv-std/src/vector.rs
```diff
@@ -1,33 +1,58 @@
...
+impl_vector! {
+    f32: glam::Vec2 => 2, glam::Vec3 => 3, glam::Vec3A => 3, glam::Vec4 => 4;
+    f64: glam::DVec2 => 2, glam::DVec3 => 3, glam::DVec4 => 4;
+    u32: glam::UVec2 => 2, glam::UVec3 => 3, glam::UVec4 => 4;
+    i32: glam::IVec2 => 2, glam::IVec3 => 3, glam::IVec4 => 4;
+}
 
 /// Trait that implements slicing of a vector into a scalar or vector of lower dimensions, by
-/// ignoring the highter dimensions
+/// ignoring the higter dimensions
```
fee1-dead:
```suggestion
/// ignoring the higher dimensions
```

Firestar99: squashed with the commit that tried to fix this

PR #20, opened 2024-09-27
Panic structurization improvements

Firestar99:
This PR improves structurization of panics and other functions that exit the invocation, like `OpKill` (`discard;` in glsl) in fragment shaders. This should hopefully reduce the amount of structurization errors in the resulting spv.

Mainly developed by @eddyb with some test fixes by @LegNeato

PR #21, opened 2024-09-27
Linker: speedup and debug info preservation

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/20 to be merged first. Contains a lot of WIP commits.

Some improvements to the linker by @eddyb:
* mem2reg speedups, which has been known to be exponentially slow
* preserves debug info when the linker does inlining

This PR contains a lot of WIP commits, but it has worked flawlessly in tests and in my Project.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2435,14 +2435,13 @@ dependencies = [
 [[package]]
 name = "spirt"
 version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "64e1f7903720ff818d6da824edf2c4082c6e7a029a99317fd10c39dd7c40c7ff"
+source = "git+https://github.com/Rust-GPU/spirt?rev=39140c7f74d4cf77412f619b3b74d3be236ee5d3#39140c7f74d4cf77412f619b3b74d3be236ee5d3"
 dependencies = [
  "arrayvec",
  "bytemuck",
  "derive_more",
  "elsa",
- "indexmap 1.9.3",
+ "indexmap 2.5.0",
  "internal-iterator",
  "itertools",
  "lazy_static",
--- a/crates/rustc_codegen_spirv/Cargo.toml
+++ b/crates/rustc_codegen_spirv/Cargo.toml
@@ -55,7 +55,9 @@ rustc_codegen_spirv-types.workspace = true
 rustc-demangle = "0.1.21"
 sanitize-filename = "0.4"
 smallvec = { version = "1.6.1", features = ["union"] }
-spirt = "0.3.0"
+# FIXME(eddyb) release 0.4.0
+# spirt = "0.4.0"
+spirt = { version = "0.3.0", git = "https://github.com/Rust-GPU/spirt", rev = "39140c7f74d4cf77412f619b3b74d3be236ee5d3" }
 spirv-tools.workspace = true
 lazy_static = "1.4.0"
 itertools = "0.10.5"
--- a/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
@@ -462,6 +462,11 @@ impl CodegenArgs {
             );
             opts.optflag(
                 "",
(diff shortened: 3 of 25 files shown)


schell:
@Firestar99  I'm trying this out on `renderling`...

schell:
I ran into this compilation error with `ahash`: https://github.com/tkaitchuck/aHash/issues/200

schell:
Running `cargo update -p ahash` seems to have fixed that error.

schell:
I'm getting a lot of one error (over and over) when building `crabslab`:

```
       from `*f32`
         to `*f32x4`
   --> /Users/schell/.cargo/registry/src/index.crates.io-6f17d22bba15001f/crabslab-0.5.1/src/impl_slab_item/mod.rs:37:25
    |
37  |         let mut array = [T::default(); N];
    |                         ^^^^^^^^^^^^^^^^^
    |
```

The code is pretty innocuous, and compiled on the previous nightly:
```rust
impl<T: SlabItem + Copy + Default, const N: usize> SlabItem for [T; N] {
    const SLAB_SIZE: usize = { <T as SlabItem>::SLAB_SIZE * N };

    fn read_slab(index: usize, slab: &[u32]) -> Self {
        let mut array = [T::default(); N];
        for i in 0..N {
            let j = index + i * T::SLAB_SIZE;
            let t = T::read_slab(j, slab);
            let a: &mut T = crate::array_index_mut(&mut array, i);
            *a = t;
        }
        array
    }

    fn write_slab(&self, mut index: usize, slab: &mut [u32]) -> usize {
        for i in 0..N {
            let n = crate::slice_index(self, i);
            index = n.write_slab(index, slab);
        }
        index
    }
}
```

schell:
Now I guess the question is - is this related to changes in `rust-gpu` or is this something in the standard library or something in `rustc`?

schell:
Adding `[T; N]: Default` to the `impl` and then changing line 37 to this: 
```rust
        let mut array: [T; N] = Default::default();
```
got it compiling. After that everything seems to work as expected in my project. 

But I think `[T::default(); N]` _should work_ and it should probably be a synonym for `Default::<[T; N]>::default()`, do you agree?

schell:
The great news is that my shaders compile _much_ faster on this branch. 5 seconds vs the previous 40 seconds! An ~87% reduction - that's pretty epic!

schell:
> If [T::default(); N] works the same as vec! does...

I don't think it would, as `vec!` is a macro but `[T::default(); N]` would get parsed as is.

> Could you check your array_index_mut()? Can it potentially panic, cause that has been my main trouble maker in rust-gpu in the past (don't know about this new branch though). Maybe try spirv_std::arch::IndexUnchecked::index_unchecked?

`array_index_mut` is indeed using `index_unchecked_mut`, see [here](https://github.com/schell/crabslab/blob/d8a03b587fcbb810e4ea5b6a0eb8f2d573832b49/crates/crabslab/src/lib.rs#L80). 

I'm not going to make a fuss, but something changed, and I think we **should** be able to initialize arrays this way, but it doesn't block me, and I've already released a new `crabslab`.

LegNeato:
Minimal repro of what I think is the same issue:

```rust
use spirv_std::spirv;

#[spirv(fragment)]
pub fn main() {
    let x = [[1; 2]; 1];
}
```

LegNeato:
Lame, I screwed up the merge because I used github's UI.

Firestar99:
Rebased it for you :D

LegNeato:
@Firestar99 do you think these WIP commits are landable? I actually started working on a different speedup for this section of the code before I remembered about this PR ha.

LegNeato:
They _look_ fine to me, and there are some tests, and it is fairly self-contained...but 🤷

Firestar99:
I've been using the non-rebased version for the entirety of my masters and have not encountered a single issue. That's all the guarantees I can give you, but I'd still say this is mergeable.

LegNeato:
Sweet, let's do it!

LegNeato (approved):
yolo

PR #30, opened 2024-10-11
feat: added image access qualifier, fixes #28

### LINKED ISSUE #28: Add `AccessQualifier` to `Image`
I'm [working on occlusion culling](https://renderling.xyz/devlog/index.html#thu_oct_10_2024) and can't seem to write to a texture in my compute shader. `wgpu` wants the storage texture to be write-only, but there's no way to express that from my `rust-gpu` shader.  

Specifically we need to add [`AccessQualifier`](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Access_Qualifier) to the `Image` type variables.

Related to: 

* https://github.com/EmbarkStudios/rust-gpu/pull/1116
* https://github.com/EmbarkStudios/rust-gpu/issues/1097

And in one of those issues @eddyb says: 

> That is, this is an OpenCL feature, not a Vulkan one, and wgpu couldn't possibly make use of it (unless they misunderstood the specification, too?).

But it looks like this **is** used in Vulkan, at least in compute. And I'm hitting it in my use of `wgpu`. 

I expect that I can try to tackle this, unless someone gets to it first.
 
### Expected Behaviour
I expect to be able to specify that an `Image` is read-only, write-only, or read and write.

Create a shader that uses a texture that gets written to. Try to use that shader in a `wgpu` compute pipeline. 

My specific pipeline validation error is this: 

```
wgpu error: Validation Error

Caused by:
  In Device::create_compute_pipeline, label = 'compute-occlusion-copy-depth-to-pyramid'
    Error matching shader requirements against the pipeline
      Shader global ResourceBinding { group: 0, binding: 2 } is not available in the pipeline layout
        Texture class Storage { format: R32Float, access: StorageAccess(STORE) } doesn't match the shader Storage { format: R32Float, access: StorageAccess(LOAD | STORE) }
```

But yours would be different.

I should also add that simply marking my storage texture as `LOAD | STORE` (read+write) from `wgpu` is not an option for me as I am targeting WebGPU, and read+write storage textures is a native-only opt-in feature ;).

## Shader 

```rust 
pub type DepthImage2d = Image!(2D, type=f32, sampled=true, depth=true);
pub type DepthPyramidImage = Image!(2D, format = r32f, sampled = true, depth = false);
pub type DepthPyramidImageMut = Image!(2D, format = r32f, sampled = false, depth = false); // this needs to be write-only

/// Copies a depth texture to the top mip of a pyramid of mips.
#[spirv(compute(threads(32)))]
pub fn compute_copy_depth_to_pyramid(
    #[spirv(descriptor_set = 0, binding = 0, storage_buffer)] desc: &PyramidDescriptor,
    #[spirv(descriptor_set = 0, binding = 1)] depth_texture: &DepthImage2d,
    #[spirv(descriptor_set = 0, binding = 2)] mip: &DepthPyramidImageMut,
    #[spirv(global_invocation_id)] global_id: UVec3,
) {
    if desc.should_skip_invocation(global_id) {
        return;
    }

    let depth: Vec4 = depth_texture.fetch_with(global_id.xy(), sample_with::lod(0));
    unsafe {
        mip.write(global_id.xy(), depth);
    }
}
```

schell:
This adds the ability to specify the image access qualifier from the `Image!` macro, or from the `Image` type variables themselves. 

Fixes #28 .

### DIFF
--- a/crates/spirv-std/macros/src/image.rs
+++ b/crates/spirv-std/macros/src/image.rs
@@ -413,7 +413,7 @@ impl quote::ToTokens for ImageType {
         let sampled = params::sampled_to_tokens(&self.sampled);
         let sampled_type = &self.sampled_type;
         let components = self.components;
-        let access = params::image_access_to_tokens(self.access);
+        let access = params::image_access_to_const_u32(self.access);
 
         tokens.append_all(quote::quote! {
             #crate_root::image::Image<
@@ -425,7 +425,7 @@ impl quote::ToTokens for ImageType {
                 { #crate_root::image::#sampled as u32 },
                 { #crate_root::image::#format as u32 },
                 { #components as u32 },
-                #crate_root::image::__private::#access
+                #access
             >
         });
     }
@@ -573,12 +573,12 @@ mod params {
         }
     }
 
-    pub fn image_access_to_tokens(access: Option<AccessQualifier>) -> proc_macro2::TokenStream {
+    pub fn image_access_to_const_u32(access: Option<AccessQualifier>) -> proc_macro2::TokenStream {
         match access {
-            Some(AccessQualifier::ReadOnly) => quote!(ImageAccessReadOnly),
-            Some(AccessQualifier::WriteOnly) => quote!(ImageAccessWriteOnly),
-            Some(AccessQualifier::ReadWrite) => quote!(ImageAccessReadWrite),
-            None => quote!(ImageAccessUnknown),
+            Some(AccessQualifier::ReadOnly) => quote!(AccessQualifier::ReadOnly as u32),
+            Some(AccessQualifier::WriteOnly) => quote!(AccessQualifier::WriteOnly as u32),
+            Some(AccessQualifier::ReadWrite) => quote!(AccessQualifier::ReadWrite as u32),
+            None => quote!(3),
         }
     }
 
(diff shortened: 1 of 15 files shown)


schell:
I just have a few tests to fix up.

crates/spirv-std/macros/src/image.rs
```diff
@@ -420,6 +435,20 @@ mod params {
     use super::*;
     use proc_macro2::TokenStream;
 
+    pub fn image_access_from_str(s: &str) -> Result<AccessQualifier, &'static str> {
```
LegNeato, line 438: Any reason to not use TryFrom instead?

LegNeato:
I'm on my phone so I may have missed it, but I can't find where a test actually uses it?

schell:
@LegNeato thanks for looking. After running with this in my pipeline, it doesn't fix the problem I was trying to solve. I think I understand now what @eddyb was saying when he said that `AccessQualifier` is Kernel only, and doesn't apply to shaders. 

So, I'm not sure these changes are warranted.

crates/spirv-std/macros/src/image.rs
schell: Just that this is a smaller diff and it follows the style that's already in the module.

PR #32, opened 2024-10-12
make subgroup ops match glsl more closely

Firestar99:
I wasn't a big fan of how I exposed subgroup ops that require a value from the `GroupOperation` enum. This PR changes all of those functions to be much more like glsl exposes them. 

These operations:
```rust
spirv_std::arch::subgroup_i_add::<{ GroupOperation::Reduce as u32 }, _>(value)
spirv_std::arch::subgroup_i_add::<{ GroupOperation::InclusiveScan as u32 }, _>(value)
spirv_std::arch::subgroup_i_add::<{ GroupOperation::ExclusiveScan as u32 }, _>(value)
spirv_std::arch::subgroup_i_add_clustered::<8, _>(value)
```

Got changed to this:
```rust
spirv_std::arch::subgroup_i_add(value)
spirv_std::arch::subgroup_inclusive_i_add(value)
spirv_std::arch::subgroup_exclusive_i_add(value)
spirv_std::arch::subgroup_clustered_i_add::<8, _>(value)
```
The new function names are *almost* exactly the same as glsl, except that they have the extra `i` suffix in their name to say that they operate on integers (or `f` floats, `u` unsigned, `s` signed). If anyone got an idea how to unify that, please let me know :D
(Note that the asm instruction must be a string literal, so eg. you can't use trait constants to implement it)

### DIFF
--- a/crates/spirv-std/src/arch/subgroup.rs
+++ b/crates/spirv-std/src/arch/subgroup.rs
@@ -30,8 +30,14 @@ pub enum GroupOperation {
     /// A binary operation with an identity I and n (where n is the size of the workgroup)
     /// elements[a0, a1, … an-1] resulting in [I, a0, (a0 op a1), … (a0 op a1 op … op an-2)].
     ExclusiveScan = 2,
-    // /// See [`GROUP_OPERATION_CLUSTERED_REDUCE`]
-    // ClusteredReduce = 3,
+    /// The [`GroupOperation`] `ClusteredReduce`.
+    ///
+    /// All instructions with a [`GroupOperation`] require an additional `ClusterSize` parameter when [`GroupOperation`] is
+    /// `ClusteredReduce`. To map this requirement into rust, all function have a base version accepting [`GroupOperation`]
+    /// as a const generic, and a `_clustered` variant that is fixed to `ClusteredReduce` and takes the additional
+    /// `ClusterSize` parameter as a const generic. To not accidentally use a `ClusteredReduce` in the base variant of the
+    /// function, it was removed from the [`GroupOperation`] enum and instead resides individually.
+    ClusteredReduce = 3,
     /// Reserved.
     ///
     /// Requires Capability `GroupNonUniformPartitionedNV`.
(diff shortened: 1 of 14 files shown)


LegNeato:
Is there a way with traits we can flip it? Like adding extension traits:

```rust
    use spirv_std::arch::SubgroupOperations as _;

    let float_result = float_val.subgroup_add(float_val);
    let float_inclusive = float_val.inclusive_subgroup_add(float_val);
    let float_exclusive = float_val.exclusive_subgroup_add(float_val);
 ```

 or something like:

```rust
    let float_result = <f32 as SubgroupAdd>::add(float_val);
    let float_inclusive = <f32 as SubgroupAdd>::inclusive_add(float_val);
    let float_exclusive = <f32 as SubgroupAdd>::exclusive_add(float_val);
```

Not sure if those are more ergonomic, but I think they can be implemented?

Firestar99:
I'll be honest, I'm not a fan of it ergonomic-wise. Also current atomics are named simiarly, eg. `atomic_i_add`, though they are `i_add` whereas this PR uses `add_i`.

Here's also a small example of shader code using the renamed `subgroup_ballot_exclusive_bit_count()`, unfortunately a ballot op as I have not yet used an arithmetic op anywhere. It's atomically allocating space in a buffer (`atomic_counter`) to write elements in a compacted manner, so the resulting buffer has one element after the next without some holes in between.
```rust
pub fn subgroup_write_non_uniform<const WG: usize>(&mut self, t: T) -> bool {
	let index = unsafe {
		let ballot = subgroup_ballot(true);
		let count = subgroup_ballot_bit_count(ballot);
		let base_index = if subgroup_elect() {
			atomic_i_add::<_, { Scope::Device as u32 }, { Semantics::NONE.bits() }>(self.atomic_counter, count)
		} else {
			0
		};
		let base_index = subgroup_broadcast_first(base_index);
		let inv_index = subgroup_ballot_exclusive_bit_count(ballot);
		base_index + inv_index
	};
	// [...]
}
```

LegNeato:
I agree that is probably better, but bare methods still don't feel very idiomatic to me coming from rust with very little graphics background.

Can we do something like a Subgroup struct you can initialize and then call functions on it? I'm thinking how Mutex gives you a guard or how threads can be scoped or not scoped and then joined or how tokio gives you a handle you can later poke at. Subroups are a statedul "thing", so it feels weird to me to call bare functions to manipulate the thing rather than having the handle to the thing.

Firestar99:
> Subroups are a stateful "thing"

Subgroup intrinsics are *not* stateful, they are just plain functions that do special stuff. And in theory come with a control barrier across the subgroup, though in practice that barrier just disappears cause invocations within a subgroup should always progress at the same rate (at least on AMD and most other archs, not sure what the new Nvidia cards do). 

The only state there is is `SubgroupMask` used in subgroup_ballot operations, where each bit represents an invocation in the subgroup (the `ballot` variable in the example). And that's just a fancy wrapper of an UVec4 I added for convenience, as the spec only speaks of UVec4.
```rust
#[derive(Copy, Clone, Default, Eq, PartialEq)]
pub struct SubgroupMask(pub glam::UVec4);
```

LegNeato:
Subgroups are stateful during a single shader invocation but lose this state once the invocation ends. So for the code running they are stateful, which is what I meant.

LegNeato:
If we have stateful APIs and `must_use` we can eliminate some correctness / UB issues and footguns:

"Subgroup operations depend on the correct state of invocations... incorrect mask values can result in undefined behavior or synchronization failure" 

"Ignoring the result of subgroup operations can lead to invalid or non-deterministic outcomes, especially in communication primitives like broadcast."

LegNeato (approved):
I think we can merge this as-is and do some more design in the future.

PR #34, opened 2024-10-13
add vulkan indirect command structs

Firestar99:


crates/spirv-std/Cargo.toml
```diff
@@ -25,3 +26,4 @@ glam = { workspace = true, default-features = true }
 
 [features]
 default = []
+bytemuck = ["dep:bytemuck", "glam/bytemuck"]
```
Firestar99, line 29: I've added bytemuck as an optional dependency to mark all the `*IndirectCommand` structs as Pod, as most graphics APIs will require this. Should we make the bytemuck dependency required, eg. not hidden behind a feature?

schell: I think keeping `bytemuck` optional is good. It's true that most graphics projects will have it somewhere in the tree, but I don't think we should be deciding how folks marshal data to the GPU.

schell (approved):

LegNeato (approved):
Seems reasonable to me!

PR #36, opened 2024-10-14
Update to `nightly-2024.06.20`

LegNeato:
This **should not be merged** as the `alloca` changes are not correct. I'm putting it up just to save future effort of forward porting.


LegNeato:
This is obsoleted by https://github.com/Rust-GPU/rust-gpu/pull/170

PR #38, opened 2024-10-15
Support-dependent timestamping in wpgu-runner compute example 

ArrEssJay:
Gracefully handle absence of adapter timestamping support in wpgu-runner.
Check for adapter support at runtime and only collect timing data if it does.


LegNeato (approved):
Looks great, thank you!

PR #43, opened 2024-10-26
update rspirv 0.11 to 0.12

Firestar99:
required for mesh shaders

## changelog
```
- [PR#43](https://github.com/Rust-GPU/rust-gpu/pull/43) update rspirv to 0.12
    - added support for `VK_KHR_fragment_shader_barycentric` with `bary_coord`/`bary_coord_no_persp` built-ins
```

PR #44, opened 2024-10-26
Mesh shaders

Firestar99:
# Depends on https://github.com/Rust-GPU/rust-gpu/pull/43

Diff without dependent PR: https://github.com/Firestar99/rust-gpu-new/compare/update-rspirv...Firestar99:rust-gpu-new:mesh-shaders

changelog:
```
- [PR#44](https://github.com/Rust-GPU/rust-gpu/pull/44) added support for mesh and task shaders
```


LegNeato:
It would be good to have a test, changelog entry, and an example for this. Might be good to do a blog post too!

Firestar99:
Plenty of tests already exist in this PR and I added a changelog entry in the description. I can make you a quick example shader later.

Firestar99:
There are two vendor-specific issues I know of with regards to "payloads" (sending data from task to mesh shader):
* AMD on Windows assumes the payload is always a struct, even if the spec says it could be anything like an uint. It not being a struct will segfault the driver on pipeline creation. If you define a struct with just a single primitive member, like a uint, rust-gpu will "inline" the struct and make the payload be a uint. Workaround: add a second dummy member.
* RenderDoc suffers from the same issue as above, and will display something like "shader likely compiled by some old weird dxc version"
* RenderDoc assumes the struct for the payload is never constructed, only it's fields written to. For it's debugging, it will inject a new field into said struct to emit some debug info. If you just assign fields to that struct it's fine, but it won't fixup the `OpConstruct` statements to make a new struct, resulting in invalid spirv and RenderDoc complaining. Rust-gpu merges structs with the same members, making it rather easy to trip onto this one. 

Suggested fix: Wrap all task shader payloads in a separate struct that is not optimized away.

LegNeato:
> Plenty of tests already exist in this PR and I added a changelog entry in the description. I can make you a quick example shader later.

Oh weird, they didn't show up on mobile!

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -945,6 +953,182 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::SubgroupAvcSicGetInterRawSadsINTEL => {
             reserved!(SPV_INTEL_device_side_avc_motion_estimation);
         }
+        // SPV_EXT_mesh_shader
+        Op::EmitMeshTasksEXT | Op::SetMeshOutputsEXT => {
+            // NOTE(eddyb) we actually use these despite not being in the standard yet.
+            // reserved!(SPV_EXT_mesh_shader)
```
LegNeato: Should this be uncommented?

Firestar99: Removed the comment and code. Previously, that was `reserved!()` aka. not implemented and now it should just do nothing when validating type constraints on these intrinsics, as they're already validated by the functions representing these intrinsics (and so do most other intrinsics as well).

Firestar99: And rebased.

LegNeato:
@eddyb or @schell , I don't know anything about mesh shaders...can either of you review?

Firestar99:
I think nobody but me knows about mesh shaders...
I wouldn't worry too much about correctness of the mesh shader stuff, I've tested it plenty so hopefully all issues have already been caught.

LegNeato:
Let's land this and iterate if needed, it has been sitting too long.

LegNeato (approved):
:shipit:

Issue #40, opened 2024-10-21
CI should check more than just `--release`


LegNeato:
Due to an underlying rustc change when forward porting, building without `--release` was broken on aarch64 (see https://github.com/Rust-GPU/rust-gpu/issues/29 and https://github.com/Rust-GPU/rust-gpu/issues/31). Unfortunately, CI only runs `--release` so we were blind to this breakage.

While running everything with and without `--release` is probably not worth the additional CI time, we should have at least one job on each platform that ensures both configs work at a basic level.

LegNeato:
Fixed by https://github.com/Rust-GPU/rust-gpu/pull/52, thanks @zanciks ! 🥳

Issue #50, opened 2024-11-08
Update to a newer nightly


LegNeato:
We need to update to as new as nightly as possible. At the very least, we need to update past a nightly that fixes https://github.com/Rust-GPU/rust-gpu/issues/29. Unfortunately, there are some hard breaks from `rustc` in the update that we'll have to do substantial work to port past. Specifically, these untyped alloc changes: https://github.com/rust-lang/rust/commit/f4426c189f2587fc5e8f48bf518cc75a37d93d0f

LegNeato:
@eddyb is already working on this but I wanted to get a task know so people know it is coming 🚀

LegNeato:
When this lands, we should turn on Mac dev builds (see https://github.com/Rust-GPU/rust-gpu/pull/52)

eddyb:
Current status: `nightly-2024-11-22` (~1.84) achieved by:
- https://github.com/Rust-GPU/rust-gpu/pull/170

At the cost of having to eventually deal with those hacks:
- https://github.com/Rust-GPU/rust-gpu/issues/182

Issue #29, opened 2024-10-10
Segfault while compiling `rustc_codegen_spirv`


schell:
### Expected Behaviour
`cargo build` should finish without error.

### Example & Steps To Reproduce
Run `cargo build` after a clean clone of the repo.

## System Info

 - Rust: rustc 1.79.0-nightly (244da22fa 2024-04-23)
 - OS: macOS 15.0 (24A335)
 - GPU: integrated

## Output

```
error: could not compile `rustc_codegen_spirv` (lib)

Caused by:
[...]

PR #52, opened 2024-11-10
Added non release (dev) shader build

zanciks:
Hey all! First time contributor (to any open-source project!!!!)

I saw issue #40 and decided to try and see if I could get an action added. I've pretty much just added another example (also runs example-runner-wgpu-builder) but left out --release. 

I was a bit unsure about the OUT_DIR environment variable, and if it was necessary (not even sure why it's there for the existing example run?) but I assumed it should remain for consistency. 

On top of that, I was going to look at adding a dev profile build for android, but it seems that cargo-apk is outdated, and the maintainers of the project recommend [xbuild](https://github.com/rust-mobile/xbuild), though I am unsure if profiles (--release) are forwarded to cargo or not. Will have to play around with it.

I'm also unsure if I'm able to run actions on my fork of this repo for before submitting this, so its possible that it will return an error, but at first glance it seems fine. 

Thanks!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -79,6 +79,12 @@ jobs:
           OUT_DIR: "target/tmp"
         run: cargo run -p example-runner-wgpu-builder --release --no-default-features --features "use-installed-tools"
 
+      - name: build example shaders dev
+        if: ${{ matrix.target != 'aarch64-linux-android' }}
+        env:
+          OUT_DIR: "target/tmp"
+        run: cargo run -p example-runner-wgpu-builder --no-default-features --features "use-installed-tools"
+
       - name: Build WGPU Example for Android
         if: ${{ matrix.target == 'aarch64-linux-android' }}
         shell: bash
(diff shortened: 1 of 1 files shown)


LegNeato:
Awesome, thank you so much for the change and welcome to open source contributing! 🍻 Your change will help catch issues in the future, saving many people time and effort 🎉.

I have made it so CI will now run on this PR, let's see how it goes...the change looks great to me.

LegNeato:
Ah yes, silly me. #40 was created because of https://github.com/Rust-GPU/rust-gpu/issues/29. So, running this in CI now (correctly) exposes #40 😆 .

So we have 2 paths forward:

1. We can update this to run this on mac for now and file a task to enable it when https://github.com/Rust-GPU/rust-gpu/issues/50 is fixed. Running it still has value as it may catch issues on windows / linux.
2. Keep this in limbo until https://github.com/Rust-GPU/rust-gpu/issues/50 is fixed and then land it shortly after.

@zanciks , your call!

zanciks:
Yeah, I think we can go on with the first route! I'll add a commit here in a moment to block the test on both aarch64-linux-android, and now x86_64-apple-darwin!

LegNeato (approved):
Thanks again!

PR #53, opened 2024-11-11
linker: also dump SPIR-T on panic, not just during successful compilation.

eddyb:
Originally opened at:
- https://github.com/EmbarkStudios/rust-gpu/pull/1088

I had kept putting off landing this because ideally there would be fractal panic catching and automatic retries, but the existing improvements (wrt debugging e.g. panics originating in SPIR-T passes) are worth landing already.


LegNeato (changes requested):
Lint is failing

LegNeato (approved):
I'll just land a fix to lint, for some reason I can't push to the PR's brnach

LegNeato:
Lint fix is in https://github.com/Rust-GPU/rust-gpu/pull/54

Issue #22, opened 2024-10-01
Can no longer package with nix


abel465:
I cant keep rust-gpu updated for my projects since [this commit](https://github.com/Rust-GPU/rust-gpu/commit/775b5d5fa4a8ac7417c5d6abc9176522c000f88c)

### Example & Steps To Reproduce
I have made a [minimal example to demonstrate](https://github.com/abel465/rust-gpu-nixos-example)

```
nix run github:abel465/rust-gpu-nixos-example/b117267d254310df77e69debde9e510ed193ff56
```
The shader builds just fine for the above command. it uses [this version of rust-gpu](https://github.com/Rust-GPU/rust-gpu/commit/40456ae2f59055680befb903b2aa27ed0bba920e)
```
> nix run github:abel465/rust-gpu-nixos-example/6c9e2ca63da09ab1d119a06a596cd41f84255876
cargo:rerun-if-env-changed=RUSTGPU_CODEGEN_ARGS
cargo:rerun-if-env-changed=RUSTGPU_RUSTFLAGS
error: target path "/build/cargo-vendor-dir/spirv-builder-0.9.0/target-specs/spirv-unknown-vulkan1.1.json" is not a valid file

Caused by:
  No such file or directory (os error 2)
thread 'main' panicked at runner/src/main.rs:17:42:
called `Result::unwrap()` on an `Err` value: BuildFailed
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
The above command is the same except it uses the [next commit of rust-gpu](https://github.com/Rust-GPU/rust-gpu/commit/775b5d5fa4a8ac7417c5d6abc9176522c000f88c) which seems to hard code CARGO_MANIFEST_DIR into the binary. This is problematic because the build environment is no longer available during runtime with nix

## System Info
 - Rust: 1.78.0-nightly (3b1717c05 2024-03-10)

schell:
From what I can tell, this is hooking into the nightly-only feature of target specification [explained in this RFC](https://rust-lang.github.io/rfcs/0131-target-specification.html). Here is the relevant part: 

> Rather than hardcoding a specific set of behaviors per-target, with no recourse for escaping them, the compiler would also use this mechanism when deciding how to build for a given target. The process would look like:
>
[...]

PR #55, opened 2024-11-11
feat: spirv-builder takes optional path to target spec JSON file

schell:
This adds an optional field to `SpirvBuilder` that will set the "target specification" file. This will enable [cargo-gpu](https://github.com/rust-gpu/cargo-gpu) to point to these files, which can be stored in its cache directory.

## Relations 
* https://github.com/Rust-GPU/cargo-gpu/issues/10
* https://github.com/Rust-GPU/rust-gpu/issues/22
* https://github.com/EmbarkStudios/rust-gpu/issues/1137


LegNeato (changes requested):
rustfmt

schell:
Ah, yes - I forgot that since we can only build on macOS with `--release`, my editor's LSP doesn't work on this project...

LegNeato (approved):

PR #158, opened 2024-11-25
Update .cargo/config section of writing-shader-crates.md

### LINKED ISSUE #157: Improve .cargo/config section in the Writing Shader Crates documentation
I wasn't able to get the instructions in the [.cargo/config section](https://rust-gpu.github.io/rust-gpu/book/writing-shader-crates.html#using-cargoconfig) of the book to work.

Here are the changes I had to make.
- Use `target = <path to target spec in spirv-builder>` instead of `target = <target triple>`.
- Copy `rust-toolchain` from `rust-gpu` into my repo to be able to load the codegen dylib.

Also, the `.cargo/config` name is deprecated, the new name is `.cargo/config.toml`.

I'd be happy to make a PR with these fixes to the docs, if they are indeed correct. But I'm not sure if anyone is still using the method (I only did it to learn how the pieces fit together). So, if `spirv-builder` and/or `cargo-gpu` are the main recommended approaches, another option could be to just remove the section?

mbernat:
Resolves #157 

Improves the .cargo/config section because the instruction weren't quite correct.

### DIFF
--- a/docs/src/writing-shader-crates.md
+++ b/docs/src/writing-shader-crates.md
@@ -95,13 +95,13 @@ which may slow down full rebuilds a bit. Please read [this
 issue](https://github.com/EmbarkStudios/rust-gpu/issues/448) for more
 information, there's a few important caveats to know about this.
 
-### Using `.cargo/config`
+### Using `.cargo/config.toml`
 
 > **Note** This method will require manually rebuilding `rust-gpu` each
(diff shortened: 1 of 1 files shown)


LegNeato (approved):

LegNeato:
Thanks so much! 🍻🍻

PR #166, opened 2024-12-02
(bufix) `debug_printf` macros don't escape '{' and '}' correctly.

FractalFir:
The `debug_printf` and `debug_printfln` passes its format string to the `asm!` macro, which uses `{` and `}` to separate its arguments. Passing `{` and `}` to `asm!` directly is incorrect, and leads to compilation issues:
```
error: invalid asm template string: expected `'}'`, found `'"'`
     --> examples/shaders/sky-shader/src/lib.rs:13016:2
      |
13015 | unsafe{debug_printf!("Variant3{")};
      |        -------------------------- in this macro invocation
13016 |         unsafe{debug_printf!("fld0:")};
      | ----^ expected `'}'` in asm template string
      | |
      | because of this opening brace
      |
      = note: if you intended to print `{`, you can escape it using `{{`
      = note: this error originates in the macro `debug_printf` (in Nightly builds, run with -Z macro-backtrace for more info)
``` 

This commit escapes those characters using `{{` and `}}`, which removes this issue and produces correct behaviour.

### DIFF
--- a/docs/src/writing-shader-crates.md
+++ b/docs/src/writing-shader-crates.md
@@ -95,13 +95,13 @@ which may slow down full rebuilds a bit. Please read [this
 issue](https://github.com/EmbarkStudios/rust-gpu/issues/448) for more
 information, there's a few important caveats to know about this.
 
-### Using `.cargo/config`
+### Using `.cargo/config.toml`
 
 > **Note** This method will require manually rebuilding `rust-gpu` each
   time there has been changes to the repository.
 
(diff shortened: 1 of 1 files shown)


LegNeato (approved):
Thanks! I was just looking at this code, likely needs a major revamp in the future.

LegNeato:
The commit is over in https://github.com/Rust-GPU/rust-gpu/pull/167 now and I'll land it. Not sure what screwed up!

PR #171, opened 2024-12-08
Fix zero-sized Array GEPs

### LINKED ISSUE #151: GEP not implemented for type struct [_; 0]
### Expected Behaviour

This compiles fine with rustc (I guess this bit is probably eliminated?).

I was using const generic parameters to determine the length of some arrays. At some point, a base case ended up declaring a zero-length array, like so:

```rust
#![cfg_attr(target_arch = "spirv", no_std)]
use spirv_std::spirv;

fn example<const LENGTH: usize>() {
    let mut array = [0; LENGTH];
    for i in 0..array.len() {
        array[i] += i;
    }
}

#[spirv(compute(threads(1, 1, 1)))]
pub fn compute() {
    example::<0>();
}
```

Resulting in:

```
error: GEP not implemented for type struct [usize; 0] {  }
 --> examples/shaders/compute-shader/src/lib.rs:7:9
  |
7 |         array[i] += i;
  |         ^^^^^^^^^^^^^
```

## System Info

    * rustc: 1.71.0-nightly (1a5f8bce7 2023-05-26)

    * rust-gpu: [b2e5eb7](https://github.com/EmbarkStudios/rust-gpu/commit/b2e5eb75951545782b121665c20f9f1c66fe5bdd)

Original issue: https://github.com/EmbarkStudios/rust-gpu/issues/1075

JulianKnodt:
Previously GEP on zero-sized arrays, would fail. This change fixes arrays to instead emit runtime arrays. I do not know if this will lead to any runtime cost, but it fixes all the compile errors.

I did try out `[(); 0].map(|()| 1)`, and found that that does not compile.

Fixes #151

### DIFF
--- a/tests/ui/lang/core/array/oob_0_array.rs
+++ b/tests/ui/lang/core/array/oob_0_array.rs
@@ -1,0 +1,10 @@
+// build-fail
+
+#![cfg_attr(target_arch = "spirv", no_std)]
+use spirv_std::spirv;
+
+#[spirv(compute(threads(1, 1, 1)))]
+pub fn compute() {
+    let mut array = [0; 0];
+    array[0] = 1;
+}
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -650,8 +650,11 @@ fn trans_aggregate<'tcx>(cx: &CodegenCx<'tcx>, span: Span, ty: TyAndLayout<'tcx>
                 }
                 .def(span, cx)
             } else if count == 0 {
-                // spir-v doesn't support zero-sized arrays
-                create_zst(cx, span, ty)
+                // create a run-time array instead of ZST ADT
+                SpirvType::RuntimeArray {
+                    element: element_type,
+                }
+                .def(span, cx)
             } else {
                 let count_const = cx.constant_u32(span, count as u32);
                 let element_spv = cx.lookup_type(element_type);
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -528,8 +528,13 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                     let stride = ty_kind.sizeof(self)?;
                     ty_size = MaybeSized::Sized(stride);
 
-                    indices.push((offset.bytes() / stride.bytes()).try_into().ok()?);
-                    offset = Size::from_bytes(offset.bytes() % stride.bytes());
+                    if stride.bytes() == 0 {
+                        indices.push(0);
+                        offset = Size::from_bytes(0);
+                    } else {
+                        indices.push((offset.bytes() / stride.bytes()).try_into().ok()?);
+                        offset = Size::from_bytes(offset.bytes() % stride.bytes());
+                    }
                 }
                 _ => return None,
             }
@@ -566,9 +571,8 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
             .iter()
             .map(|index| {
                 result_pointee_type = match self.lookup_type(result_pointee_type) {
-                    SpirvType::Array { element, .. } | SpirvType::RuntimeArray { element } => {
-                        element
-                    }
+                    SpirvType::Array { element, .. }
+                    | SpirvType::RuntimeArray { element } => element,
                     _ => self.fatal(format!(
                         "GEP not implemented for type {}",
                         self.debug_type(result_pointee_type)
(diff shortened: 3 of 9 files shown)


LegNeato:
Awesome, thanks for the PR! 🍻 I'm not super familiar with this code but the change looks reasonable to me. I'll wait for some other maintainers to look at it before reviewing 🤙

Looks like you need to run `cargo fmt` to pass CI.

JulianKnodt:
Somehow this passes locally but not on CI 🤣, I'll have to test a bit more

JulianKnodt:
I've changed my expectation, previously it was erroring for accessing zero sized arrays, but it was hidden behind a deny that is on by default. Instead, I think we should always panic, but with a clear error message. I'm unsure what the behavior would be if it compiled.

LegNeato:
Can we make the error messages match rustc? [For example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b3ba40a592ca0e30abcdb72cd911d92b):

```
error[E0080]: evaluation of constant value failed
 --> src/main.rs:2:18
  |
2 | const VAL: u32 = ARR[0];
  |                  ^^^^^^ index out of bounds: the length is 0 but the index is 0

For more information about this error, try `rustc --explain E0080`.
```

tests/ui/lang/core/array/oob_0_array.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(unconditional_panic)]
+
+#![cfg_attr(target_arch = "spirv", no_std)]
+use spirv_std::spirv;
+
+#[spirv(compute(threads(1, 1, 1)))]
+pub fn compute() {
+    let mut array = [0; 0];
+    // for some reason this now compiles?
```
LegNeato: We should look into this case, no?

tests/ui/lang/core/array/oob_0_array.rs
```diff
@@ -0,0 +1,11 @@
+#![allow(unconditional_panic)]
```
LegNeato, line 3: Is this needed?

LegNeato (changes requested):
Sorry, for the error messages I meant matching the same case in rustc to the case here where it makes sense. I'll leave it up to you to change them so they make sense!

A couple of small comments, thanks again for the PR and sorry for the slow response!

tests/ui/lang/core/array/oob_0_array.rs
JulianKnodt: Those are needed, otherwise it hits the lint and compilation stops.

In theory, most users would stop there, since they see it would panic. But in some rare cases people may try to avoid the lint and keep compiling and we shouldn't fail with a cryptic error message.

tests/ui/lang/core/array/oob_0_array.rs
JulianKnodt: yea, I want to see if it compiles to include a panic, but I'm unsure of how to run the rest to see what happens

Is it run automatically by compiletest

tests/ui/lang/core/array/oob_0_array.rs
LegNeato: 👍

tests/ui/lang/core/array/oob_0_array.rs
LegNeato: No, sadly compiletest just checks it compiles. We don't have any automated runtime tests right now sadly. I usually just drop the code in one of the examples when I am doing a quick manual test.

JulianKnodt: I've checked the output spirv assembly for that example (by adding the compile flags), and it seems that the assignment is getting optimized out somehow. I checked all the spirv builder methods, and none of them seem to be hit. I've found that any reads from the array hit gep and fail, but stores do not. I feel like ideally that should give a runtime panic, but I'm not entirely sure where to inject it.

While investigating this, I also noticed that zst arrays as input parameters do not work.`&mut [T; 0]` also gives the old cryptic error message. I'm not sure if I should fix that in this PR, or another

LegNeato: Is the store behavior changed by this PR? If it is unchanged I'd say just leave it and file a followup task.

Feel free to fix the input message in this PR if you want.

LegNeato: Not sure if this is answered or ready to review, ping if so!

JulianKnodt: Oops I forgot to reply because of the holidays, I checked but the behavior isn't changed, this previously also compiled on main

LegNeato:
@eddyb can you review this when you get a chance?

Issue #169, opened 2024-12-02
buggy bswap implementation seems to assume unsigned arguments


FractalFir:
This snippet of Rust code:
```rust
#![feature(core_intrinsics)]
#![no_std]
use spirv_std::spirv;
#[spirv(compute(threads(1)))]
pub fn main() {
    core::intrinsics::bswap(-1_i32);
}

```
causes a panic in `rust-gpu`:
```
thread 'rustc' panicked at crates/rustc_codegen_spirv/src/builder/builder_methods.rs:1243:9:
Expected types to be equal:
i32
==
u32
stack backtrace:
   0:     0x7fc78dc61935 - std::backtrace_rs::backtrace::libunwind::trace::h19be1b2eef092323
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/std/src/../../backtrace/src/backtrace/libunwind.rs:105:5
   1:     0x7fc78dc61935 - std::backtrace_rs::backtrace::trace_unsynchronized::h5fb87216a898ae13
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7fc78dc61935 - std::sys_common::backtrace::_print_fmt::hb7c11d9dfb93defb
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/std/src/sys_common/backtrace.rs:68:5
   3:     0x7fc78dc61935 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hb67a26e8fbf63961
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x7fc78dcb0a6b - core::fmt::rt::Argument::fmt::hf262f5ae63b7280d
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/core/src/fmt/rt.rs:165:63
   5:     0x7fc78dcb0a6b - core::fmt::write::hf662053177767752
                               at /rustc/244da22fabd9fa677bbd0ac601a88e5ca6917526/library/core/src/fmt/mod.rs:1157:21
   6:     0x7fc78dc5653f - std::io::Write::write_fmt::hbe0746a38afa3ed8
[...]

PR #174, opened 2024-12-09
Support signed integers for `bswap` intrinsic

LegNeato:
Fixes https://github.com/Rust-GPU/rust-gpu/issues/169


LegNeato:
Not sure where we add runtime tests?

LegNeato:
@FractalFir No rush, but I would love it if you can confirm this fixes the issue you saw. If you are busy, no worries!

FractalFir:
It *seems* to work fine, and fix the original issue. There is a *very slim* chance I missed something, but it should be OK.

LegNeato:
Well, even if it is wrong it is better than a panic I guess. Thanks for testing!

PR #176, opened 2024-12-11
Auto self build, or rust-gpu on stable rustc

Firestar99:
A rebased auto self build branch: `spirv-builder` builds rust-gpu's compiler backend automagically in a separate workspace to allow the workspace in which spirv-builder is being used to use a different rustc version than the compiler backend requires. Potentially enables using stable rustc for everything except compiling shaders.

Feel free to pick this up and do something with it!


Firestar99:
Closing in favor of [cargo-gpu](https://github.com/Rust-GPU/cargo-gpu)

PR #177, opened 2024-12-11
Non uniform for everything!

Firestar99:
Adds the NonUniform declaration to pretty much everything, even if it's not required. I don't think this will have a huge performance impact, as Nvidia and Intel don't seem to care about this flag, their accesses are always non-uniform, and Amd's driver is really good at figuring out whether something is uniform / scalar across the workgroup / subgroup or not. 

Feel free to cherry-pick these commits if you do need non-uniform now and can't wait until rust-gpu has figured out a proper solution.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
+++ b/crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
@@ -1,7 +1,7 @@
 use super::Builder;
 use crate::builder_spirv::{SpirvValue, SpirvValueExt, SpirvValueKind};
 use crate::spirv_type::SpirvType;
-use rspirv::spirv::Word;
+use rspirv::spirv::{Decoration, Word};
 use rustc_codegen_ssa::traits::BuilderMethods;
 use rustc_errors::ErrorGuaranteed;
 use rustc_span::DUMMY_SP;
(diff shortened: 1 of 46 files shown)


LegNeato:
I wonder if we should land this behind an off-by-default flag?

Firestar99:
I'm shocked I haven't thought of this earlier... then I wouldn't have to keep updating this PR constantly

Firestar99:
@LegNeato Buffers are easy, but for images, we can't just `cfg` within an `asm!`:
```rust
asm! {
    // error[E0658]: attributes on expressions are experimental
    // error: removing an expression is not supported in this position
    #[cfg(feature = "non-uniform")]
    "OpDecorate %image NonUniform"
    ...
}
```

We can't put it in a separate `asm!` either as we need to decorate the access instruction and all it's parameters (as a precaution). The correct way would be to `cfg` the entire `asm!` and duplicate all the image access `asm!`, about 200 loc. 

Would you mind if I just add a linking stage instead to remove all `OpDecoration` if the target feature `ShaderNonUniform` is *not* present? (cc @eddyb)

LegNeato:
I'm personally fine with either a linking stage or copying 200 lines (or a macro so no need to copy?)

Firestar99:
I've taken the linking stage filter route and added the `NonUniform` decoration it everywhere, hopefully didn't miss anything. Tested it on my project too. 

I've added some tests in branch `byte_adressable_buffer_from_runtime_array`, which work on `spv1.3` but fail on `vulkan1.2`. That behavior doesn't change on master, so unrelated to this PR. Weirdly enough, that's exactly how I'm using them in my project, yet it compiles with `vulkan1.2` just fine.

Firestar99:
@LegNeato this is ready to review, in case it has gone under. Just needed a rebase due to compiletest conflicts.

LegNeato (approved):
LGTM!

PR #187, opened 2024-12-20
make vector and scalar traits imply `Send + Sync + 'static`

Firestar99:
This makes the traits `VectorOrScalar`, `Vector`, `Scalar` and `image::params::SampleType` implicitly be `Send + Sync + 'static`. This has always been true for all impl anyway, and as these are all sealed traits, this change is *not* breaking. 

This mostly helps me with my bindless API where I always have to do `Generic: SampleType + Send + Sync + 'static` and now I can just do `Generic: SampleType`.

### DIFF
--- a/crates/spirv-std/src/vector.rs
+++ b/crates/spirv-std/src/vector.rs
@@ -9,7 +9,7 @@ use glam::{Vec3Swizzles, Vec4Swizzles};
 /// # Safety
 /// Implementing this trait on non-scalar or non-vector types may break assumptions about other
 /// unsafe code, and should not be done.
-pub unsafe trait VectorOrScalar: Default {
+pub unsafe trait VectorOrScalar: Default + Send + Sync + 'static {
     /// Either the scalar component type of the vector or the scalar itself.
     type Scalar: Scalar;
 


crates/spirv-std/src/vector.rs
```diff
@@ -9,7 +9,7 @@ use glam::{Vec3Swizzles, Vec4Swizzles};
 /// # Safety
 /// Implementing this trait on non-scalar or non-vector types may break assumptions about other
 /// unsafe code, and should not be done.
-pub unsafe trait VectorOrScalar: Default {
+pub unsafe trait VectorOrScalar: Default + Send + Sync + 'static {
```
eddyb: Would it make sense to also throw in `Copy`? (I can't think of a lot of other traits, maybe `PartialEq` but ehhh).

Oh, actually:
https://github.com/Rust-GPU/rust-gpu/blob/8788f924cbc8aa67bd9f9ba0627a6fb5693bc411/crates/spirv-std/src/scalar.rs#L11-L12

So that confirms my expectation that `Copy` should be included. ~~Can you also add `Send + Sync + 'static` to that trait?~~

~~Also, why is this trait not actually sealed the same way the `Scalar` trait is? Maybe leftover from the `glam` days? (unless it was sealed and that got undone)~~

eddyb (changes requested):

crates/spirv-std/src/vector.rs
eddyb: Oops I missed the `VectorOrScalar<Scalar = Self>` that `trait Scalar` requires.

In that case, can you just move `Copy` from `Scalar` to `VectorOrScalar`?

Firestar99: ~~I added that trait so likely I just forgot~~ Actually `trait Vector` is not sealed, only `Scalar` is

eddyb (approved):

PR #188, opened 2024-12-23
spirv-builder: add `RUSTGPU_CARGOFLAGS`, as a Cargo counterpart to `RUSTGPU_RUSTFLAGS` (for `rustc`).

eddyb:
Had this lying around intermixed with `-Zscript` experiments I never completed, and we keep running into situations where this simple addition would be useful.


LegNeato (approved):
Would be nice to add this to the changelog or docs somewhere but not blocking.

Issue #192, opened 2025-01-02
Add `tracing` support to `rustc_codegen_spirv`


LegNeato:
When working on or debugging the rust-gpu compiler backend, it is extremely hard to see what is going on. Ideally we have `tracing` support so that we can put logs and spans in to see what the compiler is doing (behind a feature).

`rustc` already has support: https://rustc-dev-guide.rust-lang.org/tracing.html

One complication is that `rustc_codegen_spirv` is used as a dylib and I don't believe tracing works in that situation. I found https://github.com/JakkuSakura/tracing-shared-rs, but it doesn't look great.

LegNeato:
For posterity that page mentions that the official compiler includes tracing for info:
> "This is an `info!`  log rather than a `debug!` log so it will work on the official compilers."

LegNeato:
Looks like we can reuse rustc's version, but again we hit the only info is supported. Rustdoc deals with this by using both:

https://github.com/rust-lang/rust/blob/bf6f8a4d328f7f3b0f6ea8205ad28591cc11aafd/src/librustdoc/lib.rs#L167

I am investigating this solution.

PR #193, opened 2025-01-02
Add tracing to rustc_codegen_spirv

LegNeato:
Fixes https://github.com/Rust-GPU/rust-gpu/issues/192


LegNeato:
Open questions:

1. Do we want to do it in `__rustc_codegen_backend()` so we can trace very early?
2. Do we want to use our own env var? (I mirrored it off `rustc` and `rustdoc`, but I can see a case for reusing `RUSTC_LOG`
3. Do we want a test? We can probably put an "I am alive" type log item and then run a passing compiletest asserting on the stderr output.

LegNeato:
https://github.com/tokio-rs/tracing/pull/3183 for `nu-ansi-term` version fix.

LegNeato:
I'm not sure I agree with the lint...I am mirroring what `rustc` and `rustdoc` do themselves. Thoughts? Should I change it to only use codegen args?

LegNeato:
https://github.com/tokio-rs/tracing/pull/3184 for regex stuff.

LegNeato:
I am not going to document this until a future PR where there is actually logging in place.

LegNeato:
Another open question, do we want a `tracing` feature that completely compiles tracing-related code out if we want? There is some non-trivial work in https://github.com/Rust-GPU/rust-gpu/pull/196 to generate info for the logs that happens regardless if tracing is being used or not.

It looks like tracing has a couple of knobs to query, so we can likely gate code behind that (but it will still be in the binary).

I would trend towards compiling it in always without a feature, as that is what rustc does, but wanted to confirm with others.

LegNeato:
@Firestar99 any thoughts on the open questions in https://github.com/Rust-GPU/rust-gpu/pull/193#issuecomment-2568295863 ?

Firestar99:
To most of these: I don't really know, as I haven't used the tracing feature, nor done more than 2 weeks of hacking in the actual compiler. Most of my contributions have been intrinsics, which don't need deep end compiler changes, or just rebasing stuff.

3. I don't think we really need a feature for the compiler

LegNeato:
Ok, I am going to land these then and we can change later.

PR #196, opened 2025-01-03
Replace printing codegen args with `tracing`

LegNeato:
Depends on https://github.com/Rust-GPU/rust-gpu/pull/193.

This removes the following options and replaces them with `tracing`:

* `--specializer-debug` => RUSTGPU_LOG=rustc_codegen_spirv::specializer=debug
* `--print-zombie` => RUSTGPU_LOG=print_zombie=debug
* `--print-all-zombie` => RUSTGPU_LOG=print_all_zombie=debug


LegNeato:
Questions:
1. Do we want some sort of prefix for all these instead of `print` for future expansion?
2. Do we want to do kebab case or underscore?

Firestar99:
Suggestion: use the [`profiling`](https://crates.io/crates/profiling) crate instead of `tracing`

Profiling uses features to specify which profiler you want to use, and ofc it has `profile-with-tracing` to forward all the zones to tracing. But can also support other profilers, like the excellent [tracy profiler](https://github.com/wolfpld/tracy). And only the final binary has to specify the features, the intermediary crates can just use profiling without any features and incur no runtime overhead or feature forwarding hell. 
However, debugging via logging would no longer be universally available and require a feature flag to work.

LegNeato:
I think that is a good idea, but this isn't for profiling--just logging debug data. Looks like they suggest still using tracing for non-profiling / more advanced use cases. We do want profiling, but that is orthogonal (and can use tracing as it's backend)

LegNeato:
I also found https://docs.rs/tracing-tracy/latest/tracing_tracy/. Not sure which way is ultimately better, will probably require some hands on experience.

LegNeato:
Looks like rustc has profiling built in, so we should probably use that for profiling I think.

Issue #199, opened 2025-01-05
Codegen error for `log10`


bspeice:
### Expected Behaviour
Able to compile a SPIR-V compute shader using the `log10` operation.

### Example & Steps To Reproduce

Minimal reproducer available at: https://github.com/bspeice/rust-gpu-log10-err/blob/main/crates/shader/src/lib.rs

I hit an error when `spirv-builder` attempted to run a shader through `spirv-val`:

```rust
#![no_std]
use spirv_std::glam::UVec3;
use spirv_std::num_traits::Float;
use spirv_std::spirv;

#[spirv(compute(threads(1)))]
pub fn render_image_histogram(
    #[spirv(global_invocation_id)] _global_id: UVec3,
    #[spirv(storage_buffer, descriptor_set = 0, binding = 0)] input: &[f32],
    #[spirv(storage_buffer, descriptor_set = 0, binding = 1)] output: &mut [f32],
) {
    output[0] = input[0].log10();
}
```

```
  error: Expected int scalar or vector type as Result Type: IMul
    |
    = note: module `/home/bspeice/Development/log10/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/shader.spv`

  warning: an unknown error occurred
    |
    = note: spirv-opt failed, leaving as unoptimized
    = note: module `/home/bspeice/Development/log10/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/shader.spv`

  error: error:0:0 - Expected int scalar or vector type as Result Type: IMul
           %35 = OpIMul %float %float_0_434294492 %34
    |
    = note: spirv-val failed
    = note: module `/home/bspeice/Development/log10/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/shader.spv`

  warning: `shader` (lib) generated 1 warning
  error: could not compile `shader` (lib) due to 2 previous errors; 1 warning emitted
  Error: BuildFailed
```

If I understand the output correctly, codegen produced a SPIR-V output that included an `OpIMul`, when it should have been an `OpFMul`. `spirv-opt` then failed, and the output was left unoptimized. Afterward, `spirv-val` failed validation for the same problem that `spirv-opt` failed on.

[...]

PR #200, opened 2025-01-06
Use `fmul` in `log10` intrinsic

LegNeato:
Fixes https://github.com/Rust-GPU/rust-gpu/issues/199.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed 🩹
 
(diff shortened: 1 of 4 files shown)


Firestar99 (approved):
Looks good. Don't know why there's two locations for the log10 intrinsic, but let's just get that fixed first.

LegNeato:
One is active when using libm, the other is when not using libm. Not sure if both are still needed or not, original was https://github.com/EmbarkStudios/rust-gpu/issues/81. @eddyb care to enlighten us?

LegNeato:
I'm going to land this and file a followup for possible cleanup, as at the very least the code as is is wrong in both places and this PR fixes them, so it is a net improvement.

PR #213, opened 2025-01-29
add support for leading_zeros and trailing_zeros

Firestar99:
In main they require some obscure Intel Extension. This PR changes them to use the `GLSL.std.450` functions `FindILsb` and `FindUMsb`, which do the exact same thing.

Currently panics during linking, probably cause I screwed up branching *somewhere*


Firestar99:
I basically want this:
```glsl
#version 450

layout(location=0) flat in uint inbla;
layout(location=0) out uint outbla;

uint trailing_zeros(uint v) {
    uint ret;
    if (v == 0) {
        ret = 32;
    } else {
        ret = findLSB(v);
    }
    return ret;
}

void main() {
    outbla = trailing_zeros(inbla);
}
```

LegNeato:
Sweet! I actually looked at this too, just didn't put my WIP up yet. Differences between what I did:

* Kept the intel stuff, but made it queried / optional.
* Used bitshifts and masking to support rather than the GlOp (I would imagine the GLop is faster?). I basically just looked at stack overflow (https://stackoverflow.com/questions/23856596/how-to-count-leading-zeros-in-a-32-bit-unsigned-integer)
* handled zero with a select, and didn't handle it for the non-zero case as it is unnecessary:

```rust
                // Handle the zero case for `cttz`
                if name == sym::cttz {
                    let is_zero =
                        self.icmp(IntPredicate::IntEQ, input, self.constant_int(input.ty, 0));
                    let result_if_zero = self.constant_int(input.ty, bit_width as u128);

                    let nonzero_result = compute_ctlz_nonzero(input, self);

                    // Select between the zero case and non-zero calculation
                    self.select(is_zero, result_if_zero, nonzero_result);
                }

                // For `ctlz_nonzero`, directly compute the result
                compute_ctlz_nonzero(input, self)
```

Firestar99:
This is ready for review, but completely untested beyond compile tests. Would love some integration tests or fuzzing right now, as it's quite easy to screw up bitcasts or misread documentation.

LegNeato:
You could rebase on https://github.com/Rust-GPU/rust-gpu/pull/216 ?

LegNeato:
Yep! Now you can add a runtime diff test for this if you want. But I'll also be revamping a bit and squashing that PR, so I don't want to make your git stuff more annoying. Up to you if it is worth it or want to land as-is!

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -418,6 +414,78 @@ impl Builder<'_, '_> {
...
+                        let undef = self.undef(ty).def(self);
+                        self.zombie(undef, &format!(
+                            "counting leading / trailing zeros on unsupported {ty:?} bit integer type"
+                        ));
+                        undef
+                    }
+                }
+                    .with_type(ty)
+            }
+            _ => self.fatal("count_ones on a non-integer type"),
```
LegNeato: Nit: wrong function name in comment

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -418,6 +414,78 @@ impl Builder<'_, '_> {
...
+                        if signed {
+                            self.emit().bitcast(ty, None, result).unwrap()
+                        } else {
+                            result
+                        }
+                    }
+                    _ => {
+                        let undef = self.undef(ty).def(self);
+                        self.zombie(undef, &format!(
+                            "counting leading / trailing zeros on unsupported {ty:?} bit integer type"
```
LegNeato: Nit: wrong function name in comment

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -418,6 +414,78 @@ impl Builder<'_, '_> {
...
+                        let reverse = self.emit().bit_reverse(u32, None, arg).unwrap();
+                        self.emit().bitcast(ty, None, reverse).unwrap()
+                    },
+                    (64, signed) => {
+                        let u32_32 = self.constant_u32(self.span(), 32).def(self);
+                        let arg = arg.def(self);
+                        let lower = self.emit().s_convert(u32, None, arg).unwrap();
+                        let higher = self
+                            .emit()
+                            .shift_left_logical(ty, None, arg, u32_32)
```
LegNeato: I think this should be `shift_right_logical` ?

LegNeato (changes requested):

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
Firestar99: all the 64bit bitshifts were wrong :D

Firestar99:
I've removed the differential testing harness again, as you've said it isn't ready for review yet, and it tests rust-gpu vs wgsl whereas I need rust-gpu vs standard rust on the cpu. 

Just did some basic diff testing in my engine against a software impl I've been using as a replacement, and something isn't right with trailing / leading zeros.

Firestar99:
I made a test in my bindless framework ([shader](https://gitlab.com/spacegamedev/space-rust/-/blob/count_leading_trailing_zeros/rust-gpu-bindless/tests/integration-test-shader/src/count_leading_trailing.rs) [cpu](https://gitlab.com/spacegamedev/space-rust/-/blob/count_leading_trailing_zeros/rust-gpu-bindless/tests/integration-test/src/count_leading_trailing.rs)) and can confirm that `u32` is working! 

However, I'm running into issues testing out other unsigned types: 
```rust
error: cannot load type u8 in an untyped buffer load
    --> /home/firestar99/.cargo/git/checkouts/rust-gpu-11142fd2aadc2318/57c7d9d/crates/spirv-std/src/byte_addressable_buffer.rs:97:9
     |
  97 |         buffer_load_intrinsic(self.data, byte_index)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  error: cannot store type struct count_leading_trailing::CountLeadingTrailingZerosTransfer<u8, u8, u8> { count: u32, leading: u32, trailing: u32, input: u8, swap_bytes: u8, reverse_bits: u8 } in an untyped buffer store
     --> /home/firestar99/.cargo/git/checkouts/rust-gpu-11142fd2aadc2318/57c7d9d/crates/spirv-std/src/byte_addressable_buffer.rs:150:9
      |
  150 |         buffer_store_intrinsic(self.data, byte_index, value);
      |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: due to containing type 49
```

I may rework this to cast them to u32 and back

Firestar99:
u64 fails in the exact same way... 

@LegNeato I say we just merge it for now and hope anything non-u32 works as well. At the very least I fixed u32.

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -398,6 +367,254 @@ impl<'a, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'tcx> {
 }
 
 impl Builder<'_, '_> {
+    pub fn count_ones(&self, arg: SpirvValue) -> SpirvValue {
+        let ty = arg.ty;
+        match self.cx.lookup_type(ty) {
+            SpirvType::Integer(bits, signed) => {
+                let u32 = SpirvType::Integer(32, false).def(self.span(), self);
+
+                match bits {
```
LegNeato, line 376: Just a style thing, but I generally feel it is clearer to matched on `signed` as well, it makes all the cases clearer, similar to what you did in bitcast.

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -398,6 +367,254 @@ impl<'a, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'tcx> {
...
+                    32 => self.emit().bit_count(u32, None, arg.def(self)).unwrap(),
+                    64 => {
+                        let u32_32 = self.constant_u32(self.span(), 32).def(self);
+                        let arg = arg.def(self);
+                        let lower = self.emit().s_convert(u32, None, arg).unwrap();
+                        let higher = self
+                            .emit()
+                            .shift_right_logical(ty, None, arg, u32_32)
+                            .unwrap();
+                        let higher = self.emit().s_convert(u32, None, higher).unwrap();
```
LegNeato: Doesn't this need to change depending on sign?

LegNeato (changes requested):

Firestar99:
To correct things: I implemented it for u8, u16, u32 and u64 and today I wrote a test with my bindless stuff to debug why `u32::trailing_zeros()` was giving me crap values. But I was not able to easily adapt the test to anything other than u32, due to limitations with ByteAddresableBuffer not allowing you to read or write u8, u16 and u64 for whatever reason. (Which I was completely unaware of)

Anyways, I noticed that rustc actually does not call any of these methods on the signed types, as the rust std converts them to unsigned beforehand. So I can just remove all the signed conversions, *except* for `count_leading_trailing_zeros()` for some odd reason, even though it too converts to unsigned in rust std. 

Give me another iteration on this one, but I'd like to land it before we release.

Firestar99:
@LegNeato Did another iteration, hopefully a lot cleaner now. Removed the signed-ness everywhere it's not needed, with just one extra branch for the `leading_zeros()` special case, see previous edited comment.

LegNeato (approved):
LGTM!

PR #218, opened 2025-03-16
Use ARTIFACT_SUFFIX

LegNeato:


schell:
@LegNeato could you add a little description just to give me some context? Where does `ARTIFACT_SUFFIX` come from?

LegNeato:
Sure! It is defined in that file as the same static string. The suffix is json metadata that is written out next to the artifact...sort of an API between the compiler backend and spirvbuilder. It appears to be only written with certain crate types. I was debugging another issue and saw that in the same file it uses the hardcoded string and the constant, likely just an oversight:

https://github.com/Rust-GPU/rust-gpu/pull/218/files#diff-84c98121dbc8466f5e6e1503a3771f2ec4f8f37fa048ebf01c201b1b5cd1b191R903

schell (approved):

PR #219, opened 2025-03-16
Add a message to suggest checking crate-type

LegNeato:
This was my issue and it was unclear the cause. I am not sure if this is the only case where this error shows up, hence the wording. Open to better wording if you have it.


schell (approved):

Firestar99:
> Verify that `crate-type` is set correctly, eg. `crate-type = ["lib", "dylib"]` within the `[lib]` section of your shader's `Cargo.toml`.

I know technically "dylib" is sufficient, but without `lib` I'm concerned that if the crate is also used on the CPU, it switches to dynamic linking instead of static linking and inlining + LTO can't do much then. But take this with a grain of salt, not too familiar with this topic.

PR #223, opened 2025-03-30
add `SpirvBuilder::target_dir_path` to set the target dir explicitly

Firestar99:
If you have two shader crates, one has to wait for the other to finish compiling before it can compile. That's a good thing, as you don't have to recompile any dependencies twice. 

But if you have two shader crates **with differing capabilities or features**, they are forced to do a clean rebuild every time, **adding 30s each build, even if nothing changed**. Explicitly specifying a target dir enables me to specify a separate build dir for each shader crate and work around this issue. 

I do think a better policy wrt. capabilities and features will be needed in the future, but this will do for now.


LegNeato (approved):
I could have used this for my difftest PR as I just set the arg directly 🤔

PR #227, opened 2025-04-09
Array init with 0i32 uses 0u32 instead

LegNeato:


LegNeato:
Thanks @JulianKnodt for the original fix!

Firestar99:
btw in compiletest you can put multiple entry points in a single file if you just want to check that they compile and not disassemble them

Firestar99 (approved):

LegNeato:
Yeah, those were just the original tests and figure it doesn't hurt really.

PR #230, opened 2025-04-10
refactor derivatives

Firestar99:
Adds a new `trait Derivative` that is implemented for `f32`, `Vec2`, `Vec3`, `Vec3A` and `Vec4`. Disallows f64 to be derived, spec expects f32 only. Renamed derivatives from `ddx()` to `dfdx()` to better match glsl's `dFdx()`.

Spec section: https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_derivative_instructions

### DIFF
--- a/crates/spirv-std/src/arch/derivative.rs
+++ b/crates/spirv-std/src/arch/derivative.rs
@@ -36,6 +36,7 @@ pub unsafe trait Derivative: Sealed + Default {
     ///
     /// This instruction is only valid in the Fragment Execution Model.
     #[crate::macros::gpu_only]
+    #[inline]
     fn ddx(self) -> Self {
         deriv_fn!(OpDPdx, self)
     }
@@ -48,6 +49,7 @@ pub unsafe trait Derivative: Sealed + Default {
     ///
     /// This instruction is only valid in the Fragment Execution Model.
     #[crate::macros::gpu_only]
+    #[inline]
     fn ddx_fine(self) -> Self {
         cap_deriv_control!();
         deriv_fn!(OpDPdxFine, self)
@@ -63,6 +65,7 @@ pub unsafe trait Derivative: Sealed + Default {
     ///
     /// This instruction is only valid in the Fragment Execution Model.
     #[crate::macros::gpu_only]
+    #[inline]
     fn ddx_coarse(self) -> Self {
         cap_deriv_control!();
         deriv_fn!(OpDPdxCoarse, self)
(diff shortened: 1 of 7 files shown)


crates/spirv-std/src/arch/derivative.rs
```diff
@@ -1,102 +1,163 @@
-use crate::float::Float;
+use crate::sealed::Sealed;
+use glam::{Vec2, Vec3, Vec3A, Vec4};
+
+#[cfg(target_arch = "spirv")]
+macro_rules! cap_deriv_control {
```
LegNeato: This macro is used elsewhere but impl is commented out?

LegNeato (changes requested):
Is there any way we can polyfill for `f64`?

Firestar99:
Forget about polyfiling this for f64 or anything else really. This is a fragment shader only instrinsic for computing derivatives with your neighboring fragments, which are needed in eg. texture operations for mip level selection.

crates/spirv-std/src/arch/derivative.rs
Firestar99: I'll remove it for now, can always undo later if you don't squash.

But interestingly, you can use this to enable a capability from the code itself, and do not need to predeclare it in SpirvBuilder. I want to follow this up a little more, cause I would like to get rid of predeclaring them at all, and you just enable features on the things you use implicitly. Note that using that shader (in Vulkan, unsure about wgpu) requires you to explicitly enable said device feature, so you'd still notice when you accidentally were to use too many features. 

Refactoring this would completely remove recompilations due to changed predeclared capabilities, which change absolutely nothing anyway and is a blocker for redistributing build scripts with shaders. And *could* allow shaders within the same crate to only require the capabilities that the shader actually needs, but this needs more research on how DCE is performed on capabilities.

LegNeato: That would be cool, but commented out it does nothing. So yeah, let's remove and follow-up. I'm a little hesitant that the capabilities code relies on would not be statically known. I think we should take a large hint to how simd is handled in rustc with compile control and runtime checks.

Firestar99: I have a commit to just comment it out, now I got one commit that removes it and can be reverted later

LegNeato:
Looks like CI failing:

`docs for unsafe trait missing `# Safety` section`

LegNeato (approved):
Awesome!

Issue #46, opened 2024-11-04
Regression on main when allocating arrays of arrays


LegNeato:
Minimal repro that fails to compile:

```rust
use spirv_std::spirv;

#[spirv(fragment)]
pub fn main() {
    let x = [[1; 2]; 1];
}
```

Bisects to https://github.com/Rust-GPU/rust-gpu/commit/02cd3248ad13df8e0b4fa2962ceb57ac089e2d49

It was found in https://github.com/Rust-GPU/rust-gpu/pull/21#issuecomment-2381578892 (assuming it is the same issue, but at least that is what I started with to get a repro)

LegNeato:
@eddyb not sure if this will be fixed by your qptr stuff.

LegNeato:
Yep, that's the error:

```
error: cannot cast between pointer types
       from `*i32`
         to `*[i32; 2]`
```

LegNeato:
Also happens with arrays of vecs, which is more common: https://github.com/Rust-GPU/rust-gpu/issues/189

LegNeato:
Well, I am not smart enough / haven't cracked this yet. Spent a couple of hours on it.  I am using the following so I can see each part:

```
pub fn main() {
    let x = [(true, 2.0); 3];
}
```

What I am seeing is that the array is handled ok, but when it recurses into the ADT it then loops over the fields and uses the first field as the leaf type. So the issue seems to be with `maybe_inbounds_gep` and the interaction with `recover_access_chain_from_offset` (which is obvious as those were added in the commit that caused this of course) and leafs vs compound types.

PR #233, opened 2025-04-12
Correctly calculate access chains for compound types 

LegNeato:
Fixes https://github.com/Rust-GPU/rust-gpu/issues/46.

This was extremely painful and I am still not sure I understand the code completely and all the various cases.

@schell , @Firestar99 , please try it in your projects and see if I broke anything!


Firestar99:
My stuff still works with this branch (+ nonuniform patches) :tada:

schell:
My project compiled just fine and passes tests 👍 . I also successfully built a shader with `[Default::default(); N]` in it. 

Awesome work!

schell (approved):
LGTM but take my review with a grain of salt 🧂 as I never got a chance to dive in very deep, but I have tested the results.

LegNeato:
@eddyb said he wanted to take a look at this so I'll wait a bit before merging.

LegNeato:
Ok, I am just going to land it and we can fix forward or revert if it isn't correct.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -434,11 +447,14 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
         let ptr = ptr.strip_ptrcasts();
         let mut leaf_ty = match self.lookup_type(ptr.ty) {
             SpirvType::Pointer { pointee } => pointee,
-            other => self.fatal(format!("non-pointer type: {other:?}")),
+            other => self.fatal(format!("`ptr` is non-pointer type: {other:?}")),
         };
 
-        // FIXME(eddyb) this isn't efficient, `recover_access_chain_from_offset`
-        // could instead be doing all the extra digging itself.
```
eddyb, line 441: I am not sure what this comment meant exactly, but I suspect it still applies to the loop being a loop, and not a single call to `recover_access_chain_from_offset`.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
                 }
                 SpirvType::Vector { element, .. }
                 | SpirvType::Array { element, .. }
                 | SpirvType::RuntimeArray { element }
                 | SpirvType::Matrix { element, .. } => {
+                    trace!("recovering access chain from Vector, Array, RuntimeArray, or Matrix");
                     ty = element;
+                    trace!("setting ty = element: {:?}", self.debug_type(element));
                     ty_kind = self.lookup_type(ty);
+                    trace!("looked up ty kind: {:?}", ty_kind);
```
eddyb, line 599: I don't really have anything against the extra `trace!`s but it does make me wish for something halfway between `tracing` and a debugger - one of the few things higher-level languages happen to have a much easier time supporting.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
                 && leaf_ty.map_or(true, |leaf_ty| leaf_ty == ty)
             {
+                trace!("returning type: {:?}", self.debug_type(ty));
+                trace!("returning indices with len: {:?}", indices.len());
                 return Some((indices, ty));
             }
         }
     }
 
+    #[instrument(level = "trace", skip(self), fields(ty = ?self.debug_type(ty), ptr, combined_indices = ?combined_indices.iter().map(|x| (self.debug_type(x.ty), x.kind)).collect::<Vec<_>>(), is_inbounds))]
```
eddyb, line 630: Surprised `rustfmt` didn't make this multi-line, I feel like it could be more readable if each entry in `fields` was on its own line.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+        //  The base pointer value for the GEP operation.
         ptr: SpirvValue,
+        // A slice of indices used for the GEP calculation. The *first* index is treated
+        // as the main offset from the base pointer `ptr`, scaled by the size of `ty`.
+        // Subsequent indices navigate through nested aggregate types (structs, arrays,
+        // vectors, etc.) starting from `ty`.
         combined_indices: &[SpirvValue],
+        // If true, generate an `OpInBoundsAccessChain`, which has stricter requirements
+        // but allows more optimization. If false, generate `OpAccessChain`.
         is_inbounds: bool,
```
eddyb, line 646: I wonder if `///` doc comments work on function parameters - and ofc I'm more used to having the whole function being documented instead, but this is quite thorough nonetheless and it will be sad when we eventually have powerful enough legalization to be able to delete all of this.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+                | SpirvType::RuntimeArray { element }
+                | SpirvType::Vector { element, .. }
+                | SpirvType::Matrix { element, .. } => element,
+                // Special case: If we started with a byte GEP (`is_byte_gep` is true) and
+                // we are currently indexing into a byte type, the result is still a byte type.
+                // This prevents errors if `indices` contains non-zero values when `ty` is u8/i8.
+                SpirvType::Integer(8, signedness) if is_byte_gep => {
+                    // Define the resulting byte type as it might not exist yet).
+                    SpirvType::Integer(8, signedness).def(self.span(), self)
+                }
```
eddyb, line 699: I don't see how this could be reached - `indices` can only dig into structs/arrays, and only a bug elsewhere would allow `indices.len() > 0` when doing byte offsetting.

Locally removing this case does not fail any tests, so at the very least none were added that could hit it.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
             SpirvType::Pointer { pointee } => pointee,
             other => self.fatal(format!("gep called on non-pointer type: {other:?}")),
         };
 
-        // HACK(eddyb) `struct_gep` itself is falling out of use, as it's being
-        // replaced upstream by `ptr_add` (aka `inbounds_gep` with byte offsets).
-        //
-        // FIXME(eddyb) get rid of everything other than:
-        // - constant byte offset (`ptr_add`?)
-        // - dynamic indexing of a single array
```
eddyb, line 623: Again, my comment is being removed but I don't see why - in fact, after https://github.com/rust-lang/rust/pull/134117 the situation is even weirder, and I'm not sure that a GEP originating from `rustc_codegen_ssa` can even have any `indices` anymore (only `ptr_base_index`), so we might be able to transition to 100% "recover from offset".

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-            .and_then(|idx| Some(u64::try_from(idx).ok()? * self.lookup_type(ty).sizeof(self)?));
-        if let Some(const_ptr_offset) = const_ptr_offset {
+            .lookup_const_scalar(ptr_base_index) // Check if ptr_base_index is constant scalar
+            .and_then(|idx| {
+                let idx_u64 = u64::try_from(idx).ok()?;
+                // Get the size of the input type `ty`
+                self.lookup_type(ty)
+                    .sizeof(self)
+                    // Calculate offset in bytes
+                    .map(|size| idx_u64.saturating_mul(size.bytes()))
```
eddyb, line 753: AFAICT `a.saturating_mul(b)` is the wrong tool in these kinds of situations - the only reason I used `*` is because this should be in bounds of some allocation, but if you want to protect from bugs elsewhere you need `a.strict_mul(b)` aka `a.checked_mul(b).unwrap()`.

eddyb: Wait, why use `map`? You're in an `and_then`, so `?` will work just fine - also, why multiply `.bytes()`?

Ohhh I see - this actually relates to the other change: the `*` (`std::ops::Mul`) on `Size` ~~might actually~~ [does a `checked_mul` itself](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_abi/lib.rs.html#662-679) (just checked and linked the current source code, that's why `*` was used in this code before).

It's good to use `Size` because it checks for overflow and it semantically indicates the unit (i.e. bytes), so this part should be changed back to what it was before.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+            // attempts to find a sequence of constant indices (`base_indices`) into
+            // `original_pointee_ty` that matches the `const_ptr_offset_bytes`.
             if let Some((base_indices, base_pointee_ty)) = self.recover_access_chain_from_offset(
+                // Start from the pointer's actual underlying type
                 original_pointee_ty,
-                const_ptr_offset,
+                // The target byte offset
+                Size::from_bytes(const_ptr_offset_bytes),
+                // Allowed range (not strictly needed here?)
                 Some(Size::ZERO)..=None,
```
eddyb, line 770: I forget exactly what `None..=None` does, but `Size::ZERO` as a lower bound could indicate that a field containing that exact offset must be found *not just a field somewhere after that offset*.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-                        Some((base_ptr, indices))
-                    } else {
-                        None
-                    }
-                });
-            base_ptr_and_combined_indices
+            // Search the current function's instructions...
+            // FIXME(eddyb) this could get ridiculously expensive, at the very least
+            // it could use `.rev()`, hoping the base pointer was recently defined?
+            let search_result = {
```
eddyb, line 867: Why the extra variable? It just inflates the diff needlessly.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+                    );
+                    final_spirv_ptr_type
+                } else {
+                    trace!(
+                        "Byte GEP allowing recovered type for wrapper: {}",
+                        // Use type based on recovery result
+                        self.debug_type(base_pointee_ty)
+                    );
+                    self.type_ptr_to(base_pointee_ty)
+                };
```
eddyb, line 804: ~~This looks like the only addition to the "constant offset" path and I don't understand it.~~

~~You *have to* use `final_spirv_ptr_type` (i.e. pointer-to-`calculated_pointee_type`) because that's what the type *has to be*.~~

When `indices` is empty, that does mean that pointer-to-`base_pointee_ty` can be obtained and it has the right offset, even if it's not pointing to a `ty`/`calculated_pointee_type` (which are equal in the empty `indices` case) - but there are two issues there:
- "`ty` is NOT byte, OR `base_pointee_ty` *is* byte" doesn't seem like it matches the comment, nor can it relate to "do we have the desired pointer, even if of the wrong type?"
  - maybe the actual issue here is that the "then" and "else" sides got flipped for no good reason, so the condition ends up negated, and should be `is_byte_gep && ...`?
- the caller *definitely expects* a value of type `final_spirv_ptr_type`, so if the access chain *inherently* has a different type, it must be cast to the right type (which would create a `LogicalPtrCast`, and hopefully not create downstream issues)
  - to explain "inherently" a bit further: an access chain *always* has *only one* valid output type, given the input pointer type and the number of indices (and value, when dealing with struct fields), and we can't just put in a different type and have valid SPIR-V

---

My bad, I kept missing the `let result_pointee_type = if indices.is_empty() {` in the old code, now it makes sense that there was something there (but I've kept my original comment for posterity).

That's exactly the condition I was describing earlier, where `indices` being empty determines what the *correct* type is. However, I suppose the old code was lacking the `pointercast` I mentioned above.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
 
-        // HACK(eddyb) temporary workaround for untyped pointers upstream.
-        // FIXME(eddyb) replace with untyped memory SPIR-V + `qptr` or similar.
+        // --- Fallback / Default Path ---
+
+        // This path is taken if neither the Recovery nor the Merging path succeeded or applied.
+        // It performs a more direct translation of the GEP request.
+
+        // HACK(eddyb): Workaround for potential upstream issues where pointers might lack precise type info.
+        // FIXME(eddyb): Ideally, this should use untyped memory features if available/necessary.
```
eddyb, line 970: These should've stayed next to the `pointercast` - the new comments might not even make sense, because the cast is only due to the mismatch - anyway, it's all a historical mess, and lots of concepts don't fit the new (much more untyped) approach, sadly.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: This keeps `issue-46` passing, but I will admit I don't know why it's valid:
```patch
@@ -785,8 +778,9 @@
                 // type, we can use the pointer type derived from the *recovered* type
                 // (`base_pointee_ty`). This helps preserve type information when
                 // recovery works for byte addressing.
-                let result_wrapper_type = if !is_byte_gep
-                    || matches!(self.lookup_type(base_pointee_ty), SpirvType::Integer(8, _))
+                let result_wrapper_type = if !(is_byte_gep
+                    && !matches!(self.lookup_type(base_pointee_ty), SpirvType::Integer(8, _))
+                    && indices.is_empty())
                 {
                     trace!(
                         "Using strictly calculated type for wrapper: {}",
@@ -828,13 +822,17 @@
                     );
                     // Emit a single AccessChain using the original pointer `ptr_id` and the fully combined index list.
                     // Note: We don't pass `ptr_base_index` here because its effect is incorporated into `base_indices`.
-                    return self.emit_access_chain(
+                    let mut final_ptr = self.emit_access_chain(
                         result_wrapper_type, // The chosen result pointer type
                         ptr_id,              // The original base pointer ID
                         None,                // No separate base index needed
                         combined_indices,    // The combined structured + original indices
                         is_inbounds,         // Preserve original inbounds request
                     );
+                    if result_wrapper_type != final_spirv_ptr_type {
+                        final_ptr = self.pointercast(final_ptr, final_spirv_ptr_type);
+                    }
+                    return final_ptr;
                 } else {
                     // Recovery happened, but the recovered type `base_pointee_ty` doesn't match the input `ty`,
                     // AND there are more `indices` to process. Using the `base_indices` derived from
```

I suppose the trick is that `is_byte_gep` *requires* `indices.is_empty()` (as bytes have no inner structure to further index), so:
- my `&& indices.is_empty()` is redundant (explains why nothing broke with your version)
- you *strengthened* the condition for using `base_pointee_ty`, meaning `final_spirv_ptr_type` is used even in cases where it's incorrect to do so (and a cast should have been used instead)
  - generating the cast to `final_spirv_ptr_type` doesn't help, which means... what exactly? that two wrongs make a right?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-                    }
-                });
-            base_ptr_and_combined_indices
+            // Search the current function's instructions...
+            // FIXME(eddyb) this could get ridiculously expensive, at the very least
+            // it could use `.rev()`, hoping the base pointer was recently defined?
+            let search_result = {
+                let emit = self.emit();
+                let module = emit.module_ref();
+                emit.selected_function().and_then(|func_idx| {
```
eddyb, line 870: This accounts for `selected_function()` being `None` - that should never be possible when you are in a method of a builder that will emit instructions into a selected *block*, not just function.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -540,190 +578,434 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-                original_indices,
-                is_inbounds,
-            );
+            trace!("has original access chain, attempting to merge GEPs");
+
+            // Check if merging is possible. Requires:
+            // 1. The original AccessChain had at least one index.
+            // 2. The *last* index of the original AccessChain is a constant.
+            // 3. The *first* index (`ptr_base_index`) of the *current* GEP is a constant.
+            // Merging usually involves adding these two constant indices.
```
eddyb, line 910: Wait, you replaced a potentially-runtime addition which *allows array indexing GEPs to merge*, with a constant-only addition?

How could that help? If they were constants, the outcome is unchanged, you're still using `add`, and if they were not constants, not merging now means it hits the worst path and I can't think of any reason why that would result in better recovery downstream.

---

Changing this back to what it was before doesn't impact tests, btw, so that's another instance of that.

I am almost certain now that the weird `!is_byte_gep || ...` condition is the only actual change in this PR that impacted tests, and that's the one I don't understand yet.

eddyb (commented):
I was late on this and I'm not even sure I should leave this review.

I've got mixed feelings about it - don't mind all the `tracing` usage and the comments, but it turns out that there are fewer functional changes than I expected - in fact, just one, and _it doesn't make sense to me_.

There's a non-trivial chance this is a case of "two wrongs make a right", or more specifically _something else_ ends up reacting better to what seems to be invalid SPIR-V (AFAICT).

Now, too be fair, I finally looked at this because of:
- https://github.com/rust-lang/rust/pull/134117

And the simplest workaround for that seems to be "run the GEP merge logic twice, using a loop, and the second time use `pointercast` first which can generate its own GEP via `recover_access_chain_from_offset`" - so there's a chance the "byte GEP" stuff hits a similar weird combo.

In fact, funnily enough, I think we could've always done that `pointercast` before the merge, because of the `if ty == original_pointee_ty` that currently gates merging: if that condition holds, the `pointercast` is a noop _anyway_.
(**EDIT**: yupp, that works, and it amounts to moving just two `let`s slightly earlier in the function)

I should try to figure out the overall "decision trees" for `maybe_inbounds_gep` and other `recover_access_chain_from_offset` callers, then reduce duplicated work and handle all possible cases that can be statically resolved.

LegNeato:
@eddyb feel free to fix or back out whatever you need!

The new code flow made more sense to me and I confirmed a bunch of stuff with traces in various contexts but I am still grappling with the underlying model and how best to apply it.

Also feel free to just ignore this and focus on qptr, as it is a bit of a side quest that will be obsoleted by your work (AFAIK).

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: Yeah, the code was super hard to follow with all sort of indirection, reentry, closures, etc. I needed the tracing to just see what the heck was going on. I left it in because if I needed it to even start to understand this code, future people would.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: Yeah, again I added these as breadcrumbs for myself...I kept getting turned around.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: Oh weird, I added it specifically because byte indexing was broken in tests. Perhaps a later change fixed this and I didn't notice.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: I probably did this for logging which looks like it was removed

eddyb:
I've stumbled into a reproducer for the case I mentioned in https://github.com/Rust-GPU/rust-gpu/pull/233#discussion_r2078033290:
> you *strengthened* the condition for using `base_pointee_ty`, meaning `final_spirv_ptr_type` is used even in cases where it's incorrect to do so (and a cast should have been used instead)

<details>
<summary>(click to open the exact testcase that reproduced this problem, by @SiebenCorgie)</summary>

```rust
// Tests the simplest `TypedBuffer` case: Multiple structs of of the same type and size.

use spirv_std::{glam::UVec3, spirv, RuntimeArray, TypedBuffer};

#[derive(Clone, Copy)]
pub struct MyData {
    some_big_data: [u32; 1 << 24],
}

#[spirv(compute(threads(1)))]
pub fn compute(
    #[spirv(global_invocation_id)] global_invocation_id: UVec3,
    #[spirv(descriptor_set = 0, binding = 0, storage_buffer)] my_data: &mut RuntimeArray<
        TypedBuffer<MyData>,
    >,
) {
    let mut load_dta = unsafe { my_data.index(global_invocation_id.x as usize) }.some_big_data[0];
    load_dta = 32;

    let mut target = unsafe { &mut my_data.index_mut(global_invocation_id.y as usize) };
    target.some_big_data[0] = load_dta;
}
```
</details>

<sup>(I will try to submit those tests ASAP, I think they got lost when an outdated copy of `TypedBuffer` landed)</sup>

---

If `p: &([T; N],)` (or a struct with an array at offset `0`, which is what you can see),
then `p.0[0]` will invoke (after https://github.com/rust-lang/rust/pull/134117, which simplifies array GEPs)
GEP on `p: *([T; N],)` with type `T` and index `0` - in other words, `p.cast::<T>().add(0)`,
*but* without the cast happening separately, just being implied by the GEP itself.

Because the *array* index itself is `0`, the offset will be `0`, and so `recover_access_chain_from_offset` will succeed *in one step*, returning `[0]` and `[T; N]`, i.e.: `&(*p).0: *[T; N]` - no array indexing happens!

There's only two valid courses of action at that point:
1. accept the recovery, with the return type `*[T; N]` (this is what my logic before this PR used to do)
   - this PR seems to only be correct if the GEP *happened to be* on `i8`
     (which is what `struct` offsets use nowadays, and `[u8; N]` but those are rarer in Rust-GPU)
2. give up on the recovery, and continue onto the merge logic (and if that fail, emit the GEP anyway)

Of course, if I "fix" the logic to be what it was before, `issue-46` errors again, so I suppose it's time I understood `issue-46` myself (using the plethora of `trace!`s from this PR, yet another reason I don't want to just revert all of this, esp. after going through the trouble of reviewing it).

---

**EDIT**: a pre-`rustup` reproduction seems to have occurred in #323 (or at least it has the same fix).

eddyb:
Looks like all that was ever needed to fix #46 was this:
```rust
if const_ptr_offset_bytes == 0 {
    return self.pointercast(ptr, final_spirv_ptr_type);
}
```
As emitting any new `OpAccessChain`/`OpInBoundsAccessChain` instructions for offset `0` can easily get a pointer "stuck in a field/element" even when it's eventually used to access its original pointee type.

The only way I think this PR ever worked, was that it created a value *of the wrong pointer type* that then hit the "merge access chains" code path and got turned into something with the correct type instead.

---

I am considering "backporting"/polyfilling https://github.com/rust-lang/rust/pull/134117 on `main`, before the rustup that would naturally bring it (i.e. special-case `inbounds_gep(p, array_ty, [0, i])` and rewrite it to `inbounds_gep(p, elem_ty, [i])`, where `elem_ty` is the element type of array `array_ty`).

Then use that as a springboard to remove support for arbitrary GEPs, and keep only the `<*T>::offset(p, off)`-style `gep(p, T, &[off])` support (which is the only form `rustc_codegen_ssa` ever uses, these days).

Because that means `indices`/`gep_indices_ids` will *always* be empty, and so a `recover_access_chain_from_offset` success will *always* be used, and in general the complexity of this gnarly logic should drastically reduce.

---

**EDIT**: went with the idea described above, in:
- https://github.com/Rust-GPU/rust-gpu/pull/327

LegNeato:
Sounds good to me, though I would suggest just jamming forward if that is a shorter path.

PR #235, opened 2025-04-18
ci: update ubuntu runners to 24.04

Firestar99:
> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101


eddyb (approved):
Thanks! Does the SPIR-T repo also need this bump?

PR #239, opened 2025-04-21
Only decorate physically sized types with ArrayStride

jwollen:
Stop decorating `RuntimeArrays` of unsized types with `ArrayStride`. Fixes the `spirv-val` rule:
```
Array containing a Block or BufferBlock must not be decorated with ArrayStride
```
This already works correctly when using `qptrs`.


Firestar99 (approved):
Thanks!
just fixed the lint errors so we can merge this

LegNeato:
Thank you so much for the PR!!

jwollen:
@eddyb Out of curiosity. if we reverted this, and SPIR-T needs this info, how would it work on valid non-`rust-gpu` SPIR-V in the future?

PR #243, opened 2025-04-24
Update lockfile

LegNeato:
Just ran `cargo update`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,7 +31,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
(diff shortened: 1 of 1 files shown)


Firestar99:
I think we need to force `libm < 0.2.12` just like we did with glam

simensgreen:
> I think we need to force `libm < 0.2.12` just like we did with glam

I [tried](https://github.com/simensgreen/rust_gpu_union_issue/commit/ca5db1777764435d97203e02869044663b59ffb1) this, cargo adds another `libm` 0.2.4 to Cargo.lock, compiles both 0.2.13 and 0.2.4 and fails.

Only `libm = "=0.2.11"` works for [me](https://github.com/simensgreen/rust_gpu_union_issue/commit/7a845dafa7f6b8ad9659d4f6e287837bcf9db3ea)

But I don't know if cargo decides to add another version of `libm` in another circumstances

Issue #242, opened 2025-04-24
Rust-gpu main fails to build


simensgreen:
### Expected Behaviour
Successful compilation

### Example & Steps To Reproduce

1. clone repository https://github.com/simensgreen/rust_gpu_union_issue
2. `cargo build`

## System Info
- Rust: rustc 1.84.0-nightly (b19329a37 2024-11-21)
- OS: MacOS 15.3.2
- GPU: M2 Max

## Backtrace

<details><summary>Backtrace</summary>
<p>

```
thread 'rustc' panicked at compiler/rustc_middle/src/ty/mod.rs:1676:13:
Box<dyn Any>
stack backtrace:
   0:        0x1129ed7b0 - std::backtrace::Backtrace::create::hf5e5ba1728bb8de7
   1:        0x110844da8 - <alloc[e2e5e8515d00274d]::boxed::Box<rustc_driver_impl[c6ade677a6b0e20c]::install_ice_hook::{closure#0}> as core[924c651d484e1a25]::ops::function::Fn<(&dyn for<'a, 'b> core[924c651d484e1a25]::ops::function::Fn<(&'a std[fc297465fa958267]::panic::PanicHookInfo<'b>,), Output = ()> + core[924c651d484e1a25]::marker::Send + core[924c651d484e1a25]::marker::Sync, &std[fc297465fa958267]::panic::PanicHookInfo)>>::call
   2:        0x110844bb0 - <alloc[e2e5e8515d00274d]::boxed::Box<rustc_driver_impl[c6ade677a6b0e20c]::install_ice_hook::{closure#0}> as core[924c651d484e1a25]::ops::function::Fn<(&dyn for<'a, 'b> core[924c651d484e1a25]::ops::function::Fn<(&'a std[fc297465fa958267]::panic::PanicHookInfo<'b>,), Output = ()> + core[924c651d484e1a25]::marker::Send + core[924c651d484e1a25]::marker::Sync, &std[fc297465fa958267]::panic::PanicHookInfo)>>::call
   3:        0x112a07edc - std::panicking::rust_panic_with_hook::heb73fc7746f92ba7
   4:        0x1108e9718 - std[fc297465fa958267]::panicking::begin_panic::<rustc_errors[caf545eadef7fdd]::ExplicitBug>::{closure#0}
[...]

PR #244, opened 2025-04-26
force `libm <=0.2.11`, as `0.2.12` has breaking changes

Firestar99:
closes https://github.com/Rust-GPU/rust-gpu/pull/243

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2664,6 +2664,7 @@ dependencies = [
  "bitflags 1.3.2",
  "bytemuck",
  "glam",
+ "libm",
  "num-traits",
  "spirv-std-macros",
  "spirv-std-types",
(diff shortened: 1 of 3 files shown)


eddyb (approved):

tgross35:
Would you clarifying the problem is? 0.2.12 was not a breaking changes.

Firestar99:
Wrote you a summary in https://github.com/Rust-GPU/rust-gpu/issues/242
> libm `0.2.12+` breaks rust-gpu due to introducing new intrinsics that interact with rustc. We have a bit of a weird way to process these intrinsics and emit the corresponding SPIR-V (potentially with GLSL extensions) that needs to be updated to understand these new intrinsics. For now, we're locking libm to `0.2.11` to prevent new projects or `cargo update` from picking up newer broken versions.

nazar-pc:
Well, I really don't like the solution of constraining the version here because it is toxic for downstream users that at some point will be unable to upgrade their dependencies due to conflicts.

There is also no TODO or any explanation in the `Cargo.toml` explaining what the issue is and what it would take to lift it. Consider always leaving one with a link to issue/discussion for future reference. I also don't see a linked issue about removing this requirement in general. This can't be a long-term solution.

LegNeato:
@nazar-pc libm released a fixed version so I think we are fine now?

PR #245, opened 2025-04-27
`spirv_builder` feature `compile_codegen` to disable building `rustc_codegen_spirv`

Firestar99:
(copied from zulip)

An idea I just had: What if we make `spirv-builder` have a new `compile_codegen` feature that makes it depend on `rustc_codegen_spirv`? Without the feature, we can compile it on stable, but need to supply `SpirvBuilder` with a file path to the compiled `rustc_codegen_backend.so` file. Which allows us to use it as a direct dependency within `cargo gpu` and share the structs easily. 

Open Questions:
* Name suggestions for the feature? `compile_codegen`
* Should we make `compiled-tools` and `installed-tools` depend on this feature, preserving backwards compat, or error when they are used without `compile_codegen`? Former, preserve backwards compat

### DIFF
--- a/crates/spirv-builder/Cargo.toml
+++ b/crates/spirv-builder/Cargo.toml
@@ -18,10 +18,13 @@ no-default-features = true
 # that optional dependency, from being automatically created by Cargo, see:
 # https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies
 [features]
-# See `rustc_codegen_spirv/Cargo.toml` for details on these features.
 default = ["use-compiled-tools"]
-use-installed-tools = ["dep:rustc_codegen_spirv", "rustc_codegen_spirv?/use-installed-tools"]
-use-compiled-tools = ["dep:rustc_codegen_spirv", "rustc_codegen_spirv?/use-compiled-tools"]
+# Compile `rustc_codegen_spirv`, allows constructing SpirvBuilder without
+# explicitly passing in a path to a compiled `rustc_codegen_spirv.so` (or dll)
+compile_codegen = ["dep:rustc_codegen_spirv"]
+# See `rustc_codegen_spirv/Cargo.toml` for details on these features.
+use-installed-tools = ["compile_codegen", "rustc_codegen_spirv?/use-installed-tools"]
+use-compiled-tools = ["compile_codegen", "rustc_codegen_spirv?/use-compiled-tools"]
 skip-toolchain-check = ["rustc_codegen_spirv?/skip-toolchain-check"]
 
 watch = ["dep:notify"]
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -606,6 +606,7 @@ fn dylib_path() -> Vec<PathBuf> {
     }
 }
 
+#[cfg(feature = "compile_codegen")]
 fn find_rustc_codegen_spirv() -> PathBuf {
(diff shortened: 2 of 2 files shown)


Firestar99:
> [compile_codegen: include_str! all target_spec jsons for cargo gpu](https://github.com/Rust-GPU/rust-gpu/pull/245/commits/dcbdc524c0a5c51cd32eee9a7c0078186d683bff)

this commit made my `cargo fmt` slow to an absolute crawl, needs investigation

crates/rustc_codegen_spirv/build.rs
```diff
@@ -11,6 +11,9 @@ use std::process::{Command, ExitCode};
 use std::{env, fs, mem};
 
 /// Current `rust-toolchain.toml` file
+/// WARNING!!! cargo-gpu is new relying on this being a string literal! It will
```
schell: Quick typo here, I think it should be "cargo-gpu is _now_"...

schell:
This is looking great @Firestar99 , what's left?

Firestar99:
I'd wait and merge this PR together with https://github.com/Rust-GPU/cargo-gpu/pull/69, in case I find something that still needs adjustment.

crates/rustc_codegen_spirv/build.rs
Firestar99: fixed

schell (approved):
Great!

Firestar99:
Something I forgot to mention here: The default `MetadataPrintout` changed from `Full` to `None`, so you now have to explicitly enable it to get `cargo::rerun-if-changed="path/to/file"` messages

crates/spirv-builder/Cargo.toml
```diff
@@ -38,5 +42,9 @@ memchr = "2.4"
 raw-string = "0.3.5"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+thiserror = "2.0.12"
+semver = { version = "1.0.24", features = ["serde"] }
 
 notify = { version = "7.0", optional = true }
+# Pinning clap, as newer versions have raised min rustc version without being marked a breaking change
+clap = { version = "=4.5.37", optional = true, features = ["derive"] }
```
nazar-pc, line 50: This is really annoying, can this constraint be lifted now that the toolchain was upgraded? This prevents me from using newer clap version in the workspace downstream.

tombh: I seem to remember that [I pinned that](https://github.com/Rust-GPU/cargo-gpu/issues/39). Running `cargo-gpu` against newer `rust-gpu` versions was failing. `cargo-gpu` has tests for the last ~2 years of `rust-gpu` versions. So if you want to bump `clap`, and `cargo-gpu` tests pass then I think it's fine. And if they fail, then what @schell and I discussed is that `cargo-gpu` should aim to support the last 2 versions of `rust-gpu`. But I'm sure that could be relaxed if there's good reason.

Firestar99: Note that you no longer directly depend on old spirv-builder versions since I refactored cargo-gpu, we always use the newest one. Only the codegen backend is now compiled in a potentially quite old toolchain, which doesn't need clap.

nazar-pc: Already updated in https://github.com/Rust-GPU/rust-gpu/pull/349, everything looks good so far

PR #246, opened 2025-04-30
rustup to `nightly-2025-04-28`

LegNeato:

Issue #182, opened 2024-12-17
Tracking issue: undo `rustc_codegen_ssa` patching (aka `pqp_cg_ssa`).


eddyb:
PR that introduced `rustc_codegen_ssa` patching (aka `pqp_cg_ssa`):
- https://github.com/Rust-GPU/rust-gpu/pull/170
  (its description goes into much more detail)

Legitimate solutions are needed, to replace the patching "hack"arounds:
- [ ] (easy?) transition away from `#[repr(simd)]` for SPIR-V vector types
  - easy version: use `#[spirv(vector)]` as a replacement
    - worst part is `glam` would need to depend on `spirv-std-macros`
    - however, it does unlock `glam` using `#[spirv(matrix)]`, too!
  - hard version: actually rely on `core::simd::Simd<T, N>`
    - `glam` already has some features for this but only for some types
      (and largely as an optimization, i.e. accelerating operations)
    - `glam` needs to cast `&Simd<T, 2>` to `&struct { x: T, y: T }`
      (with `#[repr(C)]` on the latter `struct`, this is 100% defined)
    - without some hacky special-cases, `qptr` may be required for this
- [ ] (hard) support untyped function-local variables (`alloca` in LLVM terms)
  - see https://github.com/rust-lang/rust/pull/122053
    (instead of a type, creating a variable now takes only size & alignment)
  - *(note: experimental `qptr` branches can handle some of this already)*
  - typed variables are currently relied on for:
  - [ ] regular data type accesses (turning offsets into field accesses)
    - would be subsumed by `qptr` (which infers typed memory in general)
    - might be fixable pre-`qptr` by collecting types from accesses
      <sub>(however, typed GEPs have been replaced with raw byte offsets by now, so this would observe N disjoint leaves, no "field `N` of struct `S`", and borderline reimplement the `qptr` type recovery algorithm)</sub>
  - [ ] inline `asm!` `typeof*`/type inference
    - <sub>sadly needed because we can't use normal value `in`/`out` in `asm!` (without being limited to no generics, no vectors, etc.) and have to resort to passing `&T` for inputs and `&MaybeUninit<T>` for outputs - or rather, `*mut T` from the latter)</sub>
    - thankfully, `rustc_codegen_ssa` does pass the Rust type of each input, so this has a relatively simple fix (just need to open a PR for it)
  - [ ] variables containing handles (e.g. `Image`s), not data
[...]

PR #249, opened 2025-05-12
rustup: update to `nightly-2025-05-09` (~1.88) and Rust 2024 edition.

eddyb:
<sub>

*This PR supersedes (and includes many of the changes from):*
- *#246*
</sub>

---

This brings us 4 whole Rust releases ahead, to a 1.88-equivalent nightly.
<sub>(Rust 1.88.0 stable will be released on June 26th, 2025)</sub>

The many commits exist in part because of nightly bisection (to find `nightly-2024-12-16`, containing the GEP array change), but also because there have been several changes of note:
- target specs no longer spuriously claim an OS of `unknown`
  - (the sole `unknown` component in e.g. `spirv-unknown-vulkan1.1` is the *vendor*, whereas the OS is `none` and omitted in the "triple")
- `maybe_inbounds_gep` had to be made more flexible, due to:
  https://github.com/rust-lang/rust/pull/134117
  - (supporting that pattern was achieved by attempting `pointercast` first, which can itself produce a "first element of array" GEP that can then be merged with an "intra-array indexing GEP")
- `f16`/`f128` required some stubbing out (i.e. intrinsics)
  - it may be possible to support `f16`, but I've not attempted that in this PR
- had to re-enable niches for `bool` (i.e. making `Option<bool>` byte-sized), due a `transmute` (from `ControlFlow<bool>` to `u8`) added to `core::cmp` in:
  https://github.com/rust-lang/rust/pull/138881
  - not sure `Option<bool>` ever worked in Rust-GPU ([the comments I found](https://github.com/EmbarkStudios/rust-gpu/pull/809/files#r757777825) are confusing, I thought it had users and/or tests, but maybe not)
- *(the remaining ones have to do w/ `rustc_codegen_ssa` patching)*
  - *(see also: #182)*
- `rustc_codegen_spirv` had to be switched to the Rust 2024 edition (as `rustc_codegen_ssa` had done the same upstream)
- old-style `#[repr(simd)]` (using individual `struct` fields) has had even more support ripped out upstream
  - hooking/patching that support back in was possible, but it's clearly not an intended feature and we should try to move off of it sooner than later
- some dependencies of `rustc_codegen_ssa::back` (i.e. `object`) required more patching to remove their use sites
  - only needed because nightlies have had two prebuilt versions of `object`, breaking `extern crate object;` (as it may load the wrong version)
  - this was also the original blocker for #246
    (before all the other issues were discovered through testing)
  - @LegNeato has been separately working on removing that duplication upstream

While not *necessary*, this (just like #246) includes a full migration of the entire workspace to Rust 2024 (leading to lots of `unsafe {...}` in `unsafe fn`s in `spirv-std`, mainly) - we can split that out if it seems better, but I at least tried to keep it in its own commit.

---

**EDIT**: also included a fix for inline `asm!` vs untyped pointers, sadly it didn't seem to help with any failures (as the patching we already do happens to keep enough types around), but it'd be good to have just in case (esp. since the way `asm!` works means we could have hard-to-debug miscompilations).


Firestar99:
Thanks for getting us up to the newest versions! :tada: 
I'm currently away from home, so I'll likely won't be able to review until next week.

crates/rustc_codegen_spirv/build.rs
```diff
@@ -276,6 +317,12 @@ mod maybe_pqp_cg_ssa;
 
     println!("cargo::rustc-check-cfg=cfg(rustc_codegen_spirv_disable_pqp_cg_ssa)");
 
+    // HACK(eddyb) `if cfg!(llvm_enzyme)` added upstream for autodiff support.
+    println!("cargo::rustc-check-cfg=cfg(llvm_enzyme)");
```
LegNeato, line 321: Oh cool, I didn't know you could put those here.

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -151,20 +155,108 @@ pub(crate) fn provide(providers: &mut Providers) {
...
         let TyAndLayout { ty, mut layout } =
             (rustc_interface::DEFAULT_QUERY_PROVIDERS.layout_of)(tcx, key)?;
 
         #[allow(clippy::match_like_matches_macro)]
         let hide_niche = match ty.kind() {
-            ty::Bool => true,
+            ty::Bool => {
+                // HACK(eddyb) we can't bypass e.g. `Option<bool>` being a byte,
+                // due to `core` PR https://github.com/rust-lang/rust/pull/138881
+                // (which adds a new `transmute`, from `ControlFlow<bool>` to `u8`).
```
LegNeato, line 255: Should we fix / backout stuff upstream?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -854,46 +837,65 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+        // HACK(eddyb) updated pointee type of `ptr` post-`pointercast`.
+        let original_pointee_ty = ty;
+
         // --- Attempt GEP Merging Path ---
 
         // Check if the base pointer `ptr` itself was the result of a previous
         // AccessChain instruction. Merging is only attempted if the input type `ty`
         // matches the pointer's actual underlying pointee type `original_pointee_ty`.
         // If they differ, merging could be invalid.
+        // HACK(eddyb) always attempted now, because we `pointercast` first, which:
```
LegNeato: Just update the comment rather than having a new comment modify...we have git 😄

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -908,12 +910,21 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
             // 2. The *last* index of the original AccessChain is a constant.
             // 3. The *first* index (`ptr_base_index`) of the *current* GEP is a constant.
             // Merging usually involves adding these two constant indices.
+            //
+            // FIXME(eddyb) the above comment seems inaccurate, there is no reason
```
LegNeato: Again, just fix or delete the comment.

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -46,16 +46,10 @@ impl Builder<'_, '_> {
...
-            32 => (
-                self.constant_u32(self.span(), 1 << 31),
-                self.constant_u32(self.span(), u32::MAX >> 1),
-            ),
-            64 => (
-                self.constant_u64(self.span(), 1 << 63),
-                self.constant_u64(self.span(), u64::MAX >> 1),
-            ),
-            _ => bug!("copysign must have width 32 or 64, not {}", width),
+        let [mask_sign, mask_value] = {
```
LegNeato, line 49: This seems a lot more obtuse than the previous code.

LegNeato (changes requested):
Some minor comments. I liked how many of your changes are better than mine ha. Only rejecting so it doesn't merge and others can review.

crates/rustc_codegen_spirv/src/abi.rs
eddyb: Nope, it's our absurd problem that we want `bool` to have a layout `Option<bool>` can't have (cue "`qptr` will fix this").

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: To be clear, when I do this it's because I'm already several steps deep and accurately rewording comments might not be the easiest/quickest thing at that moment (but I agree I could/should fix it before landing the PR).

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
eddyb: It's bit math, it can sadly do that even when unintended (though if you prefer the formulation in `rustc_apfloat`, we could probably reuse some of their logic). Thankfully we have new tools coming to the standard library (like "unbounded shifts" IIRC?), so we could talk about powers of two more directly (without having to do right shifts of `!0`).

Anyway, the previous code had the fatal flaw of handling bit widths needlessly separately: unless they diverge in a way not parameterizable by the bit width itself, such code *should not exist in a compiler*.

(e.g. anything that relies on host `uN` instead of working with `u128` and making masks from `(1 << N) - 1` - or more accurately, `!0u128 >> (128 - N)` - is liable to diverge for various N and/or not handle all possible N and/or be annoying to update for no reason)

I welcome any suggestions for how to make the bit math more readable, but sharing a single codepath is a necessary cleanup for compiler code IMO.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: Or just delete the comment!

Firestar99:
Did a quick test of this branch and get a ton of:
```
error: OpPtrNotEqual without OpCapability VariablePointers
     --> /home/firestar99/.rustup/toolchains/nightly-2025-05-09-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/maybe_uninit.rs:619:38
      |
  619 |             ManuallyDrop::into_inner(self.value)
      |                                      ^^^^^^^^^^
      |
  note: used from within `<core::mem::maybe_uninit::MaybeUninit<&[u32]>>::assume_init`
     --> /home/firestar99/.rustup/toolchains/nightly-2025-05-09-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/maybe_uninit.rs:619:38
      |
  619 |             ManuallyDrop::into_inner(self.value)
      |                                      ^^^^^^^^^^
  note: called by `<spirv_std::typed_buffer::TypedBuffer<[u32]> as core::ops::deref::Deref>::deref`
     --> /home/firestar99/.cargo/git/checkouts/rust-gpu-d06d15e2ba0f0ae2/ea25c7e/crates/spirv-std/src/typed_buffer.rs:79:13
      |
  79  |             result_slot.assume_init()
      |             ^^^^^^^^^^^^^^^^^^^^^^^^^
  note: called by `<&rust_gpu_bindless_shaders::descriptor::descriptors::Descriptors as rust_gpu_bindless_shaders::descriptor::descriptors::DescriptorAccess<rust_gpu_bindless_shaders::descriptor::buffer::Buffer<[rust_gpu_bindless_egui_shaders::vertex::Vertex]>>>::access::<rust_gpu_bindless_shaders::descriptor::transient::Transient>`
     --> /home/firestar99/workspace/space-rust/rust-gpu-bindless/crates/shaders/src/descriptor/descriptors.rs:55:36
      |
  55  |         unsafe { BufferSlice::from_slice(self.buffers.index(desc.id().index().to_usize()), self.meta) }
      |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]
         from `*void`
           to `*[u32]`
     --> /home/firestar99/.rustup/toolchains/nightly-2025-05-09-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/maybe_uninit.rs:619:38
[...] // same backtrace
```

Also when you first update to this PR, cargo errors telling you that your rustc version is too old for edition 2024, cause it's still locked to the master nightly. I usually figure out what the new nightly is by running it, making updating a bit weird, as I first need to switch to eg. stable just to get the correct version displayed. Just something to watch out for, as other people updating may experience this over the next few months. (Except cargo gpu users, for them it should be flawless)

eddyb:
> cargo errors telling you that your rustc version is too old for edition 2024

Oh, that's cursed, I wish we could completely avoid `edition = "2024"` for any crates (and do that migration separately), but:
- how do we deal with the (eventual) release, same problem there, right?
  - (even if we do a separate release that just bumps the edition, some users might update from 0.9 to a Rust-GPU that needs Rust 2024, and get that issue anyway)
- it would require more `rustc_codegen_ssa` patching to account for the edition differences (as upstream `rustc` fully switched to Rust 2024 at some point)
  - maybe not hard, but could get finnicky and weird (IIRC pattern binding ergonomics rules got tightened so some `&`/`ref` patterns needed rewrites)

---

As for the pointer errors: if that's some straight-forward `TypedBuffer` usage, it really shouldn't be able to break like that, without also breaking tests, was it landed without tests, or are they too simple perhaps?

(If you have the patience, you could probably use some of the commits to bisect at least among the nightlies I had passing all tests, to narrow down the breakage - or I could try building your project myself, if that's simpler overall etc.)

LegNeato:
I don't think we have any users on 0.9 so I wouldn't worry about it. We can just document what is going on, and cargo-gpu will help eventually.

Firestar99:
> I don't think we have any users on 0.9...

Oh you wouldn't believe the amount of people that prefer to use a 2 year old outdated version over the master of a repo :sweat_smile:
Luckily, cargo gpu is completely unaffected by any of it! (already tested this version, just requires bumping the `spirv_builder` dependency to pick up the new target jsons)

Firestar99:
I can have a look at bisecting the error or making a minimal repo. But if you already want to take a look, changing the rev in the root workspace and `cargo c` should do it

Firestar99:
bisected it for you:
* broken: [abi: more workarounds for old-style `#[repr(simd)] struct`s.](https://github.com/Rust-GPU/rust-gpu/commit/7a0f1a26ab6e074a33699a43024cc31f8b2db639)
* does not compile: rustup: update to `nightly-2025-03-29` (~1.87).
* does not compile: rustc_codegen_spirv: switch to Rust 2024 edition.
* works: [Fix new clippy warnings.](https://github.com/Rust-GPU/rust-gpu/commit/8ab84c7bdd849998c951e90e4c6293bcfd4d40b9)

Firestar99:
cargo gpu: This is the first PR to update the target spec jsons, which causes cargo gpu to run into some trouble. The codegen backend verifies that these jsons match exactly what it expects, and there's no flag to skip that check, and adding one now wouldn't help since we want to keep compatible with at least 0.9.0. Cargo gpu was never designed to have multiple target-spec versions, it always just used the latest. Now we need to manage two different target spec json versions if we want to be compatible with before and after this change. We also can't easily extract the json from the `rustc_codegen_spirv` crate, as it's only contained in the compatible spirv-builder crate. So we end up in this weird state where we are likely forced to have per-codegen backend target specs (which is fine) and manage multiple target-spec versions internally and hopefully select the correct one depending on rustc version (not so nice).

Firestar99:
cargo gpu: This is the first PR to update the target spec jsons, which causes cargo gpu to run into some trouble. The codegen backend verifies that these jsons match exactly what it expects, and there's no flag to skip that check, and adding one now wouldn't help since we want to keep compatible with at least 0.9.0. Cargo gpu was never designed to have multiple target-spec versions, it always just used the latest. Now we need to manage two different target spec json versions if we want to be compatible with before and after this change. We also can't easily extract the json from the `rustc_codegen_spirv` crate, as it's only contained in the compatible spirv-builder crate. So we end up in this weird state where we are likely forced to have per-codegen backend target specs (which is fine) and manage multiple target-spec versions internally and hopefully select the correct one depending on rustc version (not so nice). 
cc @schell @tombh

Firestar99:
cargo gpu solution: Could we move the target jsons from `spirv_builder` to `rustc_codegen_spirv-types`? That would allow `cargo-gpu` to resolve the required target jsons from `rustc_codegen_spirv` depending on `rustc_codegen_spirv-types` (and `spirv_builder` will retain access to it). If we can't find any target jsons in `rustc_codegen_spirv-types`, we can just fall back on the current target jsons, which we'd have to ship with cargo-gpu for the forseeable future. But I think that's a fine compromise, at least we wouldn't need to ship all future target jsons as well. 

That would require merging this move of target jsons *before* this PR is merged, to not have a hole of unsupported commits on master.

I'll have a look at implementing this tomorrow.

Firestar99 (dismissed):
changing target jsons breaks cargo gpu, see https://github.com/Rust-GPU/rust-gpu/pull/249#issuecomment-2891787971
(just to prevent this from merging)

LegNeato:
@eddyb where are we at with this? Can I help with anything?

Firestar99:
I managed to find this quite minimal repo while deving on [graphite](https://github.com/GraphiteEditor/Graphite/pull/2690/files#diff-d6960f6d6338787037ef95e868b1ad4e05709a6508564b7b9ac202fe64a4ecd7R10-R11), but it **only fails outside the compiletest env**. Not sure why, needs more investigating. I'm using `index_unchecked()` a bunch in my Nanite-at-home project as well, so likely the culprit for it failing to compile with this branch.  

```rust
use glam::{Vec2, Vec4};
use spirv_std::spirv;

/// webgpu NDC is like OpenGL: (-1.0 .. 1.0, -1.0 .. 1.0, 0.0 .. 1.0)
/// https://www.w3.org/TR/webgpu/#coordinate-systems
const FULLSCREEN_VERTICES: [Vec2; 3] = [Vec2::new(-1., -1.), Vec2::new(-1., 3.), Vec2::new(3., -1.)];

#[spirv(vertex)]
pub fn fullscreen_vertex(#[spirv(vertex_index)] vertex_index: u32, #[spirv(position)] gl_position: &mut Vec4) {
	// broken on edition 2024 branch
	// let vertex = unsafe { *FULLSCREEN_VERTICES.index_unchecked(vertex_index as usize) };
	let vertex = FULLSCREEN_VERTICES[vertex_index as usize];
	*gl_position = Vec4::from((vertex, 0., 1.));
}
```

eddyb:
I have a potential culprit identified: https://github.com/Rust-GPU/rust-gpu/pull/233#issuecomment-3025933339

(Also, it took me this long to realize the "compiletest env" part of @Firestar99's comment won't stop me from playing around with that reproducer, even if outside of compiletest...)

eddyb:
_@Firestar99 wrote in https://github.com/Rust-GPU/rust-gpu/pull/249#issuecomment-2959352397_
> I managed to find this quite minimal repo while deving on [graphite](https://github.com/GraphiteEditor/Graphite/pull/2690/files#diff-d6960f6d6338787037ef95e868b1ad4e05709a6508564b7b9ac202fe64a4ecd7R10-R11), but it **only fails outside the compiletest env**. Not sure why, needs more investigating. I'm using `index_unchecked()` a bunch in my Nanite-at-home project as well, so likely the culprit for it failing to compile with this branch.
> 
> _(… code example elided …)_

Further reduction (that can be added at the end of e.g. `examples/shaders/simplest-shader/src/lib.rs`):
```rust
#[spirv(vertex)]
pub fn broken(#[spirv(vertex_index)] i: usize, #[spirv(position)] out: &mut Vec4) {
    use spirv_std::arch::IndexUnchecked as _;

    *out = unsafe { *[Vec4::ZERO; 3].index_unchecked(i) };
}
```

Doesn't seem to be impacted by the same thing I mentioned in https://github.com/Rust-GPU/rust-gpu/pull/233#issuecomment-3025933339 - so there's at least two bugs blocking this, and neither caught by CI? _oof_

eddyb:
_@Firestar99 wrote in https://github.com/Rust-GPU/rust-gpu/pull/249#issuecomment-2959352397_
> I managed to find this quite minimal repo while deving on [graphite](https://github.com/GraphiteEditor/Graphite/pull/2690/files#diff-d6960f6d6338787037ef95e868b1ad4e05709a6508564b7b9ac202fe64a4ecd7R10-R11), but it **only fails outside the compiletest env**.

Found the likely suspect: upstream added `assume_nonnull` *with a default impl*:
- https://github.com/rust-lang/rust/pull/136735

And because `assume` is an optimizer hint, *it's only applied with optimizations enabled*, so compiletest doesn't see it (IMO we should do both debug/release-mode-like compiletest runs).

eddyb:
With these 2 prerequisite PRs landed, this should be ready to go now:
- https://github.com/Rust-GPU/rust-gpu/pull/326
- https://github.com/Rust-GPU/rust-gpu/pull/327

(just needs a re-review from @LegNeato and @Firestar99 due to both having "requested changes" before)

LegNeato (approved):
🚀

PR #252, opened 2025-05-21
Fix windows builds

alphastrata:
Windows by default will use the `msvc` linker, which maxes out on 2^16-1 symbols, which the ash and wgpu examples in debug mode easily exceed.

Solution: rust-lld.exe, which _should_ be available already as it's in the llvm-tools that the `rust-toolchain.toml` will be ensuring is installed anyway.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,20 @@
 [alias]
 compiletest = "run --release -p compiletests --"
 
+
+[target.x86_64-pc-windows-msvc]
+# Using Rust's LLD linker to avoid MSVC linker limitations
+linker = "rust-lld.exe"
+rustflags = [
+    # Generic Code Sharing Control (Nightly-only)
+    # ------------------------------------------
+    # Disables cross-crate generic sharing to work around Windows PE export limits.
+    # Should prevent:
+    # - LNK1189 (too many objects)
+    # - LNK1248 (too many exports)
+    "-Zshare-generics=n", # =off is also an option, but you're going to increase binary size in doing so.
+]
+
 [target.'cfg(all())']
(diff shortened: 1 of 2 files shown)


Firestar99 (approved):
Can confirm that the **ash and wgpu examples** fail to build on windows, when build **in debug mode** (eg, `--release` works on master), and this PR fixes that. 

CI only `check`s the examples and builds the example shaders, never actually fully builds the examples. @LegNeato should we cahnge our CI to not run into this again?

(I didn't actually know you could use rust's lld linker for windows as well)

alphastrata:
> Can confirm that the **ash and wgpu examples** fail to build on windows, when build **in debug mode** (eg, `--release` works on master), and this PR fixes that.
> 
I will confirm this shortly.

> (I didn't actually know you could use rust's lld linker for windows as well)
I can't take the credit, I found it some years back in the Bevy community.

LegNeato:
@Firestar99 I think we had a other issue around here to make CI do debug as well. Only doing ---release has caused issues in the past.

alphastrata:
> Can confirm that the **ash and wgpu examples** fail to build on windows, when build **in debug mode** (eg, `--release` works on master), and this PR fixes that.
> 
> CI only `check`s the examples and builds the example shaders, never actually fully builds the examples. @LegNeato should we cahnge our CI to not run into this again?
> 
> (I didn't actually know you could use rust's lld linker for windows as well)

both ash and wgpu builds (debug) fail _without_ the `rust-lld` fix, they both wail with `LINK : fatal error LNK1189: library limit of 65535 objects exceeded␍` errors. So I think this is a good fix. @Firestar99

Issue #253, opened 2025-05-21
[BUG] Broken examples?


alphastrata:
### Expected Behaviour
Expect the examples to work per the instrcutions in the docs + the `clap` help i.e
```sh
Usage: example-runner-wgpu.exe [OPTIONS]

Options:
  -s, --shader <SHADER>       [default: Sky]
      --force-spirv-passthru
  -h, --help                  Print help
```

### Example & Steps To Reproduce
Here's a table summarizing which examples work and which don't:

| Command/Example | Works? |
|-----------------|--------|
| `cargo run --bin example-runner-wgpu ` | Yes |
| `cargo run --bin example-runner-wgpu -- -s Sky` | Yes |
| `cargo run --bin example-runner-wgpu -- -s simplest-shader` | No |
| `cargo run --bin example-runner-wgpu -- -s mouse-shader` | No |
| `cargo run --bin example-runner-wgpu -- -s compute-shader` | No |
| `cd ./examples/shaders/compute-shader && cargo run` | Yes |

If you throw a `dbg!` print on the `crate_name` defined here:
```rs
// ./examples/runners/wgpu/src/lib.rs:140

        let crate_name = match options.shader {
            RustGPUShader::Simplest => "simplest-shader",
            RustGPUShader::Sky => "sky-shader",
            RustGPUShader::Compute => "compute-shader",
            RustGPUShader::Mouse => "mouse-shader",
        };
```
The `EnumString` is working fine as in, `clap` is parsing it correctly. 
I do not have time right now to investigate why -- perhaps the compilation logic in each of the individual shader crates' is wrong? 

## System Info

 - Rust: rustc 1.84.0-nightly (b19329a37 2024-11-21) (Why do you ask for this if there's a version specified in the `rust-toolchain.toml` ?
 - OS: Windows 11 (using `rust-lld.exe` see https://github.com/Rust-GPU/rust-gpu/pull/252)
 - GPU: NVIDIA RTX TITAN (20-series)
 - SPIR-V:   SPIRV-Tools v2023.3 v2022.4-198-ge7c6084f
 - SysInfo Generic: (fastfetch, like neofetch for windows)
```txt
/////////////////  /////////////////    jer@winbox
/////////////////  /////////////////    ----------
/////////////////  /////////////////    OS: Windows 11 Pro x86_64
/////////////////  /////////////////    Host: X570 AORUS ELITE WIFI (-CF)
[...]

PR #254, opened 2025-05-21
fix: buff the help menu for the wgpu-runner

alphastrata:
Updates the help-menu on the `wgpu-runner` to avoid others making the mistake I made [here](https://github.com/Rust-GPU/rust-gpu/issues/253).

Before:
```sh
Usage: example-runner-wgpu.exe [OPTIONS]

Options:
  -s, --shader <SHADER>       [default: Sky]
      --force-spirv-passthru
  -h, --help                  Print help
```

After:
```sh
Usage: example-runner-wgpu.exe [OPTIONS]

Options:
  -s, --shader <SHADER>       which shader to run [default: sky] [possible values: simplest, sky, compute, mouse]
      --force-spirv-passthru
  -h, --help                  Print help
```

closes https://github.com/Rust-GPU/rust-gpu/issues/253

### DIFF
--- a/examples/runners/wgpu/src/lib.rs
+++ b/examples/runners/wgpu/src/lib.rs
@@ -72,7 +72,6 @@
 // #![allow()]
 
 use clap::Parser;
-use clap::builder::PossibleValue;
 use clap::ValueEnum;
 use std::borrow::Cow;
 use strum::{Display, EnumString};
--- a/examples/runners/wgpu/src/lib.rs
+++ b/examples/runners/wgpu/src/lib.rs
@@ -159,7 +159,7 @@ fn maybe_watch(
             .iter()
             .copied()
             .collect::<PathBuf>();
-        
+
         let has_debug_printf = options.force_spirv_passthru;
 
         let builder = SpirvBuilder::new(crate_path, "spirv-unknown-vulkan1.1")
--- a/examples/runners/wgpu/src/lib.rs
+++ b/examples/runners/wgpu/src/lib.rs
@@ -72,6 +72,8 @@
 // #![allow()]
 
 use clap::Parser;
+use clap::builder::PossibleValue;
+use clap::ValueEnum;
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


examples/runners/wgpu/src/lib.rs
```diff
@@ -224,7 +235,7 @@ fn maybe_watch(
 #[derive(Parser, Clone)]
 #[command()]
 pub struct Options {
-    #[arg(short, long, default_value = "Sky")]
+    #[arg(short, long, default_value = "Sky", help = RustGPUShader::help_string())]
```
Firestar99: > Shader to run [simplest, sky, compute, mouse] [default: Sky] [possible values: simplest, sky, compute, mouse]

Why implement special logic, when clap already prints all the possible variants?

alphastrata: Ohh it looks like I copy+pasta'd the same output twice, if you run the same command on main the before\after makes more sense...

alphastrata: it doesn't print all the possible variants, it prints ```  -s, --shader <SHADER>       [default: Sky]```

Firestar99: It is correct that current master does not print any. But your patch prints them **twice**:
> Shader to run **[simplest, sky, compute, mouse]** [default: Sky] **[possible values: simplest, sky, compute, mouse]**

alphastrata: ...you are absolutely correct. 🤦🏽
I will fix it within the next 12 hours or so -- or if you want to push a fix like the `fmt` that'd be faster if not too much of an imposition.

Firestar99: will do

Firestar99 (approved):

Firestar99:
fixing the `output filename collision` warning on windows without breaking android is not trivial, let's leave it as is for now

PR #255, opened 2025-05-21
output filename collision fix

Firestar99:


Firestar99:
unfortunately, passing `--crate-type cdylib` doesn't work

PR #256, opened 2025-05-21
Updating target json support

Firestar99:
> [This is the first PR](https://github.com/Rust-GPU/rust-gpu/pull/249) to update the target spec jsons, which causes cargo gpu to run into some trouble. The codegen backend verifies that these jsons match exactly what it expects, and there's no flag to skip that check, and adding one now wouldn't help since we want to keep compatible with at least 0.9.0. Cargo gpu was never designed to have multiple target-spec versions, it always just used the latest. [...]

https://github.com/Rust-GPU/rust-gpu/pull/249#issuecomment-2891509750

> Could we move the target jsons from `spirv_builder` to `rustc_codegen_spirv-types`? That would allow `cargo-gpu` to resolve the required target jsons from `rustc_codegen_spirv` depending on `rustc_codegen_spirv-types` (and `spirv_builder` will retain access to it). If we can't find any target jsons in `rustc_codegen_spirv-types`, we can just fall back on the current target jsons, which we'd have to ship with cargo-gpu for the forseeable future. But I think that's a fine compromise, at least we wouldn't need to ship all future target jsons as well. 

https://github.com/Rust-GPU/rust-gpu/pull/249#issuecomment-2891787971


LegNeato:
Should this be its own crate? @eddyb and I discussed it in the past.

Firestar99:
You mean just the target specs json files as a separate crate? And have cargo gpu to just permanently depend on the "legacy" version of the target specs? (legacy meaning before this PR)

Just one issue with that: `spirv-builder` wants the path to the json files so it can point rustc to them, which is ensured to exist by directly depending on that crate. But in cargo gpu the legacy target jsons are `include_str!` into the binary, so we don't depend on those legacy json files remaining within cargo's crate cache. Which may be reset independently of the cargo gpu installation.

LegNeato:
Yeah the japn as their own crate. Doing so let's us have multiple versions and doesn't belring on other deps if something just needs the json. As AI said we discussed it but I think just kicked the can down the road until problems cropped up. We may now be there ha

LegNeato:
I'd like @eddyb to chime in before approving.

Firestar99:
@LegNeato it would be nice to have `rustc_codegen_spirv_target_spec` published on crates.io after this PR merges. Cargo gpu needs a permanent version-fixed dependency on the result of this PR to support the "legacy" rust-gpu versions before this PR. And I'd like that to be a crates.io version, not a git rev.

Cargo.toml
```diff
@@ -46,6 +47,7 @@ spirv-std-macros = { path = "./crates/spirv-std/macros", version = "=0.9.0" }
 spirv-tools = { version = "0.11", default-features = false }
 rustc_codegen_spirv = { path = "./crates/rustc_codegen_spirv", version = "=0.9.0", default-features = false }
 rustc_codegen_spirv-types = { path = "./crates/rustc_codegen_spirv-types", version = "=0.9.0" }
+rustc_codegen_spirv_target_specs = { path = "./crates/rustc_codegen_spirv_target_specs", version = "=0.9.0" }
```
eddyb: Seeing these side by side made it feel off but it took me a bit to understand what was going on.
Can you rename it to `rustc_codegen_spirv-target-specs`?
(That is, `rustc_codegen_X` follows a `rustc_*`-specific convention, whereas `-` is the more common connector in crate names, and so `rustc_codegen_X-Y` makes it clearer this is helper `Y` for codegen backend `X`)

Firestar99: will do, should just be some string replacement here and in cargo gpu

Firestar99: Once this **AND** https://github.com/Rust-GPU/cargo-gpu/pull/75 checks succeed, the rename should have been successful

Firestar99: @eddyb everything green, so feel free to approve and we'll get this merged

crates/spirv-builder/Cargo.toml
```diff
@@ -22,21 +22,23 @@ default = ["use-compiled-tools"]
 # Compile `rustc_codegen_spirv`, allows constructing SpirvBuilder without
 # explicitly passing in a path to a compiled `rustc_codegen_spirv.so` (or dll)
 rustc_codegen_spirv = ["dep:rustc_codegen_spirv"]
+# Inclide target spec json files, allows constructing SpirvBuilder without
+# explicitly passing a path to the target spec json
+include_target_specs = ["dep:rustc_codegen_spirv-target-specs"]
```
eddyb: I Ctrl+F'd `_target_specs` and found this one, sorry I missed it the first time.
```suggestion
include-target-specs = ["dep:rustc_codegen_spirv-target-specs"]
```

eddyb (changes requested):

docs/src/writing-shader-crates.md
```diff
@@ -120,7 +120,7 @@ how to build SPIR-V. Here are a few things we need to mention there.
 
 - Path to a spec of a target you're compiling for (see [platform support](./platform-support.md)).
   These specs reside in a directory inside the `spirv-builder` crate and an example relative path
-  could look like `../rust-gpu/crates/spirv-builder/target-specs/spirv-unknown-spv1.3.json`.
+  could look like `../rust-gpu/crates/rustc_codegen_spirv-types/target-specs/spirv-unknown-spv1.3.json`.
```
eddyb:
```suggestion
  could look like `../rust-gpu/crates/rustc_codegen_spirv-target-specs/target-specs/spirv-unknown-spv1.3.json`.
```

eddyb (changes requested):

PR #264, opened 2025-05-28
Generic compile result

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/256

This enables us to emit wgsl and naga modules the same way as we do spv.
Needed by https://github.com/Rust-GPU/cargo-gpu/pull/76


Firestar99:
(Will be) superseded by https://github.com/Rust-GPU/rust-gpu/pull/280

PR #268, opened 2025-06-03
Fix inverted conditions in count_(leading|trailing)_zeros for 64-bit …

LegNeato:
…integers

Selecting between lower and upper 32-bit parts were swapped.


Firestar99 (approved):
I'm just gonna trust you on thus being swapped. I assume the asserts in the **compile** tests are cause you copied them from somewhere external for testing?

Issue #241, opened 2025-04-23
Union cast failed


simensgreen:
### Expected Behaviour
Can cast `[f32; 5]` as Data `{ a: f32, b: [f32; 3], c: f32 }`

### Example & Steps To Reproduce
```rust
#[repr(C)]
#[derive(Clone, Copy)]
struct Data {
    a: f32,
    b: [f32; 3],
    c: f32,
}

union DataOrArray {
    arr: [f32; 5],
    str: Data,
}

impl DataOrArray {
    fn arr(self) -> [f32; 5] { unsafe { self.arr } }
    fn new(arr: [f32; 5]) -> Self { Self { arr } }
}

```
in shader:

```rust
let dora = DataOrArray::new([0.0, 0.0, 0.0, 0.0, 0.0]);
let arr = dora.arr();
```

Produces: 

```
error: cannot cast between pointer types
         from `*struct types::Data { a: f32, b: [f32; 3], c: f32 }`
           to `*[f32; 5]`
```

## System Info
- Rust: rustc 1.84.0-nightly (b19329a37 2024-11-21)
- OS: MacOS 15.3.2
- GPU: M2 Max
- rust-gpu: 698f10ac14b7c952394ac5620004e4e973308902

LegNeato:
Can you try latest master? I landed https://github.com/Rust-GPU/rust-gpu/commit/2f915a09f1cc8cbdced277dc6ccf4c5bcb5de6ee which fixed issues with a similar message (https://github.com/Rust-GPU/rust-gpu/issues/46).

simensgreen:
Nope, even at commit 2f915a09f1cc8cbdced277dc6ccf4c5bcb5de6ee error still appears

simensgreen:
I tried trick it with `core::mem::transmute`:

```rust
    pub fn arr(self) -> [f32; 5] {
        unsafe { core::mem::transmute(self) }
    }
```

Error:

```
  error: cannot memcpy dynamically sized data
    --> solver/src/types/data_or_array.rs:16:18
     |
  16 |         unsafe { core::mem::transmute(self) }
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

LegNeato:
Ok great, I'll take a look.

simensgreen:
@LegNeato Perhaps I can assist you with the QPTR, but I require further information. Could you please elaborate on what it is and from where it originates?

nazar-pc:
[...]

PR #269, opened 2025-06-03
Fix union field access by representing unions as structs in SPIR-V

LegNeato:
When a union has multiple fields, represent it as a struct with all fields at offset 0 instead of just using the largest field. This allows pointer casts between union fields to work correctly by enabling the recover_access_chain_from_offset function to find valid access chains.

Fixes https://github.com/Rust-GPU/rust-gpu/issues/241.


LegNeato:
@simensgreen can you check this fix?

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -718,21 +718,27 @@ fn trans_aggregate<'tcx>(cx: &CodegenCx<'tcx>, span: Span, ty: TyAndLayout<'tcx>
...
+                    .map(|i| ty.field(cx, i))
+                    .max_by_key(|case| case.size);
+
+                if let Some(case) = largest_case {
+                    assert_eq!(ty.size, case.size);
+                    case.spirv_type(span, cx)
+                } else {
+                    assert_eq!(ty.size, Size::ZERO);
+                    create_zst(cx, span, ty)
+                }
```
Firestar99, line 728: Why not use a match?
```rust
match ty.fields.count() {
    0 => {
        assert_eq!(ty.size, Size::ZERO);
        create_zst(cx, span, ty)
    },
    1 => {
        assert_eq!(ty.size, ty.field(cx, 0).size);
        case.spirv_type(span, cx)
    },
    _ => trans_struct(cx, span, ty),
}
```

LegNeato: Because I didn't think of it, good call!

Firestar99 (approved):

simensgreen:
@LegNeato Ok, in this simple example it works, but in a more complicated scenario, it will not.

Now I can reproduce both `cannot cast between pointer types` and `cannot memcpy dynamically sized data`, you can check it out in my [repo](https://github.com/simensgreen/rust_gpu_union_issue)

Full error messages: 
```
         from `*struct DataMatrix { array1d: [f32; 25], array2d: [[f32; 5]; 5], data_array: [struct Data { a: f32, b: [f32; 3], c: f32 }; 5] }`
           to `*[[f32; 5]; 5]`
    --> shader/src/lib.rs:35:77
     |
  35 |     pub fn array2d_ref(&self) -> &[[f32; 5]; 5] { unsafe { &self.array2d } }
     |                                                                             ^
     |
  note: used from within `<shader::DataMatrix>::array2d_ref`
    --> shader/src/lib.rs:35:77
     |
  35 |     pub fn array2d_ref(&self) -> &[[f32; 5]; 5] { unsafe { &self.array2d } }
     |                                                                             ^
  note: called by `shader::main`
    --> shader/src/lib.rs:50:18
     |
  50 |     output[id] = input_matrices[id].array2d_ref()[0];
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ```
  
  ```
  error: cannot memcpy dynamically sized data
    --> shader/src/lib.rs:34:54
     |
  34 |     pub fn array2d(self) -> [[f32; 5]; 5] { unsafe { self.array2d } }
     |                                                      ^^^^^^^^^^^^
     |
  note: used from within `<shader::DataMatrix>::array2d`
    --> shader/src/lib.rs:34:54
     |
  34 |     pub fn array2d(self) -> [[f32; 5]; 5] { unsafe { self.array2d } }
     |                                                      ^^^^^^^^^^^^
  note: called by `shader::main`
    --> shader/src/lib.rs:52:18
     |
  52 |     output[id] = input_matrices[id].array2d()[0];
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

LegNeato:
Ok, this code is a minefield...back to the drawingboard ha.

PR #258, opened 2025-05-22
Update mac target in CI to aarch64-apple-darwin.

LegNeato:
The `macOS-latest` images on GitHub are arm:

```
Current runner version: '2.324.0'
Operating System
  macOS
  14.7.5
  23H527
Runner Image
  Image: macos-14-arm64
```

PR #271, opened 2025-06-04
add targets `vulkan1.3`, `vulkan1.4` and `spv1.6`

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/279, https://github.com/Rust-GPU/spirv-tools-rs/pull/11

~~I thought it'll be easy going, but it seems like spv1.6 introduced some new validation we're now failing.~~ Nope, just some tests excluded for vulkan that need their target list extended.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,7 +134,7 @@ jobs:
       - name: cargo fetch --locked
         run: cargo fetch --locked --target $TARGET
       - name: compiletest
-        run: cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4
+        run: cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4,spv1.3
 
   difftest:
     name: Difftest
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,7 +134,7 @@ jobs:
       - name: cargo fetch --locked
         run: cargo fetch --locked --target $TARGET
       - name: compiletest
-        run: cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,spv1.3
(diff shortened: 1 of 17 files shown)


Firestar99:
Compiletest fails due to the installed tools being too old and not knowing what `vulkan1.4` is.

Current CI downloads a pre-build version of spirv tools from [Embark's spirv-tools repo](https://github.com/Rust-GPU/rust-gpu/blob/main/.github/install-spirv-tools/src/main.rs#L39), specifically [release 0.10.0](https://github.com/Rust-GPU/rust-gpu/blob/main/.github/install-spirv-tools/src/main.rs#L39), which includes some [pre-build `*.tar.zstd` blobs](https://github.com/EmbarkStudios/spirv-tools-rs/releases). Supposedly I could just trigger the ["Precompile spirv-tools binaries" Action](https://github.com/Rust-GPU/spirv-tools-rs/actions/workflows/precompile.yml) but that one has been broken likely by basel updating something. @LegNeato tried fixing it on master, but no luck so far.

Some Options we have to fix this:
1. fixup bazal build and publish a new set of pre-builds
2. grab the binaries from lunarg, the official source and get rid of bazel
3. derp some blobs together on my local machine, but I can't do Mac
4. recompile on CI every time, too slow IMO

I'm considering trying out 2 tomorrow, or do you @LegNeato have a better idea?

This also unblocks https://github.com/Rust-GPU/rust-gpu/pull/258

Firestar99:
@LegNeato  This is effectively ready to review, it just requires https://github.com/Rust-GPU/rust-gpu/pull/279 and spirv-tools-rs PR https://github.com/Rust-GPU/spirv-tools-rs/pull/11 to be merged and released on crates.

.github/workflows/ci.yaml
```diff
@@ -151,7 +151,7 @@ jobs:
       - name: cargo fetch --locked
         run: cargo fetch --locked --target ${{ matrix.target }}
       - name: compiletest
-        run: cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,spv1.3
+        run: cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4
```
LegNeato: I am not sure I understand the ramifications of changing this so I'll just trust you

Firestar99: yea, I should add `spv1.3` back for the weird quad texture tests, though I think in practice everyone uses vulkan targets anyway

Issue #266, opened 2025-06-03
Workgroup doesn't work in MoltenVk


redain:
### Expected Behaviour
I should be able to use workgroups in compute shaders.

### Example & Steps To Reproduce

This is the most simplified example I could make to reproduce.
```rust
#[spirv(compute(threads(256)))]
    pub fn test(
        #[spirv(local_invocation_id)] lid: UVec3,
        #[spirv(workgroup_id)] gid: UVec3,
        #[spirv(workgroup)] shared: &mut [u32; 32],
        #[spirv(push_constant)] pc: &PushConstants,
        #[spirv(storage_buffer, descriptor_set = 0, binding = 0)] all_bufs: &mut RuntimeArray<
            TypedBuffer<[u32]>,
        >,
    ) {
        shared[0] = 1;
        shared[1] = 2;
        let a = Vec2::new(shared[0] as f32, shared[1] as f32);
    }
```
produces this runtime Vulkan error:
```
vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
Invalid explicit layout decorations on type for operand '17[%17]'
  %shared = OpVariable %_ptr_Workgroup__arr_uint_uint_32 Workgroup
The Vulkan spec states: All variables must have valid explicit layout decorations as described in Shader Interfaces (https://vulkan.lunarg.com/doc/view/1.4.313.0/mac/antora/spec/latest/appendices/spirvenv.html#VUID-StandaloneSpirv-None-10684)    
```
spir-v snippet:
```
 OpCapability Shader
               OpCapability Int64
               OpCapability StorageImageExtendedFormats
               OpCapability ImageQuery
               OpCapability StorageImageWriteWithoutFormat
               OpCapability RuntimeDescriptorArray
               OpExtension "SPV_EXT_descriptor_indexing"
               OpExtension "SPV_KHR_storage_buffer_storage_class"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %1 "test" %lid %gid
               OpExecutionMode %1 LocalSize 256 1 1
  OpName %lid "lid"
               OpName %gid "gid"
               OpName %shared "shared"
               OpDecorate %lid BuiltIn LocalInvocationId
               OpDecorate %gid BuiltIn WorkgroupId
               OpDecorate %_arr_uint_uint_32 ArrayStride 4
       %void = OpTypeVoid
          %8 = OpTypeFunction %void
       %uint = OpTypeInt 32 0
[...]

PR #273, opened 2025-06-07
fix: array stride validation errors

redain:
Fix for [https://github.com/Rust-GPU/rust-gpu/issues/266](https://github.com/Rust-GPU/rust-gpu/issues/266).
Based on this new SPIR-V validation [https://github.com/KhronosGroup/SPIRV-Tools/pull/6012/files](https://github.com/KhronosGroup/SPIRV-Tools/pull/6012/files)
Can compile compute shader with work groups again.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/array_stride_fixer.rs
+++ b/crates/rustc_codegen_spirv/src/linker/array_stride_fixer.rs
@@ -1,0 +1,612 @@
+//! Fix ArrayStride decorations for newer SPIR-V versions.
+//!
+//! Newer SPIR-V versions forbid explicit layouts (ArrayStride decorations) in certain
+//! storage classes (Function, Private, Workgroup), but allow them in others
+//! (StorageBuffer, Uniform). This module removes ArrayStride decorations from
+//! array types that are used in contexts where they're forbidden.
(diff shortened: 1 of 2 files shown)


Firestar99:
I saw you wrote some "normal" tests for your linker step, those would potentially make more sense within compiletests. As the current docs are pretty bad, I wrote entirely [new docs for compiletests](https://github.com/Rust-GPU/rust-gpu/blob/fe7bbd1c824698bace6464a8dff883c127255478/tests/compiletests/README.md) in PR https://github.com/Rust-GPU/rust-gpu/pull/275, which will hopefully help in debugging and fixing this issue. Let me know if you have any further questions after reading them or need other kinds of support fixing this issue. :D

redain:
I am not sure when I will be able to finish this to make all tests pass. There are some corner cases that are still failing, I'm currently just using my branch to compile compute shaders with workgroups for my own use 😅. The most challenging part is that for mixed usage (when same array type is used for both layout-required and layout-prohibited cases) I had to split array type into 2 and then update a lot of type references in SPIR-V and this gets really tricky and error-prone. If anyone wants to take control or fix this themselves - feel free to.

PR #279, opened 2025-06-11
ci: replace spirv-tools prebuilds with vulkan sdk from lunarg

Firestar99:
While caching is enabled, it seems like github runners don't care, so we got to redownload the sdk every time.

This even speeds up CI:
* windows: 1m19 -> 7s
* linux: 32s -> 23s
* mac: 35s -> 32s


LegNeato (approved):
Sweet!

.github/workflows/ci.yaml
```diff
@@ -30,12 +30,11 @@ jobs:
...
-      # of spirv-tools. This downloads the tarball, decompresses it, unpacks
-      # the binaries to the specified path, and adds them to PATH
-      - name: Install spirv-tools binaries
-        shell: bash
-        run: cargo run --manifest-path .github/install-spirv-tools/Cargo.toml -- ${{matrix.host || matrix.target}} 0.10.0 "${{github.workspace}}/bin"
+      - name: Install Vulkan SDK
+        uses: humbletim/install-vulkan-sdk@v1.2
+        with:
+          version: 1.4.309.0
+          cache: true
```
eddyb, line 37: Hah, this seems to be [the same thing SPIR-T uses](https://github.com/Rust-GPU/spirt/blob/c7a9f61875f991df6b55939c95d7436c1a60b174/.github/workflows/ci.yml#L62-L68), I found a bit awkward sometimes due to the build times, but I suppose if you couldn't get caching to work... hmm

.github/workflows/ci.yaml
```diff
@@ -30,12 +30,11 @@ jobs:
       RUSTUP_IO_THREADS: "1"
     steps:
       - uses: actions/checkout@v4
-      # Install the spirv-tools binaries from tarballs hosted on each release
-      # of spirv-tools. This downloads the tarball, decompresses it, unpacks
-      # the binaries to the specified path, and adds them to PATH
-      - name: Install spirv-tools binaries
-        shell: bash
-        run: cargo run --manifest-path .github/install-spirv-tools/Cargo.toml -- ${{matrix.host || matrix.target}} 0.10.0 "${{github.workspace}}/bin"
```
eddyb, line 38: Do we know if `"${{github.workspace}}/bin"` is enough to induce caching?
Looking at [the docs](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows), it seems like explicitly using `actions/cache` may have been needed to get any caching?
(and we don't seem to use `actions/cache` anywhere atm)

It might be worth trying to cache `$RUSTUP_HOME` (`~/.rustup` unless it's explicitly set ig?) and even `target/{debug,release}/incremental` but that could become sketchy territory (ideally only `main` branch CI runs could save such caches, and then PR CI runs would use the latest one? unsure if that's doable tho).

.github/workflows/ci.yaml
Firestar99: Interestingly, this page shows caches: https://github.com/Rust-GPU/rust-gpu/actions/caches

But it doesn't reduce CI time, the install action is run regardless

Firestar99: > cache (optional; default=false): boolean indicating whether to **cache the downloaded installer file** between builds.

It helps to read... it reinstalls the vulkan sdk every time, which is why it's "slow". And looking at their action file, it's using:
* windows (8s): 7z which is multithreaded by default
* linux (25s): single-threaded tar with zstd
* mac (39s): just calls the installer

https://github.com/humbletim/install-vulkan-sdk/blob/30ba978f977e81b72d091fc8888feb1fb26f9aff/vulkan_prebuilt_helpers.sh#L42-L77

PR #280, opened 2025-06-11
add `spirv-unknown-naga-wgsl` target via naga

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/309 https://github.com/Rust-GPU/rust-gpu/pull/310 https://github.com/Rust-GPU/rust-gpu/pull/311

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -129,9 +129,17 @@ jobs:
   compiletest:
     name: Compiletest
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-24.04, windows-2022, macOS-latest ]
+        target_env: [ "vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4" ]
+        experimental: [ false ]
+        include:
+          - os: ubuntu-24.04
+            target_env: naga-wgsl
+            experimental: true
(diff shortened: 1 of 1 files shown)


LegNeato:
I think this is fine and later when we change it to have something generic like "gpu" we can document it / support the old style as well. Do you think instead it should be `spirv-[unknown-]naga-wgsl` or `spirv-[unknown-]wgpu-wgsl`? Probably not but wanted to check in case you hadn't thought of those options and they are more appropriate. I doubt we'd have another or "pure" wgsl backend for a long time.

Firestar99:
I actually think we should add `naga` to the target, since we could be having a `*-naga-glsl` and a `*-cross-glsl` target via `spirv-cross`

Firestar99:
It has been renamed to `spirv-unknown-naga-wgsl` and should be quite usable already. There's still a bunch of failing compiletests due to naga not parsing some spv structure properly, they should either be ignored or checked upon before merging. May just opt for ignoring them all and fixing them up at a later date.

Nice side effect of naming it `naga-wgsl`: compiletest can be ignored with `ignore-naga` so any future naga targets will also be ignored.

Firestar99:
Superseded by https://github.com/Rust-GPU/rust-gpu/pull/493

PR #281, opened 2025-06-12
Add query_size_lod and query_size methods to SampledImage

LegNeato:
- Add query_size_lod for non-multisampled sampled images (Sampled=1)
- Add query_size for multisampled sampled images
- Add missing HasQuerySize trait implementations for Sampled::Yes
- Add compiletests for both methods

Fixes https://github.com/Rust-GPU/rust-gpu/issues/190

PR #282, opened 2025-06-12
Fix build

LegNeato:
fixup for https://github.com/Rust-GPU/rust-gpu/pull/281


LegNeato:
Fake news haha! They must have a race condition.

Issue #284, opened 2025-06-15
Fragment shader can be silently compiled out


LegNeato:
This works:

```rust
#[spirv(fragment)]
pub fn main_fs(
    in_color: Vec3,
    out_frag_color: &mut Vec4,
) {
    *out_frag_color = vec4(in_color.x, in_color.y, in_color.z, 1.0);
}
```

This silently does not output a fragment shader:

```rust
    out_frag_color.x = in_color.x;
    out_frag_color.y = in_color.y;
    out_frag_color.z = in_color.z;
```

Rather than failing silently, it should probably be an error when an entire shader is dead code / compiled out and does not produce an artifact.

Firestar99:
Another shader that does not emit anything, potentially for very different reasons:
```rust
#[spirv(fragment)]
pub fn main_fs(
    in_color: Vec4,
    out_frag_color: &mut Vec4,
) {
    *out_frag_color = in_color;
}
```

It wouldn't surprise me if that's rustc optimizing out the function before the codegen backend get's to see it.

dphfox:
Encountered this myself. For any future people who stumble upon this, you can get around this by consuming `frag_coord` or something else bogus like that.

<img width="1163" height="678" alt="Image" src="https://github.com/user-attachments/assets/442a99fe-cc08-4ad2-8a26-95b6e65e1b8b" />

PR #289, opened 2025-06-19
Fixing Fragment Shader Silent Optimization Issue (Issue #284)

zmr-233:
# Fixing Fragment Shader Silent Optimization Issue (Issue #284)

This PR addresses Issue #284, where the fragment shader successfully compiles but is silently optimized to have no output operations, leading to silent rendering without errors or warnings, making debugging very difficult.

#### ✅ Use `--allow-fragment-no-output` to optionally skip ERROR

> After adding this feature, many of the previous test cases failed in Github CI because they were completely optimized out by DCE (Dead Code Elimination). After analysis, marking fragment shaders with no output as errors indeed leads to unnecessary test failures. Based on the maintainer's suggestion, a compiler flag was added to control this check. You can optionally skip the error by using `--allow-fragment-no-output`.

## Root Cause Analysis

After a thorough analysis of DCE, we found that Issue #284 is actually a "false scenario," based on the following two core facts:

1. **Entry-point functions are never deleted**: In the current DCE implementation, all functions listed in the module's OpEntryPoint are unconditionally marked as "root," making them reachable rooted objects. Even if all instructions in the function body are identified as pure calculations and removed, the function definition and OpEntryPoint declaration are retained. DCE does not remove the entire entry-point function.

2. **The "silently compiling out" phenomenon is actually the simplification of the function body**: When no "obvious output" is written, DCE removes internal computations (such as OpCompositeConstruct, temporary OpLoad, etc.) and leaves only a direct OpReturn. However, it does not remove `%main_fs = OpFunction...` or `OpEntryPoint Fragment %main_fs "main_fs" ...`. In the SPIR-V module generated by the compiler, the fragment shader still exists—it just no longer writes to the Output variable.

Thus:

* The phenomenon in Issue #284, where "the entire fragment shader is silently compiled out," is a misunderstanding of the behavior: "function body is simplified to not write outputs" → "no color is written."

* The initially proposed "silent compile-out" didn’t actually happen: The SPIR-V still contains empty functions and entry-point declarations; it's just that nothing is rendered during the rendering phase.

### Part A — Add Error for Fragment Shaders with "Empty Output"

To fix this, we need to add a check/error for fragment shaders with "empty outputs," so users can immediately know: "Hey, this fragment shader doesn't write anything to the output, it’s definitely a bug," rather than generating an invalid SPIR-V that "renders nothing."

Therefore, I added a validation step after DCE to check for fragment shaders with no output operations:

* `validate_fragment_shader_outputs()`: The main validation function called after DCE.
* `fragment_shader_writes_output()`: Recursively checks for output operations.
* `has_partial_output_writes()`: Detects evidence of partial writes that were optimized away.
* `is_output_storage_variable()`: Identifies output variables in different scopes.

### Part B — Fix `CodegenArgs::from_session` Comma-Separated Parameter Parsing Issue

During testing, we also discovered that `CodegenArgs` was incorrectly parsing comma-separated parameters, such as:

```rust
// compile-flags: -C llvm-args=--disassemble-entry=main,--allow-fragment-no-output
```

So, the code at `crates/rustc_codegen_spirv/src/codegen_cx/mod.rs:340` was fixed to:

```rust
let expanded_args: Vec<String> = sess.opts.cg.llvm_args
    .iter()
    .flat_map(|arg| arg.split(',').map(|s| s.to_string()))
    .collect();
```

### Part C — Add `--allow-fragment-no-output` to Optionally Skip ERROR

All relevant test cases involving "no output fragment shaders" have now been marked with this flag. Theoretically, they should pass in Github CI. However, there are still two failing tests locally (which appear unrelated to my changes):

* `ui/spirv-attr/invalid-target.rs`: JSON output too large, causing parsing errors (pre-existing issue).
* `ui/spirv-attr/all-builtins.rs`: SPIR-V validator strictness issue (environment-related).

## Example Error Message

I added a test case that will always fail to verify that silent failures trigger an error message:

* `tests/compiletests/ui/dis/issue-284-dead-fragment.rs`: A fragment shader with no output (correctly fails and provides a clear error).

### After the fix (clear error and guidance):

```
error: fragment shader `main_fs` produces no output
  |
  = help: fragment shaders must write to output parameters to produce visible results
  = note: use complete assignment like `*out_frag_color = vec4(r, g, b, a)` instead of partial component assignments
  = note: partial component assignments may be optimized away if not all components are written
  = note: to disable this validation (e.g. for testing), use `--allow-fragment-no-output`
```

Since the validation only runs after DCE, does not impact the hot compilation path, and only checks fragment shaders (not vertex/compute shaders), its performance impact is minimal.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -14,10 +14,6 @@ rustflags = [
     "-Zshare-generics=n", # =off is also an option, but you're going to increase binary size in doing so.
 ]
 
-# Override parent directory's cranelift backend setting for rust-gpu compatibility
-[profile.dev]
-codegen-backend = "llvm"
-
 [target.'cfg(all())']
 rustflags = [
     # FIXME(eddyb) update/review these lints.
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -14,6 +14,10 @@ rustflags = [
     "-Zshare-generics=n", # =off is also an option, but you're going to increase binary size in doing so.
 ]
 
+# Override parent directory's cranelift backend setting for rust-gpu compatibility
+[profile.dev]
+codegen-backend = "llvm"
+
 [target.'cfg(all())']
 rustflags = [
     # FIXME(eddyb) update/review these lints.
--- a/.github/workflows/lint.sh
+++ b/.github/workflows/lint.sh
@@ -49,15 +49,15 @@ clippy_no_features examples/shaders/simplest-shader
 # which could be disastrous because env vars access can't be tracked by
 # `rustc`, unlike its CLI flags (which are integrated with incremental).
 if (
-    egrep -r '::\s*env|env\s*::' crates/rustc_codegen_spirv/src |
+    grep -E -r '::\s*env|env\s*::' crates/rustc_codegen_spirv/src |
     # HACK(eddyb) exclude the one place in `rustc_codegen_spirv`
     # needing access to an env var (only for codegen args `--help`).
-    egrep -v '^crates/rustc_codegen_spirv/src/codegen_cx/mod.rs:            let help_flag_comes_from_spirv_builder_env_var = std::env::var\(spirv_builder_env_var\)$' |
+    grep -E -v '^crates/rustc_codegen_spirv/src/codegen_cx/mod.rs:            let help_flag_comes_from_spirv_builder_env_var = std::env::var\(spirv_builder_env_var\)$' |
     # HACK(LegNeato) exclude logging. This mirrors `rustc` (`RUSTC_LOG`) and 
     #`rustdoc` (`RUSTDOC_LOG`).
     # There is not a risk of this being disastrous as it does not change the build settings.
-    egrep -v '^crates/rustc_codegen_spirv/src/lib.rs:.*(RUSTGPU_LOG|RUSTGPU_LOG_FORMAT|RUSTGPU_LOG_COLOR).*$' |
-    egrep -v '^crates/rustc_codegen_spirv/src/lib.rs:    use std::env::{self, VarError};$'
+    grep -E -v '^crates/rustc_codegen_spirv/src/lib.rs:.*(RUSTGPU_LOG|RUSTGPU_LOG_FORMAT|RUSTGPU_LOG_COLOR).*$' |
+    grep -E -v '^crates/rustc_codegen_spirv/src/lib.rs:    use std::env::{self, VarError};$'
 
 ); then
     echo '^^^'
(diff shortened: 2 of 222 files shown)


.cargo/config.toml
```diff
@@ -14,6 +14,10 @@ rustflags = [
     "-Zshare-generics=n", # =off is also an option, but you're going to increase binary size in doing so.
 ]
 
+# Override parent directory's cranelift backend setting for rust-gpu compatibility
+[profile.dev]
+codegen-backend = "llvm"
+
```
Firestar99: This addition is failing CI, it's probably something that should remain a local change on your machine. 

Thanks for your contribution, will review the rest later :D

Firestar99 (changes requested):

.cargo/config.toml
zmr-233: Thanks for the feedback! Understood, I'll revert this change and keep it local on my machine. I'll wait for the review of the rest of the changes.

Thanks for your help! :D

Firestar99:
@zmr-233  You don't need to close the PR, just update this one and keep pushing to the same branch

zmr-233:
I’ve removed the local `[profile.dev] codegen-backend = "llvm"` change, so that’s no longer affecting CI. The CI failures you see are expected, as they’re caused by the new validation I added `validate_fragment_shader_outputs`.

Some existing tests are now failing because of the “no output in fragment shader” error.
I set it as an error, but we can switch it to a warning if you prefer. Let me know how you want to handle it :D

LegNeato:
Just thinking out loud here, but maybe a compiler flag to turn this off would be useful? We could then set it in the compiletests and users can turn it off if for some reason they need to accept no output (I can't imagine why though?)

zmr-233:
> Just thinking out loud here, but maybe a compiler flag to turn this off would be useful? We could then set it in the compiletests and users can turn it off if for some reason they need to accept no output (I can't imagine why though?)

I've already updated it, and I've added a compiler flag to optionally turn this off. It can now be set in the compiletests, and users can disable it if, for some reason, they need to accept no output (though I can't imagine why : )

zmr-233:
@LegNeato Sorry, I am confident that the branch for this PR can pass the Github CI. Please try to check again. Previously, I kept the origin/main branch and upstream/main in sync with my fork, and I couldn't find a way to manually trigger the Github Action for the origin/pr-zmr233 branch. That's why I have been relying on the PR to trigger CI to identify where the issues were. However, this time I have directly merged origin/main into pr-zmr233, triggered the Github CI, and it has passed : )

PR #290, opened 2025-06-22
fix compiletests

Firestar99:


LegNeato (approved):

Firestar99:
Thx! I was gonna force-merge this anyway, but since merge-when-ready always merges, even if pipelines fail, forgot about it :D

PR #302, opened 2025-06-27
Optimize constant casts to avoid unnecessary capabilities

### LINKED ISSUE #300: Capability is required unexpectedly for constant cast
I have a following trivial situation:
```rust
const K: u8 = 20;

...
    let position = 2u32;
    let global_y_offset_bits = position * K as u32;
...
```

Here it should be trivial for compiler to cast `u8` to `u32` at compile time and make last line effectively this:
```rust
let global_y_offset_bits = position * 20u32;
```

Yet for some reason it doesn't happen and I'm getting this instead:
```
  error: `u8` without `OpCapability Int8`
      --> /home/nazar-pc/.rustup/toolchains/nightly-2024-11-22-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/num/mod.rs:1298:5
       |
  1298 | /     uint_impl! {
  1299 | |         Self = usize,
  1300 | |         ActualT = u32,
  1301 | |         SignedT = isize,
  ...    |
  1315 | |         bound_condition = " on 32-bit targets",
  1316 | |     }
       | |_____^
       |
       = note: used by unnamed global (%1504)
  note: used from within `ab_proof_of_space_gpu::shader::compute_f1::compute_f1`
      --> ../src/shader/compute_f1.rs:85:47
       |
  85   |         let global_y_offset_bits = position * K as u32;
       |                                               ^^^^^^^^
```

I think this is a bug. Even if it was a variable that is assigned a literal I'd expect the compiler to inline it with a correct type. And I believe rustc actually does this for CPU architectures like x86-64.

I imagine inability to handle cases like this will result in suboptimal output in many cases, so it'd be nice to address the root cause of it.

UPD: `const K_U32: u32 = K as u32;` as a workaround helps, but ideally shouldn't be necessary.

LegNeato:
Add constant folding for widening casts (u8->u32, f32->f64) to avoid creating intermediate types. Remove capability checks during type creation and add bidirectional validation in linker to remove unused type capabilities.

Fixes #300

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2033,6 +2033,31 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         if val.ty == dest_ty {
             val
         } else {
+            // If casting a constant, directly create a constant of the target type.
+            // This avoids creating intermediate types that might require additional
+            // capabilities. For example, casting a f16 constant to f32 will directly
+            // create a f32 constant, avoiding the need for Float16 capability if it is
+            // not used elsewhere.
+            if let Some(const_val) = self.builder.lookup_const_scalar(val) {
+                if let (SpirvType::Float(src_width), SpirvType::Float(dst_width)) =
+                    (self.lookup_type(val.ty), self.lookup_type(dest_ty))
+                {
+                    if src_width < dst_width {
+                        // Convert the bit representation to the actual float value
+                        let float_val = match src_width {
+                            32 => Some(f32::from_bits(const_val as u32) as f64),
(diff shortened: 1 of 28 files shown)


LegNeato:
I _think_ this is safe to do...everywhere we widen a const, just replace with a wider const, after everything is done look to see what types we use and match the capabilities used. But I haven't really thought through all the cases. One thing that can happen if there are a ton of constant casts we could be creating a ton of constants rather than a ton of cast instructions...not sure if that is better or worse from a perf perspective.

nazar-pc:
Will this handle a case like `u32::from(K)`? To rustc/LLVM it is basically the same thing, but likely requires inlining to happen first.

LegNeato:
No, this only does `as` casts.

LegNeato:
Pushed another commit to support `From::from`

PR #303, opened 2025-06-28
Check status of dependencies in CI

nazar-pc:
The previous PR was incomplete, which makes sense in retrospect.
`if: ${{ always() }}` makes it run regardless of the status of dependencies, so that status now needs to be checked explicitly or else it does nothing :see_no_evil:


LegNeato (approved):
🤷

PR #313, opened 2025-07-01
Support more test names in difftest

LegNeato:
These all now work. Previously, the path ones did not work:

  - cargo difftest -- difftests::arch::workgroup_memory (original format)
  - cargo difftest -- arch/workgroup_memory (path-like format)
  - cargo difftest -- arch (directory filter)


Firestar99 (changes requested):
Why does this not work?
* `cargo difftest difftests/simple`

works:
* `cargo difftest difftests::simple`
* `cargo difftest simple`
* `cargo difftest difftests`

LegNeato:
Thanks!

PR #317, opened 2025-07-03
 better const folding for int and bool 

Firestar99:
see https://github.com/Rust-GPU/rust-gpu/issues/300

const folding used to only support:
* basic int: add, mul
* casting from https://github.com/Rust-GPU/rust-gpu/pull/302

This PR adds const folding for:
* basic int: sub, div, rem, shr, lshl, ashl, neg, not (+ unchecked variants)
* basic bool: and, or, xor, not
* comparison for int and bool (`icmp`)
* select with const condition

Should probably add some diff tests

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -139,6 +139,72 @@ macro_rules! simple_op {
     };
 }
 
+// shl and shr allow different types as their operands
+macro_rules! simple_shift_op {
+    (
+        $func_name:ident
+        $(, int: $inst_int:ident)?
+        $(, uint: $inst_uint:ident)?
+        $(, sint: $inst_sint:ident)?
+        $(, fold_const {
+            $(int($int_lhs:ident, $int_rhs:ident) => $fold_int:expr;)?
+            $(uint($uint_lhs:ident, $uint_rhs:ident) => $fold_uint:expr;)?
+            $(sint($sint_lhs:ident, $sint_rhs:ident) => $fold_sint:expr;)?
+        })?
+    ) => {
+        fn $func_name(&mut self, lhs: Self::Value, rhs: Self::Value) -> Self::Value {
+            let result_type = lhs.ty;
+
+            $(
+                #[allow(unreachable_patterns, clippy::collapsible_match)]
+                if let Some(const_lhs) = self.try_get_const_value(lhs)
+                    && let Some(const_rhs) = self.try_get_const_value(rhs)
+                {
+                    #[allow(unreachable_patterns)]
+                    match (const_lhs, const_rhs) {
+                        $(
+                            (ConstValue::Unsigned($int_lhs), ConstValue::Unsigned($int_rhs)) => return self.const_uint_big(result_type, $fold_int),
+							(ConstValue::Unsigned($int_lhs), ConstValue::Signed($int_rhs)) => return self.const_uint_big(result_type, $fold_int),
+							(ConstValue::Signed($int_lhs), ConstValue::Unsigned($int_rhs)) => return self.const_uint_big(result_type, $fold_int as u128),
+							(ConstValue::Signed($int_lhs), ConstValue::Signed($int_rhs)) => return self.const_uint_big(result_type, $fold_int as u128),
+						)?
+						$(
+							(ConstValue::Unsigned($uint_lhs), ConstValue::Unsigned($uint_rhs)) => return self.const_uint_big(result_type, $fold_uint),
+                            (ConstValue::Unsigned($uint_lhs), ConstValue::Signed($uint_rhs)) => return self.const_uint_big(result_type, $fold_uint),
+                        )?
+                        $(
+                            (ConstValue::Signed($sint_lhs), ConstValue::Unsigned($sint_rhs)) => return self.const_uint_big(result_type, $fold_sint as u128),
+                            (ConstValue::Signed($sint_lhs), ConstValue::Signed($sint_rhs)) => return self.const_uint_big(result_type, $fold_sint as u128),
+                        )?
+                        _ => (),
+                    }
+                }
+            )?
+
+            match self.lookup_type(result_type) {
+                $(SpirvType::Integer(_, _) => {
+                    self.emit()
+                        .$inst_int(result_type, None, lhs.def(self), rhs.def(self))
+                })?
+                $(SpirvType::Integer(_, false) => {
+                    self.emit()
+                        .$inst_uint(result_type, None, lhs.def(self), rhs.def(self))
+                })?
+                $(SpirvType::Integer(_, true) => {
+                    self.emit()
+                        .$inst_sint(result_type, None, lhs.def(self), rhs.def(self))
+                })?
+                o => self.fatal(format!(
+                    concat!(stringify!($func_name), "() not implemented for type {}"),
+                    o.debug(result_type, self)
+                )),
+            }
+            .unwrap()
+            .with_type(result_type)
+        }
+    };
+}
+
 macro_rules! simple_uni_op {
     (
         $func_name:ident
@@ -1540,24 +1606,24 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
     simple_op! {frem_algebraic, float: f_rem} // algebraic=normal
     simple_shift_op! {
         shl,
-        int: shift_left_logical
-        // fold_const {
-        //     int(a, b) => a.wrapping_shl(b as u32);
-        // }
+        int: shift_left_logical,
+        fold_const {
+            int(a, b) => a.wrapping_shl(b as u32);
+        }
     }
     simple_shift_op! {
         lshr,
-        int: shift_right_logical
-        // fold_const {
-        //     int(a, b) => a.wrapping_shr(b as u32);
-        // }
+        uint: shift_right_logical,
+        fold_const {
+            uint(a, b) => a.wrapping_shr(b as u32);
+        }
     }
     simple_shift_op! {
         ashr,
-        int: shift_right_arithmetic
-        // fold_const {
-        //     int(a, b) => a.wrapping_shr(b as u32);
-        // }
+        sint: shift_right_arithmetic,
+        fold_const {
+            sint(a, b) => a.wrapping_shr(b as u32);
+        }
     }
     simple_uni_op! {
         neg,
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2462,6 +2462,62 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
 
         assert_ty_eq!(self, lhs.ty, rhs.ty);
         let b = SpirvType::Bool.def(self.span(), self);
+
+        if let Some(const_lhs) = self.try_get_const_value(lhs)
+            && let Some(const_rhs) = self.try_get_const_value(rhs)
+        {
+            let const_result = match self.lookup_type(lhs.ty) {
+                SpirvType::Integer(_, _) => match (const_lhs, const_rhs, op) {
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntEQ) => {
+                        Some(lhs.eq(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntEQ) => Some(lhs.eq(&rhs)),
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntNE) => {
+                        Some(lhs.ne(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntNE) => Some(lhs.ne(&rhs)),
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntUGT) => {
+                        Some(lhs.gt(&rhs))
+                    }
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntUGE) => {
+                        Some(lhs.ge(&rhs))
+                    }
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntULT) => {
+                        Some(lhs.lt(&rhs))
+                    }
+                    (ConstValue::Unsigned(lhs), ConstValue::Unsigned(rhs), IntULE) => {
+                        Some(lhs.le(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntUGT) => {
+                        Some(lhs.gt(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntUGE) => {
+                        Some(lhs.ge(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntULT) => {
+                        Some(lhs.lt(&rhs))
+                    }
+                    (ConstValue::Signed(lhs), ConstValue::Signed(rhs), IntULE) => {
+                        Some(lhs.le(&rhs))
+                    }
+                    (_, _, _) => None,
+                },
+                SpirvType::Bool => match (const_lhs, const_rhs, op) {
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntEQ) => Some(lhs.eq(&rhs)),
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntNE) => Some(lhs.ne(&rhs)),
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntUGT) => Some(lhs.gt(&rhs)),
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntUGE) => Some(lhs.ge(&rhs)),
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntULT) => Some(lhs.lt(&rhs)),
+                    (ConstValue::Bool(lhs), ConstValue::Bool(rhs), IntULE) => Some(lhs.le(&rhs)),
+                    (_, _, _) => None,
+                },
+                _ => None,
+            };
+            if let Some(result) = const_result {
+                return self.const_bool(result);
+            }
+        }
+
         match self.lookup_type(lhs.ty) {
             SpirvType::Integer(_, _) => match op {
                 IntEQ => self.emit().i_equal(b, None, lhs.def(self), rhs.def(self)),
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -54,13 +54,25 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
             match self.lookup_type(val.ty) {
                 SpirvType::Integer(bits, signed) => {
                     let size = Size::from_bits(bits);
-                    Some(if signed {
-                        ConstValue::Signed(size.sign_extend(x))
+                    // ensure the u128 constant didn't overflow and const-folding isn't hiding issues
+                    if x == size.truncate(x) {
+                        Some(if signed {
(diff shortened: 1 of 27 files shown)


Firestar99:
@eddyb Do you have any explanation why https://github.com/Rust-GPU/rust-gpu/pull/317/commits/43d5107819c863ad7f8a7662b79229cb3c22ccf3 is required to make `libm` compile? This PR also adds explicit type checking for each operation, and for some reason `libm` calls `add` that meant for integers with two floats instead of calling `fadd`.

eddyb (changes requested):
Love to see this, and I would love to debug & land this ASAP (at a glance it seems reasonably done), can you rebase it?

Firestar99:
There's nothing per se preventing this merge, I was going to wait on https://github.com/Rust-GPU/rust-gpu/pull/321 for some free testing :D

But a quick local rebase on it and it all seems to pass, so feel free to merge.

eddyb:
> But a quick local rebase on it and it all seems to pass

Looks like the `TypedBuffer` (`abi.rs` fix + tests) changes are gone, which is what I was hoping for, thanks!

Also, what I thought was a blocker was https://github.com/Rust-GPU/rust-gpu/pull/317#issuecomment-3036442224 - but I see now that you're asking why you had to do such a hack (I would still want to investigate it, but we could open an issue and land this PR as-is).

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         self.emit().unreachable().unwrap();
     }
 
+    // use wrapping_op everywhere, rustc will error out for us if the expr overflows
     simple_op! {
-        add, i_add,
+        add,
+        int: i_add,
+        // TODO why does libm fail to compile if `add` doesn't support float? Why is it not using `fadd`?
+        float: f_add,
```
eddyb: A quick search (https://github.com/search?q=repo%3ARust-GPU%2Frust-gpu+.add%28&type=code) reveals the culprit: https://github.com/Rust-GPU/rust-gpu/blob/ace9310444753f21c5798be787f28302390747a9/crates/rustc_codegen_spirv/src/builder/libm_intrinsics.rs#L245-L250

Just change that to `fadd` instead.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
-    simple_op! {fsub_fast, f_sub} // fast=normal
-    simple_op! {fsub_algebraic, f_sub} // algebraic=normal
+    simple_op! {fadd, float: f_add}
+    simple_op! {fadd_fast, float: f_add} // fast=normal
+    simple_op! {fadd_algebraic, float: f_add} // algebraic=normal
+    simple_op! {
+        sub,
+        int: i_sub,
+        // TODO why does libm fail to compile if `add` doesn't support float? Why is it not using `fadd`?
+        float: f_sub,
```
eddyb:
```suggestion
```

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+    simple_uni_op! {
+        neg,
+        sint: s_negate,
+        fold_const {
+            sint(a) => a.wrapping_neg();
+        }
+    }
+    simple_uni_op! {fneg, float: f_negate}
+
+    /// already unchecked by default
```
eddyb, line 1656: These comments are kind of funny - I forget exactly how this works but I believe e.g. `unchecked_uadd(a, b)` is `uN::checked_add(a, b).unchecked_unwrap()`.

That is, *overflowing becomes Undefined Behavior*.

LLVM has flags that start with `n` for this (e.g. `nuw`/`nsw` meaning "no (un)signed wrap"),
and [it turns out `rustc_codegen_llvm` is emitting exactly those flags](https://github.com/rust-lang/rust/blob/e3fccdd4a16bf3aa223749efef1fa981589e43ae/compiler/rustc_codegen_llvm/src/builder.rs#L417-L454).

SPIR-V seems to have added `nuw`/`nsw` equivalents in this extension, and later SPIR-V 1.4:
https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_no_integer_wrap_decoration.html#_decorations

I don't expect you to implement using it (which would have to be conditional on the SPIR-V version or that extension being in target-features, anyway), but it would be nice to remove these (wrong-ish) comments, maybe replace them with a big `// TODO` or `// FIXME` before the block of `unchecked_*` etc.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+    }
+    simple_op! {fsub, float: f_sub}
+    simple_op! {fsub_fast, float: f_sub} // fast=normal
+    simple_op! {fsub_algebraic, float: f_sub} // algebraic=normal
     simple_op! {
-        mul, i_mul,
+        mul,
+        int: i_mul,
         // HACK(eddyb) `rustc_codegen_ssa` relies on `Builder` methods doing
         // on-the-fly constant-folding, for e.g. intrinsics that copy memory.
```
eddyb: Either remove this comment, or move it somewhere more central (since it could apply to any number of constant-folding operations, multiplication is just the one that was a big pain point.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -33,76 +33,240 @@ use smallvec::SmallVec;
...
+                if let Some(const_lhs) = self.try_get_const_value(lhs) {
+                    if let Some(const_rhs) = self.try_get_const_value(rhs) {
+                        #[allow(unreachable_patterns)]
+                        match (const_lhs, const_rhs) {
+                            $(
+                                (ConstValue::Unsigned($int_lhs), ConstValue::Unsigned($int_rhs)) => return self.const_uint_big(result_type, $fold_int),
+                                (ConstValue::Signed($int_lhs), ConstValue::Signed($int_rhs)) => return self.const_uint_big(result_type, $fold_int as u128),
+                            )?
+                            $((ConstValue::Unsigned($uint_lhs), ConstValue::Unsigned($uint_rhs)) => return self.const_uint_big(result_type, $fold_uint), )?
+                            $((ConstValue::Signed($sint_lhs), ConstValue::Signed($sint_rhs)) => return self.const_uint_big(result_type, $fold_sint as u128), )?
```
eddyb: Hmm, I don't see a failure path here (i.e. "what if const-folding fails"), does `const_uint_big` truncate? (truncation might be enough for both signed and unsigned integers)

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -33,76 +33,240 @@ use smallvec::SmallVec;
...
+            match self.lookup_type(val.ty) {
+                SpirvType::Integer(bits, signed) => {
+                    let size = Size::from_bits(bits);
+                    Some(if signed {
+                        ConstValue::Signed(size.sign_extend(x))
+                    } else {
+                        ConstValue::Unsigned(size.truncate(x))
+                    })
+                }
+                SpirvType::Bool => Some(ConstValue::Bool(x != 0)),
```
eddyb: This method should return `None` on `bool` values other than `0` or `1`, and not risk hiding bugs.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -33,76 +33,240 @@ use smallvec::SmallVec;
...
+                _ => return None,
+            };
+            match self.lookup_type(val.ty) {
+                SpirvType::Integer(bits, signed) => {
+                    let size = Size::from_bits(bits);
+                    Some(if signed {
+                        ConstValue::Signed(size.sign_extend(x))
+                    } else {
+                        ConstValue::Unsigned(size.truncate(x))
+                    })
```
eddyb: This method should return `None` when `x != size.truncate(x)` (regardless of signedness), to avoid risking hiding some nasty bug (similar to the `bool` comment, there's just no `truncate` for 1-bit, unless you want to get creative with `bits` and implement sign/zero-extension yourself).

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+        uint: u_div,
+        fold_const {
+            uint(a, b) => a.wrapping_div(b);
+        }
+    }
+    simple_op! {
+        sdiv,
+        sint: s_div,
+        fold_const {
+            sint(a, b) => a.wrapping_div(b);
```
eddyb: Uhh why does `wrapping_div` even exist, it's not an operation that can overflow??

Okay, yeah, the docs do point that out:
> Wrapped division on unsigned types is just normal division. There’s no way wrapping could ever happen. This function exists so that all operations are accounted for in the wrapping operations.

What you want is `checked_div` and to bail out of if it returns `None`. Which probably means that you want to replace `if let Some(const_x) = f(x) {...}` with `f(x).and_then(|const_x| Some({...}))`, so that the expression can use `?` (or even `return None;`, tho that would be confusing).

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1370,132 +1534,185 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+        fold_const {
+            uint(a, b) => a.wrapping_div(b);
         }
     }
-    simple_op! {fmul, f_mul}
-    simple_op! {fmul_fast, f_mul} // fast=normal
-    simple_op! {fmul_algebraic, f_mul} // algebraic=normal
-    simple_op! {udiv, u_div}
     // Note: exactudiv is UB when there's a remainder, so it's valid to implement as a normal div.
     // TODO: Can we take advantage of the UB and emit something else?
```
eddyb, line 1583: Oh, that's funny, that's basically the exact situation `unchecked_*` is in, but only this has an accurate comment that mentions UB.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Firestar99: From the spirv spec:
> OpUDiv: Behavior is undefined if Operand 2 is 0.
> OpSDiv:  Behavior is undefined if Operand 2 is 0. Behavior is undefined if Operand 2 is -1 and Operand 1 is the minimum representable value for the operands' type, causing signed overflow.

From Rust docs:
> U: This operation will panic if `other == 0`.
> S: This operation will panic if `other == 0` or the division results in overflow.

(Does this panic require debug asserts on?)

If behavior in these rare cases is undefined in spirv anyway, why explicitly not const fold them for them to trigger UB during execution?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Firestar99: sounds like something that should be an issue and handled in a followup

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Firestar99: Currently, none of the const-fold ops can fail. 

> does const_uint_big truncate?

It seems like it does
https://github.com/Rust-GPU/rust-gpu/blob/05b34493ce661dccd6694cf58afc13e3c8f7a7e0/crates/rustc_codegen_spirv/src/builder_spirv.rs#L562-L574

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: > (Does this panic require debug asserts on?)

No, it's always there, because there is no possible value it could give (unlike wrapping arithmetic).

> If behavior in these rare cases is undefined in spirv anyway, why explicitly not const fold them for them to trigger UB during execution?

Maybe try adding a test? This code will panic and you will get an ICE, most likely.
At runtime, at least in safe Rust code, these divisions are gated behind a panic.

So e.g. this code:
```rust
let y = x / 0;
```
will generate MIR that looks like:
```rust
if 0 == 0 { panic!("division by zero"); }
let y = unsafe { x.unchecked_div(0) };
```
which should optimize to an unconditional panic (the actual UB after the panic never being reachable), *not* crash the compiler.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: I only brought it up because IIRC I saw `//` comments turn into `///` comments.
If you keep them `//` comments I can kinda see the argument for it being preexisting, but it's still sad that it's effectively a lie, and we could at least turn them into e.g.:
```
// FIXME: take advantage of `unchecked_*` making overflow UB.
```
(feel free to workshop it further, but even that should be good enough to replace the incorrect comments with)

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: Alright, this should be fine for both signed and unsigned, for both arithmetic and bitwise ops (tho the only bitwise op that would ever be a concern is flipping all the bits).

You only have to worry about anything that explicitly cares about the number of bits - e.g. instead of `x.wrapping_shl(y)` you should do `x.checked_shl(y & ((x_size.bits() - 1) as u128)).unwrap()`, and same for `wrapping_shr`.

If you add a test for something like `1u32 << 32`, the result should be `1` (as the shift amount gets masked out and becomes `1u32 << 0`) *not* `0` (like `(1u128 << 32) as u32` would cause).

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1357,132 +1533,179 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+            uint(a, b) => a.wrapping_shr(b as u32);
+        }
+    }
+    simple_shift_op! {
+        ashr,
+        sint: shift_right_arithmetic,
+        fold_const {
+            sint(a, b) => a.wrapping_shr(b as u32);
+        }
+    }
```
eddyb: It should be possible to show in a test that these misbehave - see https://github.com/Rust-GPU/rust-gpu/pull/317#discussion_r2200873033

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Firestar99: So I've setup an entire difftest around this just to notice that `1u32 << 32` doesn't make any sense. In debug this will panic, in a const block this will panic and fail the compile, and in release (without overflow checks) this is UB. How should we go about handling this "correctly" anyway?

~~Difftests are in https://github.com/Rust-GPU/rust-gpu/tree/const_folding_difftest due to requiring https://github.com/Rust-GPU/rust-gpu/pull/321~~ 
https://github.com/Rust-GPU/rust-gpu/pull/321 just got merged so there now in this branch

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Firestar99: I have managed to ICE in compiletest :tada:

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
eddyb: <sub>sorry, just noticed this comment (GitHub keeps failing me, I need to periodically rre-skim the entire list...)<sub>

> and in release (without overflow checks) this is UB

I think there's been a few layers of confusion, the closest thing to that would be `unchecked_*` (AIUI, `a.unchecked_foo(b)` roughly means `a.checked_foo(b).unchecked_unwrap()`, i.e. UB instead of panicking).

(Also I initially assumed you were talking about Rust-GPU deficiencies, but it would be extremely unsound if a safe operation like `a << b` could result in UB, and this was fixed over a decade ago IIRC)

The only mandatory checks are for division (and remainder etc.) by `0`, since there is no non-UB alternative, and that just means they unconditionally panic (similar to array indexing bounds checks).

Shifts mask the shift amount, so `1u32 << 32` needs to behave like `1u32 << 0`, when overflow checks are off.

**EDIT**: I found something I was missing myself before: the `Builder::shl` actually is for e.g. an `unchecked_shl`:
- https://github.com/rust-lang/rust/blob/a955f1cd09a027363729ceed919952d09f76f28e/compiler/rustc_mir_transform/src/lower_intrinsics.rs#L104-L121
- https://github.com/rust-lang/rust/blob/a955f1cd09a027363729ceed919952d09f76f28e/compiler/rustc_codegen_ssa/src/mir/rvalue.rs#L832-L835
- https://github.com/rust-lang/rust/blob/a955f1cd09a027363729ceed919952d09f76f28e/compiler/rustc_codegen_ssa/src/base.rs#L307-L359

So it's just badly named, and the masking will happen via `rustc_codegen_ssa`.
This does mean it's basically untestable, so maybe I can ignore all this?

tests/difftests/tests/lang/core/ops/const_fold_int/const-fold-cpu/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub type EvalResult = [[[u32; 8]; 8]; 2];
+
+impl Variants {
+    pub fn eval(&self, input_patterns: &[u32; 8]) -> EvalResult {
+        match self {
+            Variants::ConstFold => [interesting_patterns!(op_u), interesting_patterns!(op_i)],
+            Variants::ConstExpr => {
+                const { [interesting_patterns!(op_u), interesting_patterns!(op_i)] }
+            }
```
eddyb, line 75: I can see the issue - you want something like `#[cfg(not(debug_assertions))]` for testing `x << 32` and `x >> 32`, and there's no point in trying `const {...}`.

Sadly, `wrapping_shl`/`wrapping_shr` are library functions, only `unchecked_shl` is an intrinsic, so there's no way for the `rustc_codegen_spirv::builder` constant-folding logic to actually... wait, I think I just realized something, now I have to go back to the other comment, oops.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1357,132 +1540,179 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
     }
 
+    // use wrapping_op everywhere, rustc will error out for us if the expr overflows
     simple_op! {
-        add, i_add,
+        add,
+        int: i_add,
         fold_const {
-            int(a, b) => a.wrapping_add(b)
+            int(a, b) => a.checked_add(b)?;
```
eddyb, line 1548: Comment doesn't match changes, did you mean `checked_op`?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1357,132 +1540,179 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+            uint(a, b) => a.checked_shr(b as u32)?;
+        }
+    }
+    simple_shift_op! {
+        ashr,
+        sint: shift_right_arithmetic,
+        fold_const {
+            sint(a, b) => a.checked_shr(b as u32)?;
+        }
+    }
```
eddyb, line 1646: If you want to actually get `None` from `checked_{shl,shr}`, you would want something involving the bitwidth of `a` - but better safe than sorry, I suppose.

eddyb (approved):
I had some comments but I think I just want to land this, I delayed it enough as it is

tests/difftests/tests/lang/core/ops/const_fold_int/const-fold-cpu/src/lib.rs
Firestar99: continued in https://github.com/Rust-GPU/rust-gpu/pull/317#discussion_r2232013499

PR #320, opened 2025-07-04
rustup: update to `nightly-2025-06-23` (~1.89).

eddyb:
Not much to see this time, other than:
- some more `format_args!` complications (less efficient codegen for some reason)
- `-Zshare-generics=off` now being needed (because of weird special-casing done to `compiler-builtins` otherwise leaking to shader crates, via e.g. `<Range<usize> as Iterator>>:next`)
- a couple bulk Clippy-driven changes (esp. `clippy::collapsible_if`, it's surprisingly applicable)

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
@@ -168,25 +168,25 @@ impl<'tcx> CodegenCx<'tcx> {
         }
 
         // Check if this is a From trait implementation
-        if let Some(impl_def_id) = self.tcx.impl_of_method(def_id) {
-            if let Some(trait_ref) = self.tcx.impl_trait_ref(impl_def_id) {
-                let trait_def_id = trait_ref.skip_binder().def_id;
+        if let Some(impl_def_id) = self.tcx.impl_of_method(def_id)
+            && let Some(trait_ref) = self.tcx.impl_trait_ref(impl_def_id)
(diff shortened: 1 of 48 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
```diff
@@ -385,19 +387,69 @@ impl<'tcx> StaticCodegenMethods for CodegenCx<'tcx> {
...
         self.builder
             .set_global_initializer(g.def_cx(self), v.def_cx(self));
+
+        let attrs = self.tcx.codegen_fn_attrs(def_id);
+
+        let alloc = alloc.inner();
+        let align_override =
+            Some(alloc.align).filter(|&align| align != self.lookup_type(value_ty).alignof(self));
+        if let Some(_align) = align_override {
+            // FIXME(eddyb) implement, or at least error.
```
LegNeato, line 397: `todo!()` these? I don't know this code so don't know if needed.

eddyb: `todo!()` would ICE, it should at the very least be a nice user error if we do want to enforce it.
(otherwise, e.g. `assert_eq!(alloc.align, self.lookup_type(value_ty).alignof(self))` is much better than a `todo!()`, but I don't think we should do either)

These things used to be ignored before, too, I just made the code slightly clearer that that's happening.

PR #321, opened 2025-07-04
Add 7 new difftests and push constants support

LegNeato:
I wanted to increase coverage, so I worked with AI to bang out some tests and infra changes.

New tests:
- bitwise_ops: bit manipulation operations
- trig_ops: trigonometric functions
- control_flow_complex: nested loops and complex control flow
- vector_swizzle: vector component access and swizzling
- memory_barriers: workgroup memory synchronization
- vector_extract_insert: dynamic vector element access
- push_constants: push constants in compute shaders

Infrastructure changes:
- Add WgpuComputeTestPushConstants for push constant support
- Enable PUSH_CONSTANTS feature in wgpu when needed

### DIFF
--- a/tests/difftests/lib/src/scaffold/compute/mod.rs
+++ b/tests/difftests/lib/src/scaffold/compute/mod.rs
@@ -1,2 +1,5 @@
 mod wgpu;
-pub use wgpu::{RustComputeShader, WgpuComputeTest, WgslComputeShader};
+pub use wgpu::{
+    BufferConfig, BufferUsage, RustComputeShader, WgpuComputeTest, WgpuComputeTestMultiBuffer,
+    WgpuComputeTestPushConstants, WgslComputeShader,
+};
--- a/tests/difftests/lib/src/scaffold/compute/wgpu.rs
+++ b/tests/difftests/lib/src/scaffold/compute/wgpu.rs
@@ -109,6 +109,36 @@ pub struct WgpuComputeTest<S> {
     output_bytes: u64,
 }
 
+/// More flexible compute test that supports multiple buffers.
+pub struct WgpuComputeTestMultiBuffer<S> {
+    shader: S,
+    dispatch: [u32; 3],
+    buffers: Vec<BufferConfig>,
+}
+
+/// Compute test that supports push constants.
+pub struct WgpuComputeTestPushConstants<S> {
+    shader: S,
+    dispatch: [u32; 3],
+    buffers: Vec<BufferConfig>,
+    push_constants_size: u32,
+    push_constants_data: Vec<u8>,
+}
+
+#[derive(Clone)]
+pub struct BufferConfig {
+    pub size: u64,
+    pub usage: BufferUsage,
+    pub initial_data: Option<Vec<u8>>,
+}
+
+#[derive(Clone, Copy, PartialEq)]
+pub enum BufferUsage {
+    Storage,
+    StorageReadOnly,
+    Uniform,
+}
+
 impl<S> WgpuComputeTest<S>
 where
     S: ComputeShader,
@@ -122,6 +152,10 @@ where
     }
 
     fn init() -> anyhow::Result<(wgpu::Device, wgpu::Queue)> {
+        Self::init_with_features(wgpu::Features::empty())
+    }
+
+    fn init_with_features(features: wgpu::Features) -> anyhow::Result<(wgpu::Device, wgpu::Queue)> {
         block_on(async {
             let instance = wgpu::Instance::new(&wgpu::InstanceDescriptor {
                 #[cfg(target_os = "linux")]
@@ -143,10 +177,13 @@ where
                 .request_device(&wgpu::DeviceDescriptor {
                     label: Some("wgpu Device"),
                     #[cfg(target_os = "linux")]
-                    required_features: wgpu::Features::SPIRV_SHADER_PASSTHROUGH,
+                    required_features: wgpu::Features::SPIRV_SHADER_PASSTHROUGH | features,
                     #[cfg(not(target_os = "linux"))]
-                    required_features: wgpu::Features::empty(),
-                    required_limits: wgpu::Limits::default(),
+                    required_features: features,
+                    required_limits: wgpu::Limits {
+                        max_push_constant_size: 128,
+                        ..wgpu::Limits::default()
+                    },
                     memory_hints: Default::default(),
                     trace: Default::default(),
                 })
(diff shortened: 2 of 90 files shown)


Firestar99:
I'd love to have these for testing my const folding in https://github.com/Rust-GPU/rust-gpu/pull/317

Firestar99:
CI Ubuntu reports these failures:
    difftests::lang::core::ops::math_ops
    difftests::lang::core::ops::matrix_ops
    difftests::lang::core::ops::vector_ops

On my machine I get these failures:
    difftests::lang::core::ops::math_ops
    difftests::lang::core::ops::matrix_ops
    
Could this be float inaccuracies between vendors?

LegNeato:
Yeah, I think it is floats because they all passed locally after I took care of some existing float issues. I'll take a look and either round or omit.

LegNeato:
I added a `compat_round!` macro that handled most of the differences, but could not get all of them. So I added epsilon support when comparing. And then because I was debugging with python scripts I added better debug output when dealing with f32s. And because I did that I reworked all the output cases to match.

This is a lot of code and I don't really expect you to review it...AI wrote a lot of it and I reviewed and tweaked. I figure we'll iterate on main as it is super self-contained. I can probably break it up if you want though!

As an aside, I think some of these should be runtests, as I am not sure what value we get diffing with wgsl for all of them. BTW, if test times get long we can shard in CI as theoretically we could have 1 job for each test as they are unrelated.

LegNeato:
(note it is not set to automerge as I'll want to squash first)

tests/difftests/lib/src/scaffold/compute/vulkano.rs
```diff
@@ -0,0 +1,250 @@
...
+    memory::allocator::{AllocationCreateInfo, MemoryTypeFilter, StandardMemoryAllocator},
+    pipeline::{
+        ComputePipeline, Pipeline, PipelineLayout, PipelineShaderStageCreateInfo,
+        compute::ComputePipelineCreateInfo, layout::PipelineDescriptorSetLayoutCreateInfo,
+    },
+    shader::{ShaderModule, ShaderModuleCreateInfo},
+    sync::{self, GpuFuture},
+};
+
+pub struct VulkanoBackend {
```
Firestar99: Any particular reason you choose to add a vulkano backend for the workgroup memory test? 

I've used vulkano quite a bit in the past with rust-gpu and even [made an integration example](https://github.com/Firestar99/rust-gpu-vulkano-example), but I'm actually not the biggest fan nowadays. Primarily, their current sync system is deeply flawed and you can get it to quite easily cause validation errors. For a while, I ran a modded variant of their sync with [fixes](https://github.com/vulkano-rs/vulkano/pull/2515) and [performance optimizations](https://github.com/vulkano-rs/vulkano/pull/2539), until this unmaintainable mess completely broke down for me. They are in the process of replacing it with a new task graph based system, which has been in dev for like 1.5 years and still not fully ready to be used. Putting them in a bit of a weird spot currently. And even if they finish it, I'd actually still recommend wgpu over it since I'm not a fan of full task graphs nor do I think they have the manpower to properly maintain it. 

If you want a plain vulkan impl, let me know and I'll write you one in ash.

LegNeato: The [windows version of one test is failing](https://github.com/Rust-GPU/rust-gpu/actions/runs/16102760744/job/45434044828) and I can't see why. I am trying to see if it is a naga / wgpu bug, so I am adding the vulkano version to compare. I hope to not land the vulkan version.

LegNeato:
Ok, the [windows issue](https://github.com/Rust-GPU/rust-gpu/actions/runs/16102760744/job/45434044828) is definitely a wgpu + naga (dxc) bug.

LegNeato:
I've pulled out the repo for the wgpu/naga bug: https://github.com/LegNeato/wgpu-workgroup-memory-bug

Firestar99:
Why do we need swiftshader on windows?

.github/workflows/ci.yaml
```diff
@@ -29,6 +29,11 @@ jobs:
         with:
           version: 1.4.309.0
           cache: true
+      - if: ${{ runner.os == 'Windows' }}
+        name: Install Vulkan Runtime with SwiftShader (Windows)
+        uses: NcStudios/VulkanCI@v1.2
+        with:
+          sdkVersion: 1.4.309.0
```
Firestar99, line 36: Why do we need swiftshader on windows?

tests/difftests/README.md
```diff
@@ -137,13 +186,66 @@ cargo difftest --nocapture
...
 ## Debugging Failing Tests
 
-If outputs differ, the error message lists:
+When outputs differ, the harness provides detailed error reporting:
+
+### For raw byte differences
+
+- Shows which packages produced different outputs
+- Lists output file paths for manual inspection
+- Groups packages by their output values
```
Firestar99, line 192: I love that you can now see the difference in the console of the CI, like floats! Why not also do it with raw values, assuming them to be hex?

tests/difftests/lib/src/config.rs
```diff
@@ -1,9 +1,63 @@
...
+    I32,
+}
+
+impl TestMetadata {
+    /// Create metadata with a specific epsilon value, keeping default output type
+    pub fn with_epsilon(epsilon: f32) -> Self {
+        Self {
+            epsilon: Some(epsilon),
+            ..Default::default()
+        }
```
Firestar99, line 70: Default `output_type` is `Raw`, where epsilon doesn't make any sense. Rename it to `with_f32(epsilon: f32) -> Self` and adjust the current docs?

tests/difftests/lib/src/scaffold/compute/ash.rs
```diff
@@ -0,0 +1,436 @@
...
+            // Create instance
+            let app_info = vk::ApplicationInfo::default()
+                .application_name(CStr::from_bytes_with_nul_unchecked(b"difftest\0"))
+                .application_version(vk::make_api_version(0, 1, 0, 0))
+                .engine_name(CStr::from_bytes_with_nul_unchecked(b"difftest\0"))
+                .engine_version(vk::make_api_version(0, 1, 0, 0))
+                .api_version(vk::API_VERSION_1_2);
+
+            let instance_create_info =
+                vk::InstanceCreateInfo::default().application_info(&app_info);
```
Firestar99: Should we force-enable the vulkan validation layers? Potentially with GPU-assisted validation, which the current test cases likely won't benefit, but future ones might?

tests/difftests/lib/src/scaffold/compute/ash.rs
```diff
@@ -0,0 +1,436 @@
...
+                },
+                vk::DescriptorPoolSize {
+                    ty: vk::DescriptorType::UNIFORM_BUFFER,
+                    descriptor_count: 16,
+                },
+            ];
+
+            let descriptor_pool_create_info = vk::DescriptorPoolCreateInfo::default()
+                .pool_sizes(&descriptor_pool_sizes)
+                .max_sets(16);
```
Firestar99: Feels bad to hard code 16, but noone will use more than 16 buffers anyway... I'd be ok with keeping it this way. Or moving it to `run_compute` where we know how many we'll need. I assume this is AI code?

tests/difftests/lib/src/scaffold/compute/ash.rs
```diff
@@ -0,0 +1,436 @@
...
+                descriptor_pool,
+                memory_properties,
+                _entry: entry,
+            })
+        }
+    }
+
+    fn run_compute(
+        &self,
+        spirv_bytes: &[u8],
```
Firestar99, line 164: why does this take `spirv_bytes` as a `&[u8]` and not something like `impl ComputeShader` like wgpu does? Would need to adjust the `ComputeShader` trait to support just returning like a `spirv_bytes: &[u8]`. That would also remove a lot of the custom code in the single ash test we have.

Firestar99 (changes requested):
What I love about this PR:
* we print any detected changes into the logs, so you can easily see in CI the detected difference without having to boot up that platform
* epsilon for floating point
* the `ComputeShader` trait to make it easy to switch out shaders
* Skip cause we'll likely need it in more places than this

I wonder if we should separate out the difftest improvements from the actual difftests in separate PRs?

The only major concern I have is that there's a lot of copy-paste between the different variants of a single test case. They usually only differ by the `impl ComputeShader` given as a arg. Doesn't have to be this PR, can be a followup, but I'd like to see those almost identical mains collapse. Make all variants of a single difftest a single crate. Some ideas around how that could be done:
* Features. The runner reads all features of the single difftest crate (via `cargo_metadata`) and runs the test with one feature being active at a time. 
* Args, which I almost prefer. Read from some metadata toml file the variants you want and pass them in as args, via `struct Config`. Then switching between shaders is just a match, not some `#[cfg(...)]` monster. Would require every variant to depend on wgpu, naga and ash, but we need to compile all of them anyway.
3. Metadata could be like github runners to allow cartesian product:
```toml
# skip wgsl and cpu variant
[variants]
shader = ["rust", "wgsl"]
runner = ["wgpu", "ash", "cpu"]
```

.github/workflows/ci.yaml
LegNeato: Otherwise we hit https://github.com/LegNeato/wgpu-workgroup-memory-bug, which appears to be a bug on WARP.

LegNeato:
> I wonder if we should separate out the difftest improvements from the actual difftests in separate PRs?

I'll look into how hard this is.

> The only major concern I have is that there's a lot of copy-paste between the different variants of a single test case. They usually only differ by the `impl ComputeShader` given as a arg. Doesn't have to be this PR, can be a followup, but I'd like to see those almost identical mains collapse. Make all variants of a single difftest a single crate. Some ideas around how that could be done:

Yeah, I think once we get the kinks worked out I want to do stuff like compiletest where we just drop a .rs file with magic comments and everything else is taken care of. There is too much boiler plate right now, but I didn't want to get too fancy at the start.

The reason why I started it this way is I wanted to make sure we could test things like 3rd party deps, setting values in `Cargo.toml`, using SpirvBuilder from `build.rs`, etc. When I thought about all the possibilities I wanted to support, I realized I was just doing a giant DSL and it would be simpler to just use standard Rust 😄 . We can always use standard rust tools to reduce boilerplate  like the `include!()` I used in one of the tests so the same shader code is used between `ash` and `wgpu` variants or macros and such.

tests/difftests/lib/src/scaffold/compute/ash.rs
LegNeato: Will do, yep, missed this in the AI code.

tests/difftests/lib/src/scaffold/compute/ash.rs
LegNeato: I think maybe having them as a default on option and letting tests turn it off will be good.

tests/difftests/README.md
LegNeato: Screenshot 2025-07-09 at 4 44 30 PM

tests/difftests/lib/src/scaffold/compute/ash.rs
LegNeato: I'm going to leave this for a followup, feels like its own feature.

tests/difftests/lib/src/scaffold/compute/ash.rs
LegNeato: Refactored to use 16 by default but let tests set it.

Firestar99:
It's probably just too little `ε` but this fails on my machine (Ryzen 680M RDNA2 iGPU):

```
test difftests::lang::core::ops::matrix_ops ... thread 'main' panicked at tests/difftests/bin/src/main.rs:108:49:
Outputs differ:

matrix_ops (lang/core/ops/matrix_ops)
F32, ε=0.00001 • 8 differences • max: 1.526e-5 (0.00%)
─────────────────────────────────────────────────────────────────

▪ matrix_ops-rust
  → Raw:  /tmp/.tmpgPQ1J2 (6.2 KiB)
  → Text: /tmp/.tmpgPQ1J2.txt

▪ matrix_ops-wgsl
  → Raw:  /tmp/.tmpHEcalv (6.2 KiB)
  → Text: /tmp/.tmpHEcalv.txt

┌──────┬─────────────────┬─────────────────┬──────────┬───────┐
│  #   │ matrix_ops-rust │ matrix_ops-wgsl │  Δ abs   │  Δ %  │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 185  │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 385  │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 585  │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 785  │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 985  │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 1185 │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 1385 │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
├──────┼─────────────────┼─────────────────┼──────────┼───────┤
│ 1585 │ -93.107101440   │ -93.107086182   │ 1.526e-5 │ 0.00% │
└──────┴─────────────────┴─────────────────┴──────────┴───────┘
```
(did I mention that I love this diff? :heart: )

LegNeato:
@Firestar99 can you try the latest on your machine to confirm bumping the epsilon makes the tests pass?

I tried to slice this up into more logical commits but it is painful. I'll try to at least clean the history up a bit.

Firestar99:
The larger epsilon did fix that test, everything passes now!

tests/difftests/lib/src/scaffold/compute/ash.rs
Firestar99: I'd be fine deferring this to a followup as well

Firestar99:
Just FYI, difftest are not clippy checked in CI, fixed in https://github.com/Rust-GPU/rust-gpu/pull/329

Firestar99:
Apart from a rebase and https://github.com/Rust-GPU/rust-gpu/pull/321#discussion_r2194730037, this would be ready to be merged, right?

LegNeato:
Yep, just been offline. Will try to get to it in the next couple days if no one beats me to it

tests/difftests/lib/src/scaffold/compute/ash.rs
LegNeato: Think I did what you were expecting here.

LegNeato:
@Firestar99 should be ready now I think

Firestar99:
This is still unresolved: https://github.com/Rust-GPU/rust-gpu/pull/321#discussion_r2194730037

Otherwise, I think this is ready to go!

Firestar99 (approved):
I resolved https://github.com/Rust-GPU/rust-gpu/pull/321#discussion_r2194730037 myself by changing the `TestMetadata` constructors in https://github.com/Rust-GPU/rust-gpu/pull/321/commits/0e026d48dfbb73e204bcf43f5aa0caff897b7262. If you're fine with this change, feel free to merge.

LegNeato:
Thanks! Looks good. Will merge tomorrow unless you beat me to it.

Firestar99:
I have a difftest for the `const_folding` PR that needs these changes, so merging now

PR #327, opened 2025-07-07
Backport rust-lang/rust#134117 and allow only `<*T>::offset`-like "GEPs".

eddyb:

*Temporarily based on:*
- #326

---

LLVM's [`GetElementPointer`](https://llvm.org/docs/GetElementPtr.html) (akin to SPIR-V `OpAccessChain`/`OpPtrAccessChain`) was, for a long time, the main way to interact with pointers, as it could perform, in one go, all of
e.g.: `&(*(p: *Foo).add(i)).outer_field3[j].0.inner_field1[k].leaf_field2`, even today:
```llvm
%out = getelementptr Foo, ptr %p, i64 %i, i32 3, i64 %j, i32 0, i32 1, i64 %k, i32 2
```
continues to work, even in the "untyped pointer" world (i.e. `%p` having type `ptr` instead of `Foo*`).

However, both in LLVM itself, and in `rustc_codegen_llvm`, its patterns have been shrinking with time:
- https://discourse.llvm.org/t/rfc-replacing-getelementptr-with-ptradd/68699
- https://github.com/rust-lang/rust/pull/121665
- https://github.com/rust-lang/rust/pull/134117
  <sub>encountered in (yet-unlanded) https://github.com/Rust-GPU/rust-gpu/pull/249 </sub>

And they're generally converging on `<*T>::offset`-like pointer arithmetic, where `T` is:
- in the general case: the element type of a `[T; N]` or `[T]` type
  - e.g. `<[T]>::get_unchecked(xs, i)` is `&*(xs as *const [T] as *const T).add(i)`
- (common) special-case: `i8`/`u8` for arbitrary byte offsetting
  - often used with a constant offset, for e.g. forming pointers to `struct` fields
  - in theory all `T`s could go through byte offsetts by multiplying w/ the size of `T`, aka "stride", but having that multiplication in the IR is not (yet) on the table
    (even SPIR-T `qptr` uses a special "strided offset", even if it erases the specific type `T`)

---

Instead of keeping the increasingly-gnarly `maybe_inbounds_gep` (see also e.g. #233 - a valiant effort at reconciling several of its conflicting needs), this PR moves the core logic to a new method,
named `ptr_offset_strided` (suffixed `_strided` to hopefully avoid confusion with the byte-oriented `ptradd`).

By only handling `(p: *T).offset(i)`, no more, no less (i.e. no intra-`T` offsetting), a lot of the decision logic/correctness subtleties/etc., vanish away, fixing edge cases like https://github.com/Rust-GPU/rust-gpu/pull/233#issuecomment-3025933339 without breaking the `issue-46` test (which #233 itself first fixed).

Sadly I wasn't able to keep a lot of the comments/`trace!`s added in #233, but I tried to have a bare minimum, combined with the continued existence of `#[instrument]` on/`trace!` in, several helper methods.

---

**TODO**: the last commit still needs some comments to be adjusted, after an unnecessary `pointercast` was removed (it was only ever needed due to `format_args!`).


Firestar99:
Builds with my project :tada:

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -655,344 +656,220 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-                _ => self.fatal(format!(
-                    "GEP not implemented for indexing into type {}",
-                    self.debug_type(calculated_pointee_type)
-                )),
+        let indexed_base_ptr = self.ptr_offset_strided(ptr, ty, ptr_base_index, is_inbounds);
+
+        // HACK(eddyb) this effectively removes any real support for GEPs with
+        // any `indices` (beyond `ptr_base_index`), which should now be the case
+        // across `rustc_codegen_ssa` (see also comment inside `inbounds_gep`).
+        // FIXME(eddyb) are the warning + fallback path even work keeping?
```
LegNeato: I would say delete it...this code is already convoluted enough.

PR #329, opened 2025-07-10
ci: enable clippy for more crates

Firestar99:
clippy doesn't run on eg. compiletest and difftest and there's some code it's complaining about


Firestar99:
I mean, you should be able to detect clippy in the build script and return. Currently, using cargo-gpu in a build script will fail `cargo clippy` on your project, so there really should be a way :sweat_smile:

PR #331, opened 2025-07-11
ci: simplify clippy

Firestar99:
# Requires https://github.com/Rust-GPU/spirv-tools-rs/pull/16 and publishing spirv-tools 0.12.1

> Hmpf, unfortunate that we can't pass the features somehow, to a single `cargo clippy --workspace` invocation.
> 
> Alternatively, it might be nice to not have SPIRV-Tools' C++ code be compiled by `spirv-tools-rs` when built for `cargo check`/`cargo clippy`, but I don't recall if that's even feasible to do.

https://github.com/Rust-GPU/rust-gpu/pull/329#pullrequestreview-3009595291

Well, now we can :D

This also speeds up our lint CI from 7m to 6m, likely from better multi-core utilization due to checking more crates in parallel


Firestar99:
This is now ready to merge!

Issue #318, opened 2025-07-03
Switch compiletests to `ui_test`


Firestar99:
Switch from [compiletest_rs](https://docs.rs/compiletest_rs/latest/compiletest_rs/) to the more modern [ui_test](https://docs.rs/ui_test/latest/ui_test/). Haven't looked at what it improves on, but seems to have parallel test runs.

Firestar99:
ui_test also has differential test support, though it's not really documented:

https://github.com/oli-obk/ui_test/blob/main/examples/rustc_twice_with_different_flags.rs

PR #334, opened 2025-07-14
difftest: nextest support and speedups

Firestar99:
Some difftest refactors of things I just want to have:
* switch to `mimic-libtest` from `tester`
* add profile-based test filtering via nextest, as before:
  * `cargo nextest run` will ignore difftests
  * `cargo difftest` will only run difftests
* output is now in `tests/difftests/tests/target/difftest` so you can easily inspect the output instead of having to search your tmp dir
* moved runner to new `runner` crate, `bin` just wraps it as a "test without harness"
  * split was needed to have both tests within the runner and exposing it as a test runner
* moved `lib` into `tests/difftests/tests` workspace
  * prevents "Updating crates.io index" on every single test, causing massive slowdowns
  * created `types` crate with shared `mod config` between `lib` and `runner`
  * we no longer compile `rustc_codegen_spirv`, `wgpu`, `ash` etc. in both workspaces, now only in difftest workspace

### Speedups 
**These numbers are *super* outdated but should generally still apply**

* [master reference](https://github.com/Rust-GPU/rust-gpu/actions/runs/16272157467/usage)
* rerunning difftest without changes: 
  * Linux: 45s -> 6s
  * Windows: 195s -> 80s (whatever is going on there)
* **NOTE**: The CI numbers are old and have not been retested with https://github.com/Rust-GPU/rust-gpu/pull/530 or https://github.com/Rust-GPU/rust-gpu/pull/526
* CI difftest times:
  * Linux: 12m -> 8m
  * Windows: 22m -> 16m
* total CI time: 23m -> 17m


Firestar99:
Even if you specify nextest filters that exclude a certain package, nextest will still build it, just not run it to discover any potential tests. I'll have to add back in the manual package excludes in CI to get back to previous CI speeds. 

feature request: https://github.com/nextest-rs/nextest/issues/154

Firestar99:
@LegNeato this one is ready to review, apart from the macOS difftest failing. Could you do me a favor and debug that one on your mac? It's only `difftests::arch::workgroup_memory` failing with `I/O error: No such file or directory (os error 2)` when running the ash variant.
(especially since github is taking forever to launch CI again...)

LegNeato:
Pushed a fix. The problem was you were making the output directory and coming up with `temp_output_path`, but you weren't creating `temp_output_path`. Preivously the contract was the harness creates the file and passes the path to the test. The tests expected the file to be created by the harness so they could just blindly open and write and the output `Differ`s expected the file to exist so they could just read.

.github/workflows/ci.yaml
```diff
@@ -181,14 +186,18 @@ jobs:
...
         run: echo "TARGET=$(rustc --print host-tuple)" >> "$GITHUB_ENV"
+      - name: install nextest
+        uses: taiki-e/install-action@nextest
       - name: cargo fetch --locked
         run: cargo fetch --locked --target $TARGET
       - name: cargo fetch --locked difftests
         run: cargo fetch --locked --manifest-path=tests/difftests/tests/Cargo.toml --target $TARGET
-      - name: test difftest
-        run: cargo test -p "difftest*" --release --no-default-features --features "use-installed-tools"
+      - name: test difftest-runner
```
LegNeato: My main concern here is we are making CI smarter when best practice is to make it as dumb as possible.

tests/difftests/bin/src/main.rs
LegNeato, line 1: Git thinks this was deleted rather than moved to lib and edited...can we connect them somehow?

tests/difftests/runner/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+        output_dir,
+    });
+
+    let test_cases = collect_test_dirs(&base).expect("Failed to collect test case directories");
+    if test_cases.is_empty() {
+        eprintln!("No valid tests found in {}", base.display());
+        process::exit(1);
+    }
+
+    // // We build first to ensure that the tests are compiled before running them and to
```
LegNeato: Delete

tests/difftests/tests/rust-toolchain.toml
LegNeato, line 1: Just symlink this to the one in the root.

LegNeato:
`nextest list` compiles things like `wgpu`, is that expected?

LegNeato:
`nextest list` also doesn't show any compiletests or difftests and `nextest test` doesn't run them...I would expect it to run all of them by default so we can tell people "install nextest, run `nextest test` before putting up a PR". But I guess this PR is just replacing the impl for `cargo difftest` and that can wait.

BTW, running `cargo test` from the root now appears to run difftests...

tests/difftests/tests/rust-toolchain.toml
Firestar99: I tested it and it does not seem to be needed, cargo will pick whatever `rust-toolchain.toml` file it can find in any of the parent folders, regardless of workspace. (Which makes sense since it can't know what a workspace is when figuring out what toolchain to use to call cargo)

Firestar99:
nextest has a `-P profile` option to select a profile from the folllowing ones I configured:
https://github.com/Rust-GPU/rust-gpu/blob/506469f61ada04a72ff7c7291343f00031c560ac/.config/nextest.toml#L1-L10

And `cargo difftest` just points nextest to the difftest profile:
https://github.com/Rust-GPU/rust-gpu/blob/506469f61ada04a72ff7c7291343f00031c560ac/.cargo/config.toml#L3

The only problem is that nextest compiles all crates, regardless of the profile excluding the crate / package or not. So in CI and the difftest alias I use cargo's `-p` option to only select relevant packages, to compile fewer crates. See [this nextest feature request](https://github.com/nextest-rs/nextest/issues/154).

> `nextest test` doesn't run [difftests]

With `nextest run -P difftest` you can. Let me know if you'd rather have the default test everything (including compiletests as well? https://github.com/Rust-GPU/rust-gpu/issues/318)

> `nextest list` compiles things like `wgpu`, is that expected?

In that sense, yes it is expected, since nextest needs to call the test binary to figure out which tests are contained within. Although, `nextest run` matches previous behavior on master where `example-runner-wgpu` and the "old" `difftest` lib (now moved to the difftest workspace) would force you to compile `wgpu` anyway.

> BTW, running `cargo test` from the root now appears to run difftests...

`cargo test` does not care for nextest profiles (obviously), not much I can do there.

.github/workflows/ci.yaml
Firestar99: I'm still a little unsure what to do about precompiling the tests, which is not just an CI issue but also when building from your own machine. Currently, it'll look like the tests are just stuck eating CPU for minutes, apart from the occasional slow warning from nextest, as all tests are blocked on the first test holding the target folder lock and compiling all dependencies. So at least for CI, instead of windows taking 7-8min of "nothing happening", we get cargo's compilation progress reporting. 

We could move precompiling tests to the listing stage, but that would make `nextest list` take forever on first execution. Not even sure if we can even print compilation progress, as when listing tests we must only write a list of tests to stdout and nothing else, otherwise nextest will error out. Though I haven't tested if piping through stderr would work. 

Maybe we could alias `cargo difftest` to be proceeded by a test compile before running nextest?

tests/difftests/bin/src/main.rs
Firestar99: could split the file move into a separate commit (that would fail to compile)

tests/difftests/tests/Cargo.toml
```diff
@@ -44,15 +45,18 @@ unexpected_cfgs = { level = "allow", check-cfg = [
     'cfg(target_arch, values("spirv"))'
 ] }
 
+# Cargo Windows bug: workspace dependencies of crates.io dependencies are resolved
+# incorrectly in the root workspace instead of this difftest workspace. So all
+# workspace dependencies here must be mirrored into the root workspace as well.
```
Firestar99, line 66: I still don't fully understand what's causing this, I tried creating a minimal repro repo but haven't been able to. How difftest calls cargo (`--manifest-path` or inherited env vars) seems to be important in the repro, as you can just `cargo build` in the `tests/difftest/tests` repo and it'll build just fine.

.github/workflows/ci.yaml
LegNeato: Isn't that basically what we had? It ran `cargo build --release` in the tests workspace regardless of what was running. The thinking was that most of the deps would be shared, but I guess I soon as I added `ash` vs `wgpu` that went out the window.

Firestar99: Crates in a workspace will compile all their dependencies into the `target` dir of their workspace, regardless of where they're from. So on master, the difftest `lib` crate would be compiled once in the difftest workspace's target and a second time in the root workspace's target. Now only `types` is shared like that, and a lot lighter since it's just plain types and no `wgpu` or `ash`.

Firestar99:
@LegNeato what's the state on this PR?

We may want to eventually switch to `ui_test` anyway (see https://github.com/Rust-GPU/rust-gpu/issues/318) and get rid of nextest again. Though I feel like this PR has a number of improvements, so we should either merge this anyway, merge the non-nextest part of it or I switch it to `ui_test` right away. (Though finishing graphite has priority rn)

LegNeato:
I tried to switch rust-cuda to ui_test from the start and hit nothing but pain. IIRC ui_test doesn't really have directives and such built in so you have to reimplement all that yourself...compiletest has a lot more built-in.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -723,13 +723,6 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
                     (TyPat::Void, SpirvType::Void) => Ok([None]),
                     (TyPat::Pointer(_, pat), SpirvType::Pointer { pointee: ty, .. })
                     | (TyPat::Vector(pat), SpirvType::Vector { element: ty, .. })
-                    | (
```
LegNeato, line 734: This is unrelated?

tests/difftests/runner/src/runner.rs
```diff
@@ -214,75 +214,66 @@ impl ErrorReport {
...
-                temp_output_path.display()
-            );
-            trace!(
-                "Temporary metadata file created at {}",
-                temp_metadata_path.display()
-            );
+            let package_out = self.output_dir.join(&package.relative_path);
+            fs::create_dir_all(&package_out)?;
+            debug!("Writing output to '{}'", package_out.display());
+            let config_file = package_out.join("config.json");
```
LegNeato, line 262: Why are you hardcoding these names now?

tests/difftests/runner/src/testcase.rs
```diff
@@ -0,0 +1,192 @@
...
+use std::fmt::{Display, Formatter};
+use std::fs;
+use std::path::{Path, PathBuf};
+use tracing::debug;
+
+/// A test case containing multiple test binaries that should produce the same output
+pub struct TestCase {
+    /// The name of the testcase, as a rust mod path
+    pub name: String,
+    /// the relative path from the base `difftest` dir
```
LegNeato: Nit caps.

tests/difftests/runner/src/testcase.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub relative_path: PathBuf,
+    /// the absolute path
```
LegNeato: caps

LegNeato (changes requested):
We need docs that `nextest` is now required. I think we should also switch over compiletests so that everything is consistent.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
Firestar99: yep, bad rebase

tests/difftests/runner/src/runner.rs
Firestar99: > * output is now in `tests/difftests/tests/target/difftest` so you can easily inspect the output instead of having to search your tmp dir

And since each test has their own folder, why not hardcode them?

image

Firestar99:
Rebased. Had to drop "difftest: remove glam from workspace, use reeexport via spirv-std" as https://github.com/Rust-GPU/rust-gpu/pull/380 needs to change the features on glam

schell (approved):

eddyb (approved):
With the last commit removed, is there anything blocking landing this?
(assuming that last commit gets moved into its own PR - I've only reviewed the first 15 commits)

Firestar99:
Moved the last "runner cleanup" commit to https://github.com/Rust-GPU/rust-gpu/pull/587

Firestar99:
I'm not seeing the 6s rerun without changes I was expecting, it's more like 2min now for some reason. With `--no-capture` I'm seeing every test run rebuild `rustc_codegen_spirv`. Also [windows difftest](https://github.com/Rust-GPU/rust-gpu/actions/runs/24881612686/job/72851178285?pr=334) is failing to due the `rustc_codegen_spirv.dll` failing to be deleted while in use, which smells and could explain the regression. We never delete the `.so` ourselves, could this be a change within cargo?

tests/difftests/bin/src/main.rs
```diff
@@ -72,77 +49,23 @@ fn run() -> Result<()> {
...
-        .stdout(process::Stdio::inherit());
-    tracing::debug!("Running cargo command: {:?}", cmd);
-
-    let output = cmd.output().expect("build output");
-    let exit_code = output.status.code().unwrap_or(-1);
-    tracing::debug!("Cargo build exited with code {}", exit_code);
-    if !output.status.success() {
-        tracing::error!("Cargo build failed");
-        process::exit(exit_code);
-    }
```
eddyb, line 91: The mention of `--no-capture` made me remember this removed bit of code, was there a specific reason for this?

eddyb:
I'll be honest, I was a bit surprised to find out difftests are so many crates (and each of them its own Cargo package), instead of only needing separate crates (if at all) for the shaders.

If it is Cargo's fault (and nothing `nextest`-specific), then adding `-v`/`--verbose` to the invocation may show the reason for the rebuild.

---

**EDIT**: using this change:
```patch
diff --git a/tests/difftests/runner/src/runner.rs b/tests/difftests/runner/src/runner.rs
index 5005418bc1..d3b0b0297f 100644
--- a/tests/difftests/runner/src/runner.rs
+++ b/tests/difftests/runner/src/runner.rs
@@ -279,1 +279,1 @@
-            cmd.arg("run").arg("--release").arg("--manifest-path").arg(
+            cmd.arg("run").arg("-v").arg("--release").arg("--manifest-path").arg(
```
then running with `--no-capture` results in a few instances of:
```
       Dirty rustc_codegen_spirv v0.10.0-alpha.1 (/home/eddy/Projects/rust-gpu/crates/rustc_codegen_spirv): info of dependency `naga` changed
   Compiling rustc_codegen_spirv v0.10.0-alpha.1 (/home/eddy/Projects/rust-gpu/crates/rustc_codegen_spirv)
```

---

**EDIT2**: this fixes it for me (difftests taking less than 10 seconds to run):
```patch
diff --git a/tests/difftests/tests/lib/Cargo.toml b/tests/difftests/tests/lib/Cargo.toml
index e428da7313..907827c72b 100644
--- a/tests/difftests/tests/lib/Cargo.toml
+++ b/tests/difftests/tests/lib/Cargo.toml
@@ -27,0 +28,4 @@
+[dependencies]
+spirv-std.workspace = true
+glam.workspace = true
+
```

AFAICT, having `spirv-std`/`glam` deps (of individual difftests) _that are not gated by `target_arch = "spirv"`_ leads to different sets of `num-traits` features, and `naga` also depends on `num-traits`, so you end up with at least 2 different builds of `naga`, causing `rustc_codegen_spirv` rebuilds.

Long-term we probably have to adopt something like [`cargo-hakari`](https://github.com/guppy-rs/guppy/tree/main/tools/cargo-hakari), if that's even practical.

Firestar99:
It doesn't auto-merge cause @LegNeato requested changes

LegNeato (approved):

PR #335, opened 2025-07-16
Fix cargo-gpu in build script failing when called by Miri or Clippy

tuguzT:
## cargo-gpu Rust-GPU/cargo-gpu#101

@Firestar99: I've added a new `CargoCmd` struct that should handle all env vars in one central place for both spirv-builder and cargo-gpu install. In case of a failure, it also `debug!` logs the env vars which were removed and which were inherited with their values, to make debugging these sort of issues easier in the future.

As stated in Rust-GPU/cargo-gpu#93, build script which uses `rustc_codegen_spirv` via `cargo-gpu` failed to build shader crate while using Miri.
This pull request fixes this issue by removing `RUSTC_WRAPPER` environment variable from `cargo` call.

### DIFF
--- a/crates/spirv-builder/src/cargo_cmd.rs
+++ b/crates/spirv-builder/src/cargo_cmd.rs
@@ -43,20 +43,20 @@ impl CargoCmd {
         // which could affect its functionality and break the build process.
         cargo.env_remove("RUSTC_WRAPPER");
 
-        // overwritten by spirv-builder anyway
-        cargo.env_remove("CARGO_ENCODED_RUSTFLAGS");
-
+        // NOTE(firestar99, tuguzT) Other environment variables that we don't want to
+        // leak into the inner invocation of Cargo & break the build process.
         cargo
-            .env_remove("RUSTC")
-            .env_remove("RUSTC_WRAPPER")
             .env_remove("RUSTC_WORKSPACE_WRAPPER")
             .env_remove("RUSTFLAGS")
             .env_remove("CARGO")
             .env_remove("RUSTUP_TOOLCHAIN");
 
-        // ignore any externally supplied target dir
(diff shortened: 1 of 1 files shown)


tuguzT:
Please look at Rust-GPU/cargo-gpu#101 for additional context

crates/spirv-builder/src/cargo_cmd.rs
```diff
@@ -0,0 +1,157 @@
...
+
+        // NOTE(tuguzT) Used by Cargo to call executables of Clippy, Miri
+        // (and maybe other Cargo subcommands) instead of `rustc`
+        // which could affect its functionality and break the build process.
+        cargo.env_remove("RUSTC_WRAPPER");
+
+        // overwritten by spirv-builder anyway
+        cargo.env_remove("CARGO_ENCODED_RUSTFLAGS");
+
+        cargo
```
LegNeato, line 48: These seem to be added with no comment, and `RUSTC_WRAPPER` is both here and above.

LegNeato (changes requested):
Thank you for the PR! 🍻

Looks good, just one small thing.

crates/spirv-builder/src/cargo_cmd.rs
tuguzT: I've just removed duplicates from removed env vars and tried to add relevant comments

Firestar99: oh my bad, thanks for fixing it

Firestar99:
Sorry for the constant canceling of CI, see https://github.com/Rust-GPU/rust-gpu/pull/342

I'm waiting for https://github.com/Rust-GPU/rust-gpu/pull/339 to complete and then I can retrigger this one

LegNeato (approved):
Thanks again for the PR 🚀

PR #339, opened 2025-07-17
Add HitTriangleVertexPositionsKHR built-in for ray tracing pipelines

jcmnn:
Vulkan extension: https://registry.khronos.org/vulkan/specs/latest/man/html/HitTriangleVertexPositionsKHR.html
SPIRV extension: https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_ray_tracing_position_fetch.html

Note that this PR only adds the RayTracing[Pipeline] part, not the RayQuery part.

### DIFF
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -138,6 +138,10 @@ const BUILTINS: &[(&str, BuiltIn)] = {
         ("ray_tmax", BuiltIn::RayTmaxKHR),
         ("object_to_world", BuiltIn::ObjectToWorldKHR),
         ("world_to_object", BuiltIn::WorldToObjectKHR),
+        (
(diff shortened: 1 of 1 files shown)


LegNeato:
Thanks for the PR! I don't know enough about this area to review, looks sensible but not sure if other things need to be added. Are you actively testing this?

Firestar99 (approved):
Looks good to me, just squashed your commits (since we don't squash and merge queue doesn't allow you to switch to squashing for individual PRs)

Since the built-in must be of type `Vec3`, I would have loved to see some kind of validation on that, but we don't have a good system yet to do these sort of validations. So let's merge it for now and whenever we get such a system, we can still add that this must be a Vec3.

I may make a followup to also add that capability to ray query.

Firestar99:
Sorry for the constant canceling of CI, see https://github.com/Rust-GPU/rust-gpu/pull/342

PR #341, opened 2025-07-20
Customize `Function` and `BasicBlock` to carry both a SPIR-V ID and an index, for O(1) access.

eddyb:
As may be obvious from the first commit, I was hoping to remove the `RefCell` around the `rspirv::dr::Builder`.
<sub>(it's still possible, but it would require keeping global and function builders separate, and using different ID ranges, and I almost jerry-rigged some contraption using `spirt::{Context,Type,Const}` but it'd be too much work to port everything over, not to mention new exciting failure modes)</sub>

---

Instead, I ended up fixing a long-standing issue with how we expose SPIR-V functions and blocks to `rustc_codegen_ssa` - which is mostly straight-forward, as it already has associated types for `Function` and `BasicBlock`, we were just using raw SPIR-V IDs (or `SpirvValue`, in the case of `Function`) for no good reason.

After this PR, we're always tracking the `rspirv::dr::Builder` function/block indices, not just IDs (while still using IDs to validate that the index didn't go out of sync due to ad-hoc mutation etc.), so there shouldn't be any reason to iterate through functions (or through a function's blocks) searching for a specific ID.

There were already some fast paths before, so this might not have much impact on perf, but I still like the result (if nothing else, it should be harder to misuse, and I even got to remove a silly `call` special-case).

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -3,7 +3,9 @@ use crate::maybe_pqp_cg_ssa as rustc_codegen_ssa;
 
 use super::Builder;
 use crate::abi::ConvSpirvType;
-use crate::builder_spirv::{BuilderCursor, SpirvConst, SpirvValue, SpirvValueExt, SpirvValueKind};
+use crate::builder_spirv::{
+    SpirvBlockCursor, SpirvConst, SpirvValue, SpirvValueExt, SpirvValueKind,
+};
 use crate::codegen_cx::CodegenCx;
 use crate::custom_insts::{CustomInst, CustomOp};
 use crate::spirv_type::SpirvType;
@@ -166,8 +168,8 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                     | MemorySemantics::ACQUIRE_RELEASE
             }
             AtomicOrdering::SeqCst => {
-                let emit = self.emit();
-                let memory_model = emit.module_ref().memory_model.as_ref().unwrap();
+                let builder = self.emit();
+                let memory_model = builder.module_ref().memory_model.as_ref().unwrap();
                 if memory_model.operands[1].unwrap_memory_model() == MemoryModel::Vulkan {
                     invalid_seq_cst = true;
                 }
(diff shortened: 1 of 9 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -2979,16 +2972,6 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
-            // HACK(eddyb) this seems to be needed, but it's not what `get_fn_addr`
-            // produces, are these coming from inside `rustc_codegen_spirv`?
-            SpirvType::Function {
-                return_type,
-                arguments,
-            } => {
-                assert_ty_eq!(self, callee_ty, callee.ty);
-                (callee.def(self), return_type, arguments)
-            }
-
```
eddyb, line 2991: This special-case could be removed because it could only be reached by the custom entry-point logic passing a function as the callee, while `rustc_codegen_ssa` always expects a function *pointer* (this matches up with the `entry_func` -> `self.get_fn_addr(entry_instance)` change in `entry.rs`).

While direct calls should IMO be supported as first-class, that would require `call` to be able to take `Self::Function` as the callee, *not* `Self::Value`, as the only `rustc_codegen_ssa` interface to get a `Self::Value` for a function is `get_fn_addr` (which always creates a pointer).

(This also makes sense if you consider that `Self::Value` is really meant to be an IR value, ideally something vaguely-register-like, that can partake in runtime dataflow, and "a function" doesn't fit that without being referred to indirectly, i.e. as a function *pointer*)

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -831,6 +822,28 @@ impl<'tcx> HasTypingEnv<'tcx> for CodegenCx<'tcx> {
     }
 }
 
+impl<'tcx> CodegenCx<'tcx> {
+    // FIXME(eddyb) a bit odd that this isn't in `rustc_codegen_ssa`, instead of
+    // the pair of `get_fn` + `get_fn_addr`.
+    fn fn_addr_of(&self, span: Span, function: SpirvFunctionCursor) -> SpirvValue {
```
eddyb: Noticing during rebasing later work on top of this PR that this split doesn't actually work, and some other solution is required.

LegNeato (approved):
Not super familiar with this stuff, but the changes look reasonable. There is only one area that I had an issue following / it is sort of convoluted but it is merely to get to asserts so 🤷

PR #342, opened 2025-07-21
ci: fix two PRs with same branch name but different forks canceling each other

Firestar99:
Happened with https://github.com/Rust-GPU/rust-gpu/pull/339 and https://github.com/Rust-GPU/rust-gpu/pull/335 where the branch is named `main` but from two different forks. This should hopefully prevent it in the future, though I won't verify that this actually fixed it since we're already merging one of them.

https://docs.github.com/en/actions/reference/contexts-reference#github-context


LegNeato (approved):
🤷

PR #348, opened 2025-07-24
Make `SpirvValue`(`Kind`) representation and operations more orthogonal and robust (lossless).

eddyb:
`SpirvValue`, and especially `SpirvValueKind`, have been serving several purposes so far:
- tracking the type (i.e. `struct SpirvValue { kind: SpirvValueKind, ty: ... }` before this PR)
- encoding values that we can't (or just don't, currently) emit any SPIR-V for
  - ideally should become `SpirvConst`s (already have some of that done for later PRs)
- indicating that a value is illegal and therefore needs a "zombie" (deferred error) applied
  - in the case of `SpirvValueKind::IllegalConst`, a `Span` isn't known initially at all
    (so there's a need to defer until the time of `SpirvValue::def`, which may bring a `Span`)
- tracking enough information to let `SpirvValue::strip_ptrcasts` undo pointer casts
  - i.e. `SpirvValueKind::LogicalPtrCast` contains the value and type IDs for a `SpirvValue`
    (`SpirvValue { kind: SpirvValueKind::Def(original_ptr), ty: original_ptr_ty }`)

That has grown ad-hoc, and it can be a nightmare to extend any part of it, hence this work.

More specifically, this PR:
- emits all "zombie" (deferred error) *messages* early, and only defers the `Span` at most
  - a new `bool` field (`zombie_waiting_for_span`) replaces `SpirvValueKind::IllegalConst`
  - as a result, `SpirvValue::def` (which is called *everywhere* to obtain a SPIR-V ID)
    only has to check a `bool` (for registering the `Span`), no more string formatting
  - the only reason this might not be an overall perf win, is that pointer casts now have
    to generate the string message themselves, but it's unclear if that was ever skipped
    (and I wasn't able to observe any change in build times, tho more testing is welcome)
- replaces `SpirvValueKind::LogicalPtrCast` with an equivalent `Option` in `SpirvValueKind::Def`
  - by making `SpirvValue` generic and reusing it, `SpirvValue::strip_ptrcasts` becomes
    simpler *and* combining it with `pointercast` is now lossless (wrt other details)
  - in theory, this can now allow a `SpirvValue` that is *simultaneously* like the old
    `SpirvValueKind::IllegalConst` *and* `SpirvValueKind::LogicalPtrCast` at the same time
    (arguably the main motivation for this PR, specifically as a `const_bitcast` result)
- keeps whole `SpirvValue`s in the `SpirvConst`<->SPIR-V ID "const interning" maps
  - while more wasteful space-wise (not even using `SpirvValue`'s new generic param),
    this removes all duplicated logic around "deriving `SpirvValue` for a constant"
    (i.e. what used to decide whether `SpirvValueKind::IllegalConst` should be used)
  - `SpirvValue::const_fold_load` becomes trivial, but even more important, *lossless*
    (returning the same exact `SpirvValue` as the original `SpirvConst` did)

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2512,8 +2512,8 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
             }
 
             // Defer the cast so that it has a chance to be avoided.
-            let original_ptr = ptr.def(self);
-            let bitcast_result_id = self.emit().bitcast(dest_ty, None, original_ptr).unwrap();
+            let ptr_id = ptr.def(self);
+            let bitcast_result_id = self.emit().bitcast(dest_ty, None, ptr_id).unwrap();
 
             self.zombie(
                 bitcast_result_id,
@@ -2528,10 +2528,13 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
 
             SpirvValue {
                 zombie_waiting_for_span: false,
-                kind: SpirvValueKind::LogicalPtrCast {
-                    original_ptr,
-                    original_ptr_ty: ptr.ty,
-                    bitcast_result_id,
+                kind: SpirvValueKind::Def {
+                    id: bitcast_result_id,
+                    original_ptr_before_casts: Some(SpirvValue {
+                        zombie_waiting_for_span: ptr.zombie_waiting_for_span,
+                        kind: ptr_id,
+                        ty: ptr.ty,
+                    }),
                 },
                 ty: dest_ty,
             }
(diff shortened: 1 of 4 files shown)


Firestar99:
Tested on my nanite-at-home project, 3 runs, first run after branch switch discarded (due to many CPU crates recompiling), cmdline:
`rm -rf ./target/spirv-builder/release/ ./target/spirv-builder/spirv-unknown-vulkan1.2/; RUSTGPU_CARGOFLAGS=--timings cargo b --release`

older master 8ee9f2f99788134a9c0912238add509539742596: 16.3s 16.4s 16.6s
current master 87ea628070561f576aeef84c92a0e3e3c25d63ff: 15.9s 15.9s 15.9s (primarily https://github.com/Rust-GPU/rust-gpu/pull/350)
this branch b4ed8448e1e0f9ddeeba2f1999d01bdc76111b62: 15.7s 15.8s 15.9s

This branch is slightly faster :tada:

eddyb:
> This branch is slightly faster 🎉

Wow, that's in the opposite direction from what I heard from @nazar-pc, but it's good to know it can swing either way.

---

In the meanwhile, I *really* want to land certain soundness fixes in `read_from_const_alloc`, which were quite *indirectly* (`const_bitcast` of constant pointers - pun unintended) blocked on this, except *not really*.

So I've marked this PR as draft, and I hope to come up with something isolated, and we can land this later.


**EDIT**: the soundness fix PR is now up:
- https://github.com/Rust-GPU/rust-gpu/pull/353

Issue #347, opened 2025-07-23
`cargo clippy` with spirv target


Firestar99:
Currently, most projects run `cargo clippy` on the CPU target and hope that it covers most gpu-specific code as well. But that will ignore all `#[gpu-only]` functions and entry points, not checking them at all. We should investigate a way we can run clippy with spirv targets. This may be better suited for `cargo gpu`, since `cargo gpu clippy` would work quite nicely.

charles-r-earp:
It looks like clippy sets `RUSTC_WORKSPACE_WRAPPER` and invokes `cargo check`. 

CargoCmd could support applying a SpirvBuilder, which would be most of `invoke_rustc` (omitting the build command, removal of cargo / rustc env vars, and profile, etc). 

```rust
impl CargoCmd {
    pub fn spirv_builder(&mut self, builder: &SpirvBuilder);
}
```

`cargo gpu check` can use this command. `cargo gpu [command] [args..]` maps to `CARGO=cargo-gpu cargo [command] [args..]`.

This means that `cargo gpu clippy` turns into `CARGO=cargo-gpu cargo clippy`, which then invokes clippy on each package (instead of rustc), inserting additional args and env vars from spirv-builder to the command. 

It would be nice to support cargo-expand, useful for  debugging macros. This could use the same mechanism, but invokes `cargo rustc --profile=check -- -Zunpretty=expanded`. So this could be a simple as `cargo gpu expand -p foo --lib`. 

And likewise additional cargo utilities could work without explicitly supporting them.

PR #354, opened 2025-08-01
Update `ShaderPanicStrategy::DebugPrintfThenExit` documentation for Vulkan SDK changes.

eddyb:
I noticed a few changes in the Vulkan Validation Layer documentation and ended up with these items:
- `multimodule` recommendation is no longer necessary (since Vulkan SDK 1.3.275.0):
  - issue: https://github.com/KhronosGroup/SPIRV-Tools/issues/4892
  - PR: https://github.com/KhronosGroup/SPIRV-Tools/pull/5495
- `DEBUG_PRINTF_TO_STDOUT=1` is now `VK_LAYER_PRINTF_TO_STDOUT=1` (since Vulkan SDK 1.3.283.0)
  - old env var still works (just deprecated), but the new one fits in better
  - PR: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/7781
- `VK_KHR_shader_non_semantic_info` isn't required anymore (since Vulkan SDK 1.4.313.0)
  - we should stop mentioning it, but it's recent and I left it in just in case
  - issue: https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9716
  - PR: https://github.com/KhronosGroup/Vulkan-ValidationLayers/pull/9718

I'm not sure how happy I am with the result, but it tries the offer both an easy set of env vars
(`VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation VK_LAYER_PRINTF_ONLY_PRESET=1 VK_LAYER_PRINTF_TO_STDOUT=1`)
and detailed explanations of all the moving parts, in case that doesn't suffice.

Honestly, we should probably have a separate document, and/or move this to `debug_printf!`
(and merely refer to it from this setting), since users trying that out would also need this info.

---

As a sneak peek, this is a screenshot of the new docs (click to open its full size):

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -179,46 +179,72 @@ pub enum ShaderPanicStrategy {
     /// Like `SilentExit`, but also using `debugPrintf` to report the panic in
     /// a way that can reach the user, before returning from the entry-point.
     ///
-    /// Will automatically require the `SPV_KHR_non_semantic_info` extension,
-    /// as `debugPrintf` uses a "non-semantic extended instruction set".
+    /// Quick setup for enabling `debugPrintf` output (to stdout) at runtime:
+    /// - **set these environment variables**:
+    ///   - `VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`
+    ///   - `VK_LAYER_PRINTF_ONLY_PRESET=1`
+    ///   - `VK_LAYER_PRINTF_TO_STDOUT=1` (not always needed, but can help)
+    /// - if using `wgpu`, enable `wgpu::Features::SPIRV_SHADER_PASSTHROUGH`,
+    ///   and use `create_shader_module_passthrough` instead of `create_shader_module`
+    /// - in case of errors, or no output (from a `panic!()`/`debug_printf!()`),
+    ///   keep reading below for additional information and alternatives
     ///
-    /// If you have multiple entry-points, you *may* need to also enable the
-    /// `multimodule` node (see <https://github.com/KhronosGroup/SPIRV-Tools/issues/4892>).
+    /// ---
     ///
-    /// **Note**: actually obtaining the `debugPrintf` output requires:
-    /// * Vulkan Validation Layers (from e.g. the Vulkan SDK)
-    ///   * (they contain the `debugPrintf` implementation, a SPIR-V -> SPIR-V translation)
-    ///   * **set the `VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`
-    ///     environment variable** to easily enable them without any code changes
-    ///   * alternatively, `"VK_LAYER_KHRONOS_validation"` can be passed during
-    ///     instance creation, to enable them programmatically
-    /// * Validation Layers' `debugPrintf` support:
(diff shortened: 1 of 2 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -179,46 +179,72 @@ pub enum ShaderPanicStrategy {
...
+    ///   - defaults to common validation logging (itself defaulting to stdout)
+    ///   - **env**: `VK_LAYER_PRINTF_TO_STDOUT=1` (and its **instance** analogue)
+    ///     forces direct printing to stdout, bypassing `VK_EXT_debug_utils` etc.
+    ///   - validation logging can itself be controlled via `VK_EXT_debug_utils`
+    ///   - `wgpu` built in debug mode (and/or with debug-assertions enabled):
+    ///     - it uses `VK_EXT_debug_utils` internally, exposing it via `log`
+    ///     - with e.g. `env_logger`, `RUST_LOG=info` suffices for `debugPrintf`
+    ///       messages (as they specifically have the "info" level)
+    ///     - other `log`/`tracing` subscribers should be configured similarly
     #[cfg_attr(feature = "clap", clap(skip))]
```
eddyb, line 248: @Firestar99 I noticed this in passing, am I right to think this means `cargo-gpu` cannot make use of this mode? IMO it's invaluable to debugging panics, so we should try to make it easier and more visible (tbf this entire `enum` is user-facing API on top of ~6 possible strings that get passed to `rustc_codegen_spirv` via codegen args, but I'm not sure where it would be best placed).

(oh and I would be fine with not exposing `UNSOUND_DO_NOT_USE_UndefinedBehaviorViaUnreachable` to users at all, I haven't really heard of anyone going for it, so it can stay behind `RUSTGPU_CODEGEN_ARGS` or w/e instead - then the choices implied by `ShaderPanicStrategy` get simpler)

LegNeato:
I think this is probably the first place we need to have a trait for folks to implement or EII or whatever. It should not be this hard to printf from an end-user perspective! But the docs look good.

LegNeato (approved):
🚀

crates/spirv-builder/src/lib.rs
Firestar99: I think I disabled it since the original cargo-gpu (before my refactors) didn't support it and clap didn't work out of the box, the support for enums with values doesn't seem to cleanly work. But it gets even sillier:
* clap is used to parse args from cmdline, which is disabled
* serde is used to read the `[metadata.cargo-gpu]` from the crate's Cargo.toml
* if you use cargo-gpu as a library, you just get a spirv-builder you need to configure manually

I agree this is a mess, and we should eventually figure out a unified way to represent these properties. Since I made https://github.com/Rust-GPU/rust-gpu/issues/347 I thought of two directions to take things:
* encourage the `Cargo.toml` path, so that tools like `cargo gpu clippy` could pick up the properties, and make the library path pick those up by default as well 
* if we make capabilities and extensions not pre-defined but picked up as they are used, as discussed in "link-time capabilites" on zulip, then there's no need to pre-declare anything per crate. If all settings are link-time settings, then there's no need to declare them in the `Cargo.toml` for a `cargo gpu clippy` to pick them up, since it doesn't link anything. And all the linking settings could just be properties of spirv-builder.

Issue #356, opened 2025-08-03
Bindings fail to parse `1u32`


Firestar99:
Defining bindings (or potentially others) should parse `1u32` style integer literals correctly. This is important in proc macros, since `quote!` always adds the type to the end of an integer. Either we use `LitInt` from `syn` or `IntegerLit` from [`litrs`](https://crates.io/crates/litrs).
```rust
error: attribute value must be integer
    --> node-graph/graster-nodes/src/adjustments.rs:58:31
     |
  58 |         #[spirv(descriptor_set = 0, binding = 1u32)] image_input: Image2d,
     |                                     ^^^^^^^^^^^^^^

```

## System Info
- rust-gpu master

LegNeato:
BTW, there is new attribute parsing infra in rustc that is being migrated to...perhaps this should be pushed down into the compiler rather than as a normal proc macro?

PR #357, opened 2025-08-03
allow integer literals to have a suffix

Firestar99:
closes https://github.com/Rust-GPU/rust-gpu/issues/356


LegNeato (approved):
Ah, this is nice and small.

PR #361, opened 2025-08-11
Add Image::fetch_with_lod

### LINKED ISSUE #360: Image fetch with level
When porting my WGSL shader to RustGPU I tried finding the equivalent of WGSL's [`textureLoad`](https://www.w3.org/TR/WGSL/#textureload).
The closest thing I could find was [spirv_std::image::Image::fetch](https://rust-gpu.github.io/rust-gpu/api/spirv_std/image/struct.Image.html#method.fetch), which does not have a mipmap level argument.

I tried using `Image::fetch` since I was using `textureLoad` with level = 0 and figured it might work.
I translated the SPIR-V to WGSL with naga, but it generated an invalid call to textureFetch, missing the required level argument.

I'm not sure if it makes sense to add `level: u32` parameter to `Image::fetch` or provide a new `Image::fetch_with_level`.

caibear:
Fixes #360

I used the terminology `level` since that's what WGSL used, but it would also make sense to use `lod` since that's what SPIR-V and GLSL's `texelFetch` use.

### DIFF
--- a/crates/spirv-std/src/image.rs
+++ b/crates/spirv-std/src/image.rs
@@ -174,22 +174,7 @@ impl<
     where
         I: Integer,
     {
-        let mut result = SampledType::Vec4::default();
-        unsafe {
-            asm! {
-                "OpDecorate %image NonUniform",
-                "OpDecorate %result NonUniform",
-                "%image = OpLoad _ {this}",
-                "%coordinate = OpLoad _ {coordinate}",
-                "%result = OpImageFetch typeof*{result} %image %coordinate Lod {lod}",
-                "OpStore {result} %result",
-                result = in(reg) &mut result,
-                this = in(reg) self,
-                coordinate = in(reg) &coordinate,
-                lod = in(reg) lod,
-            }
-        }
-        result.truncate_into()
+        self.fetch_with(coordinate, sample_with::lod(lod))
     }
 }
 
--- a/crates/spirv-std/src/image.rs
+++ b/crates/spirv-std/src/image.rs
@@ -162,6 +162,35 @@ impl<
         }
         result.truncate_into()
     }
+
+    /// Fetch a single texel at a mipmap `level` with a sampler set at compile time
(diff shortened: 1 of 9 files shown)


Firestar99 (changes requested):
You can already fetch image texels with additional settings, like lod, bias, grad and sampleIndex. And naga for some reason requires explicitly fetching from lod 0, like below.

```rust
use spirv_std::image::sample_with::lod;
use spirv_std::image::{Image2d, ImageWithMethods};

mask_texture.fetch_with(frag_coord.xy().as_uvec2(), lod(0))
```

Also I'd call it lod since we're following spirv / glsl naming conventions. 

You may also find this thing I've been working on useful: https://github.com/Rust-GPU/rust-gpu/pull/280

LegNeato:
Not required, but it would be nice to put "level" somewhere as a doc alias (or even in the comment) so it comes up if folks search the webgpu terminology.

schell (approved):
Yup, this makes sense now 👍

LegNeato:
Looks like some compiletests are failing. You should be able to run them locally with `cargo compiletest`.

LegNeato (changes requested):
CI failures

caibear:
> CI failures

The test I added passes on my local machine. The CI failures look unrelated to the changes I made, so I'm not sure how to fix them. The tests that are failing appear to be tests that are supposed to cause compile errors being erroneously marked as failing.

```
failures:
    [ui] ui/image/gather_err.rs
    [ui] ui/image/query/query_levels_err.rs
    [ui] ui/image/query/query_lod_err.rs
    [ui] ui/image/query/query_size_err.rs
    [ui] ui/image/query/query_size_lod_err.rs
```

LegNeato:
It looks like the line numbers when it refers to that file changed, so it affects other tests that refer to those files with error messages. You can run `cargo compiletest --bless` to update the tests I believe

PR #362, opened 2025-08-11
Try dumping, when a SPIR-T pass panics, its in-progress `spirt::Module` state.

eddyb:
While SPIR-T passes should, in principle, never panic, it can still happen (esp. with WIP code, where e.g. `assert!`s or `todo!()`s might be overrepresented etc.), and before this PR, the partially modified SPIR-T module would be completely lost in the `--dump-spirt-passes` output.

And because incomplete SPIR-T mutation could result in extremely malformed modules, this PR also wraps the SPIR-T pretty-printing in `std::panic::catch_unwind`, removing one pass' "after" state, at a time, from the pretty-printed versions, while it still panics, and showing an extra warning (with all the pass names that couldn't be printed) if that was necessary at all.

Also, there's a minor ergonomic fix in that `after_pass` doesn't need to also be passed the same string as `before_pass` (since `SpirtDumpGuard` is tracking it anyway).

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -29,6 +29,7 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::ErrorGuaranteed;
 use rustc_session::Session;
 use rustc_session::config::OutputFilenames;
+use std::cell::Cell;
 use std::collections::BTreeMap;
 use std::ffi::{OsStr, OsString};
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


eddyb:
> I can't help but think this should be `tracing` spans and items

That would be useful for some stuff, but, even ignoring the `rustc` "timers"...
<sub>(that we only use an inefficient ad-hoc form of, rn, which is only accessible via `RUSTGPU_RUSTFLAGS=-Ztime-passes`, while the more advanced `measureme` APIs are the low-overhead/high-throughput event tracking powering `-Zself-profile`, which the `log`/`tracing` approach can't really ever compete with AFAIK, sadly - but really, few things need that level of profiling, compilers being deterministic and IO-bound is an edge case)</sub>
the most important part here is generating the `.spirt`/`.spirt.html` dumps, where the "pass names" show up as table headings - quick example from a dump I have open:
image

I wouldn't mind automating those strings, but right now they're quite arbitrary, e.g. after https://github.com/Rust-GPU/spirt/pull/25 the `mem::analyze_accesses` one comes from:
```rust
let profiler = before_pass("mem::analyze_accesses", module);
spirt::passes::qptr::analyze_mem_accesses(module, &SPIRT_MEM_LAYOUT_CONFIG);
after_pass(Some(module), profiler);
```
with `spirt::passes::qptr::analyze_mem_accesses` defined as:
```rust
// FIXME(eddyb) this doesn't really belong in `qptr`.
pub fn analyze_mem_accesses(module: &mut Module, layout_config: &crate::mem::LayoutConfig) {
    crate::mem::analyze::GatherAccesses::new(module.cx(), layout_config)
        .gather_accesses_in_module(module);
}
```
So there's not even any function named `analyze_accesses` in a module named `mem`, at best that could be the case for `spirt::passes::mem::analyze_accesses` if that function was moved to such a place.

But even if we take the function's path and strip either `spirt::passes::` or `rustc_codegen_spirv::linker::spirt_passes::` as a prefix (which does sound relatively reasonable, I will admit), how could we obtain that information when that function is in another crate?

LegNeato:
You can create trace spans programmatically with custom names, it doesn't have to be a function nor a proc macro. Tracing is in rustc proper and lots of crates (not sure what is actually instrumented though).

But yeah, it doesn't help with table headings unless you installed your own subscriber and collected and then output (that might actually be interesting).

LegNeato (approved):

Issue #312, opened 2025-07-01
Unclear ICE


nazar-pc:
### Example & Steps To Reproduce
I have pushed the code that triggers this into https://github.com/nazar-pc/abundance/tree/rust-gpu-ice

Reproduction:
```bash
cd crates/farmer/ab-proof-of-space-gpu
cargo build
```

`cargo-gpu` will attempt to build the crate according to manifest in `crates/farmer/ab-proof-of-space-gpu/rust-gpu-workaround`.

Things compiled before last commit, but adding `ab-blake3`+`blake3` to dependencies somehow causes multiple issues it seems.

There seems to be a lot going wrong there 😬
Feel free to change issue title to something that makes more sense, I didn't know what to put there in this case.

## System Info

 - Rust: 1.84.0-nightly (b19329a37 2024-11-21)
 - OS: Ubuntu 24.04.2
 - rust-gpu: https://github.com/Rust-GPU/rust-gpu/commit/f58374079a72ee1f73bf2a9bdaed435b5e44e20e
 - cargo-gpu: https://github.com/Rust-GPU/cargo-gpu/commit/e5744605020d9a94931519a17815e027db5fba01

## Backtrace

<details><summary>Backtrace</summary>
<p>

```
   Compiling ab-proof-of-space-gpu v0.1.0 (/web/github/abundance/crates/farmer/ab-proof-of-space-gpu)
error: failed to run custom build command for `ab-proof-of-space-gpu v0.1.0 (/web/github/abundance/crates/farmer/ab-proof-of-space-gpu)`

Caused by:
  process didn't exit successfully: `/home/nazar-pc/.cache/cargo/target/debug/build/ab-proof-of-space-gpu-fc15fec78da04eb1/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=RUSTGPU_CODEGEN_ARGS
  cargo:rerun-if-env-changed=RUSTGPU_RUSTFLAGS
  cargo:rerun-if-env-changed=RUSTGPU_CARGOFLAGS
  cargo:rerun-if-env-changed=RUSTGPU_CODEGEN_ARGS
  cargo:rerun-if-env-changed=RUSTGPU_RUSTFLAGS
  cargo:rerun-if-env-changed=RUSTGPU_CARGOFLAGS

  --- stderr
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
      Blocking waiting for file lock on package cache
[...]

PR #363, opened 2025-08-14
Make `black_box` a no-op

LegNeato:
Part of https://github.com/Rust-GPU/rust-gpu/issues/312

### DIFF
--- a/tests/difftests/tests/Cargo.lock
+++ b/tests/difftests/tests/Cargo.lock
@@ -124,6 +124,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "black_box-noop-with"
+version = "0.0.0"
+dependencies = [
+ "difftest",
+ "spirv-std",
+]
+
+[[package]]
+name = "black_box-noop-without"
+version = "0.0.0"
+dependencies = [
+ "difftest",
+ "spirv-std",
+]
+
+[[package]]
 name = "block"
 version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/tests/difftests/tests/Cargo.toml
+++ b/tests/difftests/tests/Cargo.toml
@@ -38,6 +38,8 @@ members = [
     "lang/core/ops/trig_ops/trig_ops-wgsl",
     "lang/core/ops/vector_swizzle/vector_swizzle-rust",
     "lang/core/ops/vector_swizzle/vector_swizzle-wgsl",
+    "lang/core/intrinsics/black_box_noop/with-black-box",
+    "lang/core/intrinsics/black_box_noop/without-black-box",
 ]
 
 [workspace.package]
--- a/tests/difftests/tests/lang/core/intrinsics/black_box_noop/with-black-box/Cargo.toml
+++ b/tests/difftests/tests/lang/core/intrinsics/black_box_noop/with-black-box/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "black_box-noop-with"
+edition.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+crate-type = ["dylib"]
+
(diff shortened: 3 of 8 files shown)


crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,6 +240,23 @@ impl<'a, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'tcx> {
 
             sym::ctpop => self.count_ones(args[0].immediate()),
             sym::bitreverse => self.bit_reverse(args[0].immediate()),
+            sym::black_box => {
+                // TODO(LegNeato): do something more sophisticated that prevents DCE
+                match args[0].val {
+                    // Return the value unchanged for immediates
+                    OperandValue::Immediate(v) | OperandValue::Pair(v, _) => v,
+                    // For non-immediates, store and reload
+                    OperandValue::Ref(place) => self.load(
```
LegNeato: I guess?

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,6 +240,26 @@ impl<'a, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'tcx> {
...
             sym::ctpop => self.count_ones(args[0].immediate()),
             sym::bitreverse => self.bit_reverse(args[0].immediate()),
+            sym::black_box => {
+                // TODO(LegNeato): do something more sophisticated that prevents DCE
+                self.tcx
+                    .dcx()
+                    .warn("black_box intrinsic does not prevent optimization in Rust GPU");
+                match args[0].val {
+                    // Return the value unchanged for immediates
+                    OperandValue::Immediate(v) | OperandValue::Pair(v, _) => v,
```
LegNeato: I'm not sure how to handle pair, probably just unsupported?

crates/rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,6 +240,33 @@ impl<'a, 'tcx> IntrinsicCallBuilderMethods<'tcx> for Builder<'a, 'tcx> {
...
+                let llty = layout.spirv_type(self.span(), self);
+
+                match args[0].val {
+                    // Pass through scalars
+                    OperandValue::Immediate(v) => v,
+
+                    // Preserve both elements by spilling + reloading
+                    OperandValue::Pair(..) => {
+                        let tmp = self.alloca(layout.size, layout.align.abi);
+                        self.store(args[0].immediate(), tmp, layout.align.abi);
```
Firestar99: The implementation of `args[0].immediate()` is guaranteed to panic, see my latest additions to the test ICE-ing:

```rust
    pub fn immediate(self) -> V {
        match self.val {
            OperandValue::Immediate(s) => s,
            _ => bug!("not immediate: {:?}", self),
        }
    }
```

Our ABI takes apart ScalarPairs into an `Adt(a, b)`, but only if the ScalarPair consists out of two non-ZST fields. I have no idea how to correctly load / store ScalarPairs, especially given that condition. Maybe @eddyb can help us here?
https://github.com/Rust-GPU/rust-gpu/blob/bfce65327c55fa35a8a1a06f9afd353751cb3560/crates/rustc_codegen_spirv/src/abi.rs#L567

LegNeato: Yeah, I've been feeling around in the dark, no clue here.

LegNeato: I think I found the right API.

Firestar99 (approved):
Looks good!

I just squashed the fixup commits to not pollute main too much

Issue #276, opened 2025-06-10
spirv-tools v2025.2 makes compiletest fail


Firestar99:
Newer spirv-tools v2025.2 makes our compiletest fail:
`cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.2`
```
failures:
    [ui] ui/lang/consts/issue-1024.rs
    [ui] ui/lang/consts/nested-ref.rs
    [ui] ui/lang/consts/shallow-ref.rs
    [ui] ui/lang/control_flow/for_with_custom_range_iter.rs
    [ui] ui/lang/core/ops/range-contains.rs
```

All tests fail due to something similar to
```
error: error:0:0 - [VUID-StandaloneSpirv-None-10684] Invalid explicit layout decorations on type for operand '25[%_ptr_Private__arr_bool_uint_2]'
         %3 = OpVariable %_ptr_Private__arr_bool_uint_2 Private %27
  |
  = note: spirv-val failed
  = note: module `$TEST_BUILD_DIR/lang/consts/issue-1024.vulkan1.2`
```

## System Info

 - rust-gpu master 77070b70b4af3afd7d12d980f6c997e9b77e6f9c
 - SPIR-V: SPIRV-Tools v2025.2 v2025.2.rc2-0-ga62abcb4

SiebenCorgie:
I could track the problem to [this](https://github.com/Rust-GPU/rust-gpu/blob/bd1a6fc5cb05de3bf00d13c22e086a3aac0bc12e/crates/rustc_codegen_spirv/src/builder_spirv.rs#L689) line. The `StorageClass::Private` is apparently wrong if constant arrays of composite-types are involved.


I think the overarching problem is that constant values are are loaded into a var before accessing them like this:
```
//Some constant 
%35 = OpConstantComposite ...

%1 = OpFunction %void None %23
         %45 = OpLabel
         //PreAllocation with _wrong_ StorageClass in header of function
         %46 = OpVariable %PtrConstantsType Function
         
          //eventually store the constant into %46 ...
          OpStore %46 %35
         
         //And then load it back from the var in order to read it.
         %58 = OpInBoundsAccessChain %PtrConstantsType %46 %53
         //Use of %58 ...
```

The *right* way in this case would be to use `%35` to access whatever `%58` holds atm. directly instead.

Edit: I feel like `mem2reg` should catch that pattern already 👀

PR #379, opened 2025-09-08
Erase explicit layout decorations (`Offset`/`ArrayStride`) when disallowed by Vulkan.

eddyb:
More recent Vulkan SDK versions have started complaining about types having explicit memory layout, when used with memory that doesn't inherently require explicit layouts (only push constants and buffers really do), e.g.:
```
error: line 1097: [VUID-StandaloneSpirv-None-10684] Invalid explicit layout decorations on type for operand '367[%_ptr_Function__struct_61]'
  %1207 = OpVariable %_ptr_Function__struct_61 Function
```

The solution used here is simpler and more specialized (a pass that erases the explicit layout decorations from types behind pointers in storage classes that don't support them, combined with updating affected loads/stores) than what I originally described in https://github.com/Rust-GPU/rust-gpu/issues/266#issuecomment-3196953157, so we can hopefully use this for now and not worry about `spirt::{mem,qptr}` for the next release.

- fixes https://github.com/Rust-GPU/rust-gpu/issues/276
- fixes https://github.com/Rust-GPU/rust-gpu/issues/266
  (not actually tested with MoltenVK, we should make sure it actually works now)
- see also https://github.com/KhronosGroup/SPIRV-Tools/issues/6193

---

**TODO**: before landing this, we should update `spirv-tools-rs` as well (I only changed the Vulkan SDK version that is used for CI, via the `use-installed-tools` feature).


.github/workflows/ci.yaml
```diff
@@ -27,13 +27,13 @@ jobs:
...
         with:
-          version: 1.4.309.0
+          version: 1.4.321.0
           cache: true
       - if: ${{ runner.os == 'Windows' }}
         name: Install Vulkan Runtime with SwiftShader (Windows)
         uses: NcStudios/VulkanCI@v1.2
         with:
-          sdkVersion: 1.4.309.0
+          sdkVersion: 1.4.321.0
```
eddyb: At the very least the Windows CI doesn't work, because this `NcStudios/VulkanCI` action is stuck on specific versions of the Vulkan SDK, not sure what to do about that - can we get SwiftShader from elsewhere?

LegNeato: I only did swiftshader to work around a windows bug in their fake GPU thing. Actions are updating (https://github.com/actions/runner-images/issues/12677) so maybe it isn't even needed now?

LegNeato: (I also put up a PR for the action but who knows when it will get merged)

eddyb: Based on the CI failures for https://github.com/Rust-GPU/rust-gpu/pull/400, it looks like a HLSL syntax error in `wgpu`/Naga's output (specifically when nested arrays are involved), and having SwiftShader just means `wgpu`'s Vulkan backend is used instead.

Windows 2025 changes nothing, which doesn't surprise me, now that I've seen what's going on.
Will try installing the Vulkan SDK via [`jakoch/install-vulkan-sdk-action`](https://github.com/jakoch/install-vulkan-sdk-action) and see if that passes tests.

**EDIT**: 5th time's the charm, it just worked!

Firestar99:
I just want to note that you're upgrading CI's Vulkan SDK to 1.4.321.0, but `use-compiled-tools` is still on 1.4.309.0, so we could have a scenario where the old SDK fails where the new SDK passes. I know that [upgrading spirv-tools](https://github.com/Rust-GPU/spirv-tools-rs/pull/19) isn't trivial this time, so I would let this pass as is, just something to be aware of.

eddyb:
> **TODO**: before landing this, we should update `spirv-tools-rs` as well (I only changed the Vulkan SDK version that is used for CI, via the `use-installed-tools` feature).

@Firestar99 I 100% agree, that's why I still have the above quoted bit in the PR description, and it's why this PR was in draft state even before rebasing it on top of #400 (I know it'd be slower but it'd be nice to test `use-compiled-tools` as well, not just `use-installed-tools`, in CI).

**EDIT**: fixed the `spirv-tools-rs` upgrade and landed it:
- https://github.com/Rust-GPU/spirv-tools-rs/pull/19

eddyb:
As the `spirv-tools-rs` update PR was merged:
- https://github.com/Rust-GPU/spirv-tools-rs/pull/19

I was able to get 10 failures in `cargo compiletest` by using it and turning off the new pass:
```patch
diff --git a/Cargo.toml b/Cargo.toml
index 7cf09984100..9da910407c9 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,3 @@ spirv-std-types = { path = "./crates/spirv-std/shared", version = "=0.9.0" }
 spirv-std-macros = { path = "./crates/spirv-std/macros", version = "=0.9.0" }
-spirv-tools = { version = "0.12.1", default-features = false }
+spirv-tools = { version = "0.12.1", git = "https://github.com/Rust-GPU/spirv-tools-rs", default-features = false }
 rustc_codegen_spirv = { path = "./crates/rustc_codegen_spirv", version = "=0.9.0", default-features = false }
diff --git a/crates/rustc_codegen_spirv/src/linker/mod.rs b/crates/rustc_codegen_spirv/src/linker/mod.rs
index d289c8e5fce..89c46e11f03 100644
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -564,3 +564,3 @@ pub fn link(
 
-        {
+        if false {
             let timer = before_pass("spirt_passes::explicit_layout::erase_when_invalid");
```

Re-enabling the pass (removing `if false`), while keeping the newer `spirv-tools-rs`, passes `cargo compiletest`.

---

So we should be able to land this once we release another version of `spirv-tools-rs`, AIUI.

Firestar99:
Should I release spirv-tools for you?

eddyb:
> Should I release spirv-tools for you?

Yes, that would be great, thanks!
<sub>(sorry for not being clearer - and also missing the notif for this - alternatively I could handle the release if it's easier for you, but I was worrying of subtleties around that specific project, I don't think we have a very unified flow yet across repos)</sub>

Firestar99:
(welp and I completely missed your message)

Anyway, I thought I'd give spirv-tools some love before we release:
* https://github.com/Rust-GPU/spirv-tools-rs/pull/20
* https://github.com/Rust-GPU/spirv-tools-rs/pull/21

Afterwards, a release should just be a trivial `cargo release minor` in the root workspace

eddyb:
> 🎯 [Cache] Restored Vulkan SDK in path: '/home/runner/vulkan-sdk'. Cache Restore ID: 'cache-linux-x64-vulkan-sdk-1.4.309.0'.
> Warning: Vulkan SDK path doesn't exist: /home/runner/vulkan-sdk/1.4.321.0/x86_64
> Warning: Could not find Vulkan SDK in /home/runner/vulkan-sdk/1.4.321.0/x86_64

This is very confusing. It doesn't seem to actually ignore the cache if it's the wrong version. Not sure why only Linux seems to be affected. Will purge GHA caches for old version and see what happens.

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
```diff
@@ -0,0 +1,860 @@
...
+        }
+    }
+}
+
+impl<'a> SelectiveEraser<'a> {
+    fn addr_space_allows_explicit_layout(&self, addr_space: AddrSpace) -> bool {
+        let wk = self.wk;
+
+        // FIXME(eddyb) this might need to include `Workgroup`, specifically when
+        // `WorkgroupMemoryExplicitLayoutKHR` is being relied upon.
```
Firestar99, line 409: Shouldn't Workgroup be in this list, as https://github.com/Rust-GPU/rust-gpu/issues/266 shows the declarations need to be removed? I'd be fine not supporting `WorkgroupMemoryExplicitLayoutKHR` for the moment, haven't heard of the extension myself nor anyone using it. If somebody wants support, we can always add it later.

* VK extension: https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VK_KHR_workgroup_memory_explicit_layout
* Coverage [25.68%](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_workgroup_memory_explicit_layout&platform=all)	[74.32%](https://vulkan.gpuinfo.org/listdevicescoverage.php?extension=VK_KHR_workgroup_memory_explicit_layout&platform=all&option=not) is quite good

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
```diff
@@ -0,0 +1,860 @@
...
+            _ => false,
+        };
+        if needs_erasure_of_explicit_layout {
+            ty_def.inner_transform_with(&mut EraseExplicitLayout(self))
+        } else {
+            ty_def.inner_transform_with(self)
+        }
+    }
+
+    fn in_place_transform_global_var_decl(&mut self, gv_decl: &mut GlobalVarDecl) {
```
Firestar99, line 163: spirt question: I don't fully understand why some visitor functions are `*_in_place` and others are not and returned `Transformed<T>` instead

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
```diff
@@ -0,0 +1,860 @@
...
+        let needs_erasure_of_explicit_layout = match &ty_def.kind {
+            TypeKind::SpvInst {
+                spv_inst,
+                type_and_const_inputs: _,
+            } if spv_inst.opcode == wk.OpTypePointer => match spv_inst.imms[..] {
+                [spv::Imm::Short(sc_kind, sc)] => {
+                    assert_eq!(sc_kind, wk.StorageClass);
+                    !self.addr_space_allows_explicit_layout(AddrSpace::SpvStorageClass(sc))
+                }
+                _ => unreachable!(),
```
Firestar99, line 151: spirt question: I also have no idea if this unreachable is actually unreachable, or how exactly `imms` and `type_and_const_inputs` works, since they're unfortunately not documented.

Firestar99 (commented):
I feel like I can't really comment too much on the code, as I'm totally unfamiliar with spirt. I should probably start reading into the library a bit *at some point*, since I assume most of our current passes will eventually end up in spirt anyway.

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
eddyb: A SPIR-V "immediate" (in the `spirt::spv::Imm` sense) is similar to `rspirv::dr::Operand` w/o the `Id` variants (as SPIR-V IDs always correspond to some `Type`/`Const`/`GlobalVar`/`Func`/etc. interned/owned entity in SPIR-T - in the case of types, `type_and_const_inputs` lets them refer to other `Type`s and/or `Const`s, through SPIR-V IDs, but nothing else).

So hitting this `unreachable!()` would require interning a type with a `spirt::TypeKind::SpvInst` that:
- couldn't naturally be produced by SPIR-V -> SPIR-T (`spirt::spv::lower`)
- would fail SPIR-T -> SPIR-V (`spirt::spv::lift`)
- might error/panic in other parts of SPIR-T (that e.g. decode a `spv::Inst`'s `spv::Imm`s according to the SPIR-V "grammar")

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
eddyb: It's a combination of 2-3 things:
- wanting `Transformed`'s distinction of whether a chance is actually needed, to bypass work
  (but that requires by-value returns to avoid misuse)
- anything interned (`AttrSet`/`Const`/`Type`) can't do in-place mutation
  (so they use `&FooDef -> Transformed<FooDef>` + only re-interning when changed)
- module-owned entities (`GlobalVar`/`Func` and intra-func regions/nodes) can do in-place mutation
  (and cloning the whole definition to return it via `Transformed` would be prohibitively expensive)

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
eddyb: The list is for the storage classes that allow and/or require the "explicit layout" (`Offset`/`ArrayStride`) decorations.

So everything missing here (like `Function`, `Private`, `Workgroup`, or weird storage classes used only within raytracing pipelines etc.) will have the decorations removed.

Maybe it's a bit confusing that the calls to this method look like:
```rust
!self.addr_space_allows_explicit_layout(...)
```

But it's fundamentally a small set of storage classes for which the decoderations are allowed:
> This chapter describes valid uses for a set of SPIR-V decorations. Any other use of one of these decorations is invalid,

(from https://docs.vulkan.org/spec/latest/chapters/interfaces.html - Ctrl+F `ArrayStride` on that page to find all the instances where they explicitly allow - or rather, require - explicit layout)

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
```diff
@@ -0,0 +1,860 @@
...
+            attrs: self.0.cx[old_attrs]
+                .attrs
+                .iter()
+                .filter(|attr| {
+                    // FIXME(eddyb) `rustfmt` breaks down for `matches!`.
+                    #[allow(clippy::match_like_matches_macro)]
+                    let is_explicit_layout_decoration = match attr {
+                        Attr::SpvAnnotation(attr_spv_inst)
+                            if (attr_spv_inst.opcode == wk.OpDecorate
+                                && [wk.ArrayStride, wk.MatrixStride]
```
eddyb, line 829: Oh, phew, I was worried this only mentioned `ArrayStride`, but I did include `MatrixStride` just in case (I guess the one instance of non-explicit-layout matrices would have to bein the context of raytracing pipelines - e.g. Ctrl+F "matrix" on https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_ray_tracing.html - though an even simpler example is just a local variable containing a matrix).

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
Firestar99: Oh yea, I thought it was a list of the opposite

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
Firestar99: From what I could gleam from spirt source:
* spirt does not use the vulkan headers to extract spec information
  * instead of running codegen from spec detail, it treats spec information as runtime values
  * only a few basic types that are needed for legalization are found in the source code
  * the rest of the operations are just forwarded as unknown entities it won't touch (like texture accesses)
  * you are still including the spec and loading it at runtime to know all possible instructions and what type they are (OpType, OpConst) for lift and lower, without actually knowing exactly what they do
* `Imm::Short(u32)` is sufficient for all usual operands, as they are all just u32-sized enums
* Exception: string literals, for which you use one `Imm::LongStart` with many `Imm::LongCont[inued]` to encode a stream of N-many bytes
  * I have not found any other case using `Long`

Thoughts
* With `Imm::Long*` rarely used but `Imm::Short` ubiquitous, I wonder if an `Imm::unwrap_short()` could make code more readable? And maintainable, since it would allow you to replace the backing datastructure later without much code breakage.
* With `Imm::Long` being so rare and I assume untouched by spirt, I wonder if a plain `Vec<u8>` or `&'cx [u8]` is more efficient and simpler? That would also make `OpString` only have 2 `Imm`s, which makes it fit in `Inst.imms: SmallVec<[Imm; 2]>` and only be one indirection to access, as previously.
```rust
pub enum Imm {
    Short(spec::OperandKind, u32),
    Long(spec::OperandKind, Vec<u32>),
}
```

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
Firestar99: I wanted to try out if I could somehow verify that `MatrixStride` is actually removed from the struct...
```
error: error:0:0 - Structure id 12 decorated as Block must be explicitly laid out with MatrixStride decorations.
         %_struct_12 = OpTypeStruct %mat4v3float
```
Well.. you can't remove what isn't there in the first place...

Our matrix support is really bad and has similar screw up like the Vec's before https://github.com/Rust-GPU/rust-gpu/pull/380, but since matrix are barely useful I think it just isn't a priority rn. The usecases are:
* that RT extension
* there's a few basic matrix instructions in SPIRV we don't emit but build them from basic instructions, just search `OpTypeMatrix` in the spec
  * `OpTranspose`
  * `OpMatrixTimesScalar`
  * `OpVectorTimesMatrix`
  * `OpMatrixTimesVector`
  * `OpMatrixTimesMatrix`
  * `OpOuterProduct`

crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs
eddyb: These are all good ideas (I also thought of interning immediates, and a few other possible designs, but without proper benchmarking I didn't want to start fine-tuning too much), could you copy that into an issue on the `spirt` repo? (it already feels formulated like close to that)

(this also reminds me there were a few issues on the old repo, idk if they got auto-migrated or not, I don't remember checking)

---

> spirt does not use the vulkan headers to extract spec information
> instead of running codegen from spec detail, it treats spec information as runtime values
> ... 
> the rest of the operations are just forwarded as unknown entities it won't touch (like texture accesses)
you are still including the spec and loading it at runtime 

As for this part, the reason the JSON is still dynamically loaded is that I found some really neat ways to exploit the way Khronos assigns numbers (opcode, enumerands, etc. - see `spirt::spv::spec::indexed`), and some of that data could be neatly loaded as a binary blob (emitted by e.g. a build script), but it's hard to do it for everything and you easily run into build-host-vs-target potential issues etc.
(I know of at least one place tackling this kind of problem in its generality - `icu4x` and I believe the implementation is mostly in [`zerovec`](https://docs.rs/zerovec), which is very cool but somewhat daunting)

By now the SPIR-V spec is so large (in a "breadth" sense) that generating Rust code for everything feels quite wasteful - most of it is closer to schema/protocol/interface descriptions, and I still wish they had made the format self-descriptive for the most part.

nazar-pc:
CI failed on `main` with this PR merged into it

eddyb:
> CI failed on `main` with this PR merged into it

I was worried about that, I guess I should've kept an eye on it - it's just a cache filtering issue, for some reason on Linux it doesn't ignore the wrong version cache (even though I can't find a Linux/not-Linux conditional in that Vulkan SDK action, that could explain this).

Nuking old GHA caches and retrying, if it starts working it should keep working AFAICT.

**EDIT**: retry passed, that should be it for now.

PR #381, opened 2025-09-08
Support scalar pair ABI

LegNeato:
(T, U) pairs in entrypoints and regular functions are now supported.

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -11,7 +11,10 @@ use rspirv::dr::Operand;
 use rspirv::spirv::{
     Capability, Decoration, Dim, ExecutionModel, FunctionControl, StorageClass, Word,
 };
-use rustc_codegen_ssa::traits::{BaseTypeCodegenMethods, BuilderMethods, MiscCodegenMethods as _};
+use rustc_codegen_ssa::traits::{
+    BaseTypeCodegenMethods, BuilderMethods, ConstCodegenMethods, LayoutTypeCodegenMethods,
+    MiscCodegenMethods as _,
+};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::MultiSpan;
 use rustc_hir as hir;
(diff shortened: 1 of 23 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -645,21 +652,54 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                        ),
+                    };
+                    let b_offset = a
+                        .primitive()
+                        .size(self)
+                        .align_to(b.primitive().align(self).abi);
+
+                    let elem0_ty = self.scalar_pair_element_backend_type(layout, 0, false);
+                    let elem1_ty = self.scalar_pair_element_backend_type(layout, 1, false);
+
```
LegNeato, line 691: @eddyb I am not sure this section is right, can you review?

Firestar99:
I'll need this PR for https://github.com/Rust-GPU/rust-gpu/pull/380, since removing our custom `#[repr(SIMD)]` will turn `Vec2` into a `ScalarPair` and fail

LegNeato:
That is fortuitous 🤯

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -645,21 +652,54 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                    let base_ptr = value_ptr.unwrap();
+                    let ptr1 = bx.inbounds_ptradd(base_ptr, self.const_usize(b_offset.bytes()));
```
Firestar99, line 693: Why use `inbounds_ptradd` here, when it just forwards to `inbounds_gep`?

```rust
fn inbounds_ptradd(&mut self, ptr: Self::Value, offset: Self::Value) -> Self::Value {
    self.inbounds_gep(self.cx().type_i8(), ptr, &[offset])
}
```

I'm also not so sure you want the resulting type to implicitly be an `i8`?

LegNeato: Yeah, this is what I am not sure about, doing pointer arithmetic is bad and I was going to switch it to gep but wanted eddyb to chime in.

nazar-pc:
> (T, U) pairs in entrypoints and regular functions are now supported.

Tuple layout is not guaranteed in Rust though. So at least for entrypoints I'd prefer for it to not be supported to avoid UB, forcing user to create an explicit `#[repr(C)]` struct with fields in desired order. For internal functions it is fine since compiler controls both the callee and caller.

In fact, any type that is not FFI-safe should generate at least compiler warning if not error, just like when you use `extern "C" fn`.

LegNeato:
Yeah, I know the repr is not guaranteed in Rust but as you can see in the example we are pushing each piece and then aggregating as a pair on the device side. I guess that is UB though?

nazar-pc:
That is certainly not how I'd write in an actual host-side code, I'd simply cast `#[repr(C)]` array or slice to a slice of bytes instead.

If manually pushing bytes in correct order *and* adding an extra guarantee about tuple order on shader entrypoint then it might not be UB, but it feels unnatural in Rust.

LegNeato:
Agreed, probably best to be safe and restrict them from entrypoints.

Firestar99:
This is a much more fundamental issue. You can pass structs through buffers that are `#[repr(rust)]` (the default repr), which also doesn't have any layout guarantees. So I'd suggest we merge this scalar abi patch as is, that by chance also enables tuples in entry points, and move this to a new issue.

tests/difftests/tests/lang/abi/scalar_pair/scalar_pair-cpu/src/main.rs
```diff
@@ -0,0 +1,72 @@
...
+    out[11] = reorder_sum;
+
+    let p: (i32, u32) = (-123, 456);
+    let p_mix = (p.0 as u32).wrapping_add(p.1);
+    let q: (f32, f32) = (0.75, -1.5);
+    let q_mix = q.0.to_bits() ^ q.1.to_bits();
+    let r_mix = 8u32 ^ 16u32 ^ 0.5f32.to_bits() ^ (-2.0f32).to_bits();
+    out[12] = p_mix;
+    out[13] = q_mix;
+    out[14] = r_mix;
```
Firestar99, line 46: > It looks like this function is computing something very specific but I don't know what it is. A brief comment would be useful.

@nnethercote in https://github.com/Rust-GPU/rust-gpu/pull/380/files/8bdec822a9382c34c96031617a3ae37189621583#r2373605856

Firestar99:
Updated this PR in https://github.com/Rust-GPU/rust-gpu/pull/640

PR #387, opened 2025-09-16
`run-wasm` helper (`cargo-run-wasm`) for testing the wgpu runner on the web.

eddyb:
*Originally submitted as:*
- https://github.com/EmbarkStudios/rust-gpu/pull/1016
*(kept below the main part of that description that is still relevant, and updated it)*

---

Had to work around some sRGB complications (and push constants) but it finally works!

* `rustup target add wasm32-unknown-unknown`
  <sub>&nbsp;&nbsp;(should we add this to `rust-toolchain.toml`?)</sub>
* `cargo run-wasm -p example-runner-wgpu`
  <sub>&nbsp;&nbsp;(defaults to serving the result of the build on `http://localhost:8000`)</sub>
* `google-chrome-stable --enable-features=Vulkan --enable-unsafe-webgpu http://localhost:8000`

image

### DIFF
--- a/examples/runners/wgpu/src/graphics.rs
+++ b/examples/runners/wgpu/src/graphics.rs
@@ -88,8 +88,10 @@ async fn run(
     window: Window,
     compiled_shader_modules: CompiledShaderModules,
 ) {
-    let backends =
-        wgpu::Backends::from_env().unwrap_or(wgpu::Backends::VULKAN | wgpu::Backends::METAL);
+    // FIXME(eddyb) should this just use `wgpu::Backends::PRIMARY`?
+    // (that also enables the DirectX 12 backend, not sure we want that one?)
+    let backends = wgpu::Backends::from_env()
+        .unwrap_or(wgpu::Backends::VULKAN | wgpu::Backends::METAL | wgpu::Backends::BROWSER_WEBGPU);
     let instance = wgpu::Instance::new(&wgpu::InstanceDescriptor {
         backends,
         ..Default::default()
--- a/examples/runners/wgpu/src/lib.rs
+++ b/examples/runners/wgpu/src/lib.rs
@@ -216,7 +216,8 @@ fn maybe_watch(
 #[command()]
 pub struct Options {
     /// which shader to run
-    #[arg(short, long, default_value = "sky")]
+    #[cfg_attr(not(target_arch = "wasm32"), arg(short, long, default_value = "sky"))]
+    #[cfg_attr(target_arch = "wasm32", arg(short, long, default_value = "mouse"))]
     shader: RustGPUShader,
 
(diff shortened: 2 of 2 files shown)


eddyb:
I just wanted to rebase and share this branch, but I spent a few minutes on it and I think I fixed all remaining issues (the screenshot in the description is from the current state of the branch).

Ironically(?), the sky shader doesn't work because of it demoing pipeline-overridable constants (Vulkan specialization constants), but I am happy to default to the mouse shader because of the interactivity.

examples/runners/wgpu/src/graphics.rs
```diff
@@ -367,8 +480,39 @@ fn create_pipeline(
     device: &wgpu::Device,
     pipeline_layout: &wgpu::PipelineLayout,
     surface_format: wgpu::TextureFormat,
-    compiled_shader_modules: CompiledShaderModules,
+    mut compiled_shader_modules: CompiledShaderModules,
 ) -> wgpu::RenderPipeline {
+    if options.emulate_push_constants_with_storage_buffer {
```
LegNeato, line 485: This feels like it should be somewhere reusable?

LegNeato (approved):
Loved this demo when I saw it.

examples/runners/wgpu/src/graphics.rs
eddyb: It's been a while since I whipped up that transformation, because WebGPU launched w/o push constants:
- https://github.com/gpuweb/gpuweb/issues/75

But given that I'm doing this on the SPIR-V module, before Naga sees it, it should maybe be:
- part of `wgpu`/Naga, which would let other `wgpu` users reuse this emulation
  - or at least implemented here on Naga IR, then passing WGSL/`naga::Module` to `wgpu`
- done as part of the build script, and potentially implemented in `rustc_codegen_spirv`/SPIR-T
  - @Firestar99's ideas around a WGSL-outputting target would also be relevant here
  - for wasm, it'd likely be best to disable runtime Naga, and embed the WGSL directly

eddyb:
> * `rustup target add wasm32-unknown-unknown`
>   <sub>&nbsp;&nbsp;(should we add this to `rust-toolchain.toml`?)</sub>

Oops, forgot about this.
I'm also thinking that if we build this in CI, and use a crate like [`we_clap`](http://crates.io/crates/we_clap) (which wraps `clap` to let `?foo=bar` work for web targets, like `--foo=bar` does for CLI), we could have this on the website, w/ all the examples.
<sub>(might even be more impactful for shadertoys, tho I'm not sure the current approach there is a good idea)</sub>

**EDIT**: oh and `wgpu` has a WebGL2 fallback we could enable, allowing simpler web demos to "always work".

PR #392, opened 2025-09-17
asm: deny `OpTypeVector`, always infer type from asm params

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/391

Upcoming PR https://github.com/Rust-GPU/rust-gpu/pull/380 allows `OpTypeVector` to have a "dynamic" size and alignment, to properly differentiate between `Vec3` and `Vec3A`. But that implies that we can't trivially construct an `OpTypeVector`, since we don't know the sizes or alignments it should have. In preparation, this PR denies all usages of `OpTypeVector` in `asm!` blocks and changes existing ones to infer the type with `typeof{param}` or `typeof*{param}`, the latter being a pointer to the type that can be `OpStore`'ed.


crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -717,13 +723,6 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
                     (TyPat::Void, SpirvType::Void) => Ok([None]),
                     (TyPat::Pointer(_, pat), SpirvType::Pointer { pointee: ty, .. })
                     | (TyPat::Vector(pat), SpirvType::Vector { element: ty, .. })
-                    | (
-                        TyPat::Vector4(pat),
-                        SpirvType::Vector {
-                            element: ty,
-                            count: 4,
-                        },
-                    )
```
eddyb: This one might be fine to keep? Since it expects the vector to already exist, it's not synthesizing a new one.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -367,14 +367,20 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
-            Op::TypeVector => SpirvType::Vector {
-                element: inst.operands[0].unwrap_id_ref(),
-                count: inst.operands[1].unwrap_literal_bit32(),
+            Op::TypeVector => {
+                self.struct_err(
+                    "`OpTypeVector` in asm! is not supported, since `#[spirv(vector)]` structs can \
+                    create different vector types due to varying size and alignment",
+                )
+                .with_note(
+                    "pass `glam::VecN` as parameters and use `typeof{foo}` or `typeof*{foo}` (pointer to type) to resolve vector type",
```
eddyb, line 379: In theory we could extract the operands and suggest the exact correct type, but that feels overkill.

eddyb (changes requested):

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
Firestar99: In theory yes, we could. But without the change below constructing a `SpirvType::Vector`, automatic type resolution doesn't work anyway. Just tried it by undoing the `OpSample*` explicit typing I added.

Will readd it anyway, can't hurt :D

eddyb (approved):

irisfalke:
how are you meant to use the typeof{}? you cant pass in a parameter holding a glam vec

Firestar99:
Try `typeof*{}` [like in this snippet](https://github.com/Rust-GPU/rust-gpu/blob/main/tests/compiletests/ui/entry/asm/entry1_global_asm.rs#L27-L35)? There I'm passing in `&Vec4` and `typeof*` deref's the `*` to get me the `Vec4` type.

Issue #410, opened 2025-09-24
Allow deriving `VectorOrScalar`


nazar-pc:
The trait is unsafe and currently says:
> Implementing this trait on non-scalar or non-vector types may break assumptions about other unsafe code, and should not be done.

Fair enough, but not particularly useful for someone who would like to implement it.

It'd be great if there was a way to derive its implementation with a derive macro. Very helpful for new types, structs with two numeric fields, etc. Without this capability things like `subgroup_shuffle()` currently require substantial boilerplate.

Firestar99:
Luckily, https://github.com/Rust-GPU/rust-gpu/pull/380 get us 90% there already. Just need to wrap that in a ~~derive~~ attribute macro that also implements `spirv_std::vector::Vector` and `spirv_std::vector::VectorOrScalar`.

See https://github.com/Rust-GPU/rust-gpu/pull/411

nazar-pc:
BTW, would it be difficult to implement this trait on things like `[u32; 2]` or more generally `<const N: usize> [T; N] where T: Scalar`?
It'd simplify some of the code I'm writing a lot.

Firestar99:
`[u32; 2]` is an `OpTypeArray %u32 2`, not an `OpTypeVector %u32 2`, so it wouldn't work unless we change arrays to vectors for small array sizes. 

I've been thinking about maybe a trait + derive macro specifically for subgroup intrinsics that would allow decomposing arbitrary types into the plain primitives and then apply the operation for each of those. That should be relatively easy to implement, next to what https://github.com/Rust-GPU/rust-gpu/pull/411 offers. That would also solve the array case, since it's just a bunch of primitives.

Firestar99:
@nazar-pc Made you https://github.com/Rust-GPU/rust-gpu/pull/441 with a new `trait ScalarOrVectorComposite`:

https://github.com/Rust-GPU/rust-gpu/blob/ce822f18ec270f25eebfe6ce5003a563d60e5c8e/tests/compiletests/ui/arch/subgroup/subgroup_composite.rs#L10-L46

Just be careful of the breaking changes in https://github.com/Rust-GPU/rust-gpu/pull/380 and https://github.com/Rust-GPU/rust-gpu/pull/440

PR #411, opened 2025-09-24
derive spirv vector on user structs

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/380

Since https://github.com/Rust-GPU/rust-gpu/pull/380 already implements all the infrastructure to declare spirv vectors properly in glam, may as well expose it to our end users and allow them to declare their own vector types. The `spirv_vector` proc macro expands to the same `rust_gpu::vector::v1` glam uses, but additionally implements `spirv_std::vector::Vector` and `spirv_std::vector::VectorOrScalar` on the type, so it can be used in eg. subgroup intrinsics.

Example:
https://github.com/Rust-GPU/rust-gpu/blob/6be1bffa201b348ebb63d1cf30ed43bf96f8abe2/tests/compiletests/ui/glam/spirv_vector_macro.rs#L13-L33

First two commits just move some code around, so best reviewed commit by commit. 

close https://github.com/Rust-GPU/rust-gpu/issues/410

### DIFF
--- a/crates/spirv-std/macros/src/lib.rs
+++ b/crates/spirv-std/macros/src/lib.rs
@@ -80,6 +80,9 @@ use proc_macro::TokenStream;
 use proc_macro2::{Delimiter, Group, Ident, TokenTree};
 use quote::{ToTokens, TokenStreamExt, format_ident, quote};
 use spirv_std_types::spirv_attr_version::spirv_attr_with_version;
+use syn::punctuated::Punctuated;
+use syn::spanned::Spanned;
+use syn::{GenericParam, Token};
 
 /// A macro for creating SPIR-V `OpTypeImage` types. Always produces a
 /// `spirv_std::image::Image<...>` type.
@@ -310,4 +313,55 @@ pub fn debug_printfln(input: TokenStream) -> TokenStream {
 #[doc(hidden)]
 pub fn gen_sample_param_permutations(_attr: TokenStream, item: TokenStream) -> TokenStream {
     sample_param_permutations::gen_sample_param_permutations(item)
+}
+
+#[proc_macro_attribute]
+pub fn spirv_vector(attr: TokenStream, item: TokenStream) -> TokenStream {
+    spirv_vector_impl(attr.into(), item.into())
+        .unwrap_or_else(syn::Error::into_compile_error)
+        .into()
+}
+
+fn spirv_vector_impl(
+    attr: proc_macro2::TokenStream,
+    item: proc_macro2::TokenStream,
+) -> syn::Result<proc_macro2::TokenStream> {
+    // Whenever we'll properly resolve the crate symbol, replace this.
+    let spirv_std = quote!(spirv_std);
+
+    // Defer all validation to our codegen backend. Rather than erroring here, emit garbage.
+    let item = syn::parse2::<syn::ItemStruct>(item)?;
+    let ident = &item.ident;
+    let gens = &item.generics.params;
+    let gen_refs = &item
+        .generics
+        .params
+        .iter()
+        .map(|p| match p {
+            GenericParam::Lifetime(p) => p.lifetime.to_token_stream(),
+            GenericParam::Type(p) => p.ident.to_token_stream(),
+            GenericParam::Const(p) => p.ident.to_token_stream(),
+        })
+        .collect::<Punctuated<_, Token![,]>>();
+    let where_clause = &item.generics.where_clause;
+    let element = item
+        .fields
+        .iter()
+        .next()
+        .ok_or_else(|| syn::Error::new(item.span(), "Vector ZST not allowed"))?
+        .ty
+        .to_token_stream();
+    let count = item.fields.len();
+
+    Ok(quote! {
+        #[cfg_attr(target_arch = "spirv", rust_gpu::vector::v1(#attr))]
+        #item
+
+        unsafe impl<#gens> #spirv_std::ScalarOrVector for #ident<#gen_refs> #where_clause {
+            type Scalar = #element;
+            const N: core::num::NonZeroUsize = core::num::NonZeroUsize::new(#count).unwrap();
+        }
+
+        unsafe impl<#gens> #spirv_std::Vector<#element, #count> for #ident<#gen_refs> #where_clause {}
+    })
 }
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -87,8 +87,8 @@
 /// Public re-export of the `spirv-std-macros` crate.
 #[macro_use]
 pub extern crate spirv_std_macros as macros;
-pub use macros::spirv;
 pub use macros::{debug_printf, debug_printfln};
+pub use macros::{spirv, spirv_vector};
 
 pub mod arch;
 pub mod byte_addressable_buffer;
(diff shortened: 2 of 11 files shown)


Firestar99:
Still a bit TBD around the proc macro name `spirv_std::spirv_vector`, should we just call it `vector`?

nazar-pc:
Does it need to be a proc macro vs derive macro that would be more natural?

Firestar99:
It's an attribute macro as we need to expand into another attribute macro that is placed on the struct directly, that our codegen backend can pick up.

nazar-pc:
I see, what about requiring a user to specify it directly? Or maybe it is unnecessary in certain cases (like when I have a scalar new type rather than a vector)?

Firestar99:
What exactly do you mean? Manually implementing `Vector` and `VectorOrScalar` on your type, instead of using the attribute macro? I also wrote entirely new docs for vector (mostly in #380 so it isn't obvious diff):

https://github.com/Rust-GPU/rust-gpu/blob/9ed3407653f4c22b2b5455420e4d22319edd01b4/crates/spirv-std/src/vector.rs#L19-L61

nazar-pc:
I guess there is a reason it works that way right now, but that is (generally) not the only way.

First of all, I not only may want to derive an impl of a trait for vectors, but also for single-field structs, in which case I would not call it a vector.

Second, I'd expect it to look something like this for the case when multiple fields are present instead:
```rust
#[derive(VectorOrScalar)]
#[repr(C)]
pub struct New(u32, u32);
```

If `#[repr(C)]` is not appropriate, Rust has `#[repr(simd)]`, which I know very little about beyond the fact that it exists. Maybe `#[repr(vector)]` can be introduced if semantic meaning is different here.

Then macro would fail to compile if the struct or one of its fields doesn't have the necessary `#[repr()]` defined.

Firestar99:
On single-field structs: I've recently noticed that [structs with a single scalar member and `#[transparent]` ](https://github.com/Rust-GPU/rust-gpu/pull/380#discussion_r2376117336) do compile down to simply that scalar member. So you could just `impl Scalar` on such a struct and it should work, or fail validation. (I would not expect any actual UB to stem from this, validation should easily catch that)
So we could also offer a `spirv_scalar` (derive / attribute) macro that checks all the preconditions and `impl Scalar` and `VectorOrScalar` for it. But first, I would want to double check why we emit the scalars for these in the first place, in case there is some special requirement or such. 

`#[repr(simd)]`: That has bitten us before, I'd rather not touch repr SIMD again and instead use attribute macros *we define* instead. Since repr SIMD is still unstable, rustc has changed what this repr means between versions, so we had to hack our way out of it before. https://github.com/Rust-GPU/rust-gpu/pull/380 removes those hacks and replaces all usages of `#[repr(simd)]` with the new attribute macro.

attribute vs derive macro: It's mainly a technical reason it's an attribute macro. If we wanted to have the `impl Vector` as a derive macro, you'd still have to add the attribute macro for our codegen backend. So why duplicate things, when we could instead have one attribute macro for both?

nazar-pc:
> If we wanted to have the impl Vector as a derive macro, you'd still have to add the attribute macro for our codegen backend. So why duplicate things, when we could instead have one attribute macro for both?

My point is that the information for codegen should not be a proc macro either. If `#[repr(simd)]` doesn't work, make it `#[repr(vector)]` then or something like that that makes more sense. This is how idiomatically we do it in other cases in Rust.

As for the derive, it is also idiomatic to have the name of the trait and derive macro be the same and re-exported together. Having a separate proc macro looks more unusual, implicitly indicating complexity and is not as clear about what does it actually do. While derive macros look absolutely normal since everyone uses them all the time.

The derive macro could be called `VectorOrScalar` since that is what APIs require, internally it'd check that the struct either has one field with `#[repr(transparent)]` or `#[repr(C)]` (if/when supported) or it has `#[repr(vector)]` and multiple same type fields inside. It'll then implement both `Scalar`/`Vector` and `ScalarOrVector` for the type.

Firestar99:
I can see how `#[repr(vector)]` seems to make a lot of sense from the user's perspective. But from the codegen perspective, it very much does not:
* Adding support for a custom repr is not just really hard, you'd have to change some of the [fundamental structs deep within rustc](https://github.com/rust-lang/rust/blob/94ecb52bbeeccc990545d4ebc9e9c35dcaf285e3/compiler/rustc_abi/src/lib.rs#L79-L156). 
* But also really error prone, as we have seen before [#380](https://github.com/Rust-GPU/rust-gpu/pull/380) with glam vector having different representations on CPU and GPU. I *really* don't want to mod in a custom representation and have to deal with this, if a simpler alternative like an attribute macro is available.
* It also doesn't map to SPIRV at all, since the decision of emitting an `OpTypeStruct` or an `OpTypeVector` is completely orthogonal to the abi layout of a type. This may be confusing at first from a CPU perspective, but SPIRV doesn't actually require structs or vectors to have an abi layout, which defines the size, alignment and member offsets. Since GPUs are largely a register-based architecture, most of them simply don't support pointers pointing to registers, so having an abi layout for register-only types that can't be pointed to doesn't make much sense to begin with. It *only* requires an abi layout if said struct is read from or written out to a buffer, since other shaders or the CPU must agree on a layout there. The only requirement for the abi layout is that it should be sane (eg. no overlapping members, no member oob wrt. the struct size), otherwise, it does not care for how you layout your types.

So types really should be `#[repr(C)]` both on the CPU and GPU side, to denote their abi layout. And since the decision of whether a struct is a standard `OpTypeStruct` or wants to opt into `OpTypeVector` is completely orthogonal to it's abi layout, I think it's best to just have it be a separate optional attribute macro. And it can't be a derive macro, since our codegen backend needs it to be an attribute macro on the struct itself.

nazar-pc:
Thanks for explaining!

The reason I imagined it as derive macro is because APIs require `ScalarOrVector` implementation and it is counter-intuitive/non-idiomatic when implementation of it is done using something other than derive macro. But I also understand that there are other limitations that might dictate the design here.

I'm now wondering if there is an API to do something like parsing the file with `syn` and replacing `#[repr(vector)]` with `#[repr(C)] #[spirv_std::spirv_vector]` for example, so that it looks more natural to the user and is easier to support for the codegen backend at the same time. I have no idea if such API exists, but it'd be neat if it did.

As for derive macro, I still don't fully get why it can't be a thing. Derive macro should literally just implement a trait properly. If there are extra requirements for that trait to be derivable (like `#[repr]` attribute or `#[spirv_std::spirv_vector]` attribute) then it should be checked by the macro and generate compilation error if something doesn't match expectations, but it is still orthogonal to trait implementation as such. In fact it is possible that no extra attributes are needed for non-vector new types like `struct S(u32)`.

Firestar99:
I've been thinking about this PR a bit more while waiting for https://github.com/Rust-GPU/rust-gpu/pull/380 to be reviewed, and I have to say I may prefer the derive macro alternative. Specifically, I like that we can `#[derive(Scalar)]`, which can check that the struct is `#[repr(transparant)]` and `unsafe impl Scalar for T where Primitive: Scalar`. So why not mirror this for vectors as well, and have a derive macro that checks for `#[spirv::vector::v1]` and `impl Vector for T`.

I think this can coexist with the "subgroup trait + macro" I mentioned at https://github.com/Rust-GPU/rust-gpu/issues/410#issuecomment-3398386772, since there are subgroup APIs that specifically require a float or int Scalar or Vector. Whereas the other would be better for shuffles, boardcasts and the like.

Firestar99:
This PR needs a rethink on how it relates to https://github.com/Rust-GPU/rust-gpu/pull/441. 
* Does 441 supersede this PR?
* Or if it could be an extension of it, to allow using vectors in places where `ScalarOrVectorComposite` is unsuited for?

Firestar99:
I'll close for now until someone comes around that really wants custom vectors for whatever reason

PR #413, opened 2025-09-25
Return `JoinHandle` of watching thread to the caller

tuguzT:
This would be useful for Rust-GPU/cargo-gpu#112 to state via the Rust's type system that watch thread should never return by joining it.


LegNeato (approved):
🍻

PR #415, opened 2025-09-28
`Debug` for `CargoCmd`

tuguzT:
Fix `CargoCmd` to implement `Debug` instead of `Display` (which was presumably a typo).
Also implement `From<CargoCmd>` for `Command` which will be useful a bit later (as a part of Rust-GPU/cargo-gpu#112).


Firestar99:
thanks, that one is on me

Issue #407, opened 2025-09-23
CI: swiftshader rate limit exceeded


Firestar99:
* Likely an issue since https://github.com/Rust-GPU/rust-gpu/pull/400

```
🚀 Installing SwiftShader library...
Error: HttpClientError: API rate limit exceeded for 52.159.244.80. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```
https://github.com/Rust-GPU/rust-gpu/actions/runs/17940823538/job/51016831651?pr=334

@eddyb is there a cache option?

As this only happened once, feel free to ignore for now. But we should keep track whether this happens more often.

Firestar99:
Another one:
https://github.com/Rust-GPU/rust-gpu/actions/runs/18174678854/job/51737608900?pr=422

PR #400, opened 2025-09-18
ci: switch to `jakoch/install-vulkan-sdk-action@v1` for Vulkan SDK + SwiftShader.

eddyb:
Updating the Vulkan SDK we used in CI came up in:
- https://github.com/Rust-GPU/rust-gpu/pull/379#discussion_r2335335011

Sadly, the [`NcStudios/VulkanCI@v1.2`](https://github.com/NcStudios/VulkanCI/tree/v1.2) action we were using to get SwiftShader on Windows, is stuck on an older version, and doesn't appear to be prompt to update in general.

So this PR replaces [`humbletim/install-vulkan-sdk@v1.2`](https://github.com/humbletim/install-vulkan-sdk/tree/v1.2) + [`NcStudios/VulkanCI@v1.2`](https://github.com/NcStudios/VulkanCI/tree/v1.2) with [`jakoch/install-vulkan-sdk-action@v1`](https://github.com/jakoch/install-vulkan-sdk-action/tree/v1), which supports SwiftShader (or Lavapipe, but I kept it on SwiftShader) for Windows (AFAICT that action downloads SwiftShader/Lavapipe releases from [`jakoch/rasterizers`](https://github.com/jakoch/rasterizers), separately from the Vulkan SDK).

From what I could tell, the [`jakoch/install-vulkan-sdk-action`](https://github.com/jakoch/install-vulkan-sdk-action) repo has a lot more recent activity than the others we were using, and its approach feels more robust long-term (at least in terms of bumping Vulkan SDK versions).

PR #422, opened 2025-10-01
Redesign `spirv-builder`'s watch API

tuguzT:
It was suggested to rewrite it from scratch instead of patching existing one.

My idea now is to expose refactored `Watcher`, which could just return `CompileResult` on `recv()`, then we don't need to accept any callbacks, differentiate first compile result from others, etc.


tuguzT:
I thought CI issues were already resolved at GitHub side?
Besides, it's ready to be reviewed.

Firestar99:
CI failing is just https://github.com/Rust-GPU/rust-gpu/issues/407, recently introduced in https://github.com/Rust-GPU/rust-gpu/pull/400

Firestar99 (commented):
I love the new design!

Just iterated some more on it, so I'll let others do the reviewing

My changes to it:
* we can just consume SpirvBuilder, if users need to clone it they can
* changing a file during recompilation no longer throws errors in stdout but will trigger exactly one recompile when you call `recv()` again
* wgpu example had a race condition between the first and watched compiles

Firestar99:
I added a non-blocking `try_recv()` function as well, required me to refactor some of the internal state machine

crates/spirv-builder/src/watch.rs
```diff
@@ -1,153 +1,179 @@
...
+        let path_to_crate = builder
+            .path_to_crate
+            .as_ref()
+            .ok_or(SpirvBuilderError::MissingCratePath)?
+            .clone();
+        if !matches!(builder.print_metadata, crate::MetadataPrintout::None) {
+            return Err(SpirvWatcherError::WatchWithPrintMetadata.into());
+        }
+
+        let (tx, rx) = sync_channel(1);
```
tuguzT, line 62: I've noticed that `cargo-gpu` produces duplicate last messages (finished compiling or failure) on branch Rust-GPU/cargo-gpu#117 on Windows.
Previously provided value was `0`, now it is `1`. Could this be the reason?

Firestar99: Yes it is. When a file changes while a recompilation was running, your version would spit out an error into stderr due to the channel being full. This version uses `()` to record that something has happened and discards all further messages that fail to send due to the channel being full. Which allows edits during recompilation to correctly retrigger a followup recompilation. But it seems like most fs changes emit more than one event, so (almost*) all recompilations will trigger a second time. But I'd rather recompile again to figure out nothing has changed, than having a stale state due to dropping events during ongoing compilation.

(*it is a race condition between the notif thread submitting all events and the polling thread polling an event)

crates/spirv-builder/src/watch.rs
```diff
@@ -1,153 +1,179 @@
...
+    First,
+    /// the first compile (and all consecutive ones) failed:
+    /// * only recompile when watcher notifies us
+    /// * the whole project dir is being watched, remove that watch
+    /// * success: go to [`Self::Watching`]
+    /// * fail: stay in [`Self::FirstFailed`]
+    FirstFailed,
+    /// at least one compile finished and has set up the proper file watches:
+    /// * only recompile when watcher notifies us
+    /// * always stays in [`Self::Watching`]
```
schell, line 36: Nit: These comments should be capitalized.

PR #429, opened 2025-10-08
move codegen for panic entry point to a separate mod

Firestar99:
@eddyb mind if we move that code to declutter `call`?


crates/rustc_codegen_spirv/src/builder/panic_entry_point.rs
```diff
@@ -0,0 +1,782 @@
...
+            /// a much jankier fallback still has to be used, as it it were:
+            ///
+            /// `format!("a{{0}}b{{1}}c\n  with {{…}} from: {}, {}", x, y)`
+            /// (w/ `const_pieces = ["a", "b", "c"]` & `ref_args = [&x, &y]`).
+            has_unknown_fmt_placeholder_to_args_mapping: Option<usize>,
+        }
+        struct FormatArgsNotRecognized(String);
+
+        // HACK(eddyb) this is basically a `try` block.
+        let mut try_decode_and_remove_format_args = || {
```
eddyb: I think it would make more sense to make these types (and the closure becoming a method) the real interface here, and name the module `format_args_decompiler` (because that's what this cursedness _really_ is).

In theory we could allow non-panic uses of it, like instead of `debug_printf!` we could have `debug_print!` (or even the Rust `std` `dbg!` macro), that uses `format_args!` combined with a function that we treat just like we do panic entry-points.

Firestar99: done, feel free to rereview. Not a huge fan how I handled the `Result`, but wrapping it in a new type makes the function super ugly.

Firestar99:
Added a few more things:
* move printf (and sample_param_permutation) macro implementation to their own mods
* move `debug_printf*` functions in `spirv-std`'s `lib.rs` to a new `mod debug_printf`
  * Every time `lib.rs` is edited, that one debug printf compiletest needs a bless, but with this change it won't.
  * Resolves conflicts with https://github.com/Rust-GPU/rust-gpu/pull/440

Firestar99:
Wouldn't be surprised if the merge failed on a conflict in `tests/compiletests/ui/arch/debug_printf_type_checking.stderr`, exactly what this PR tries to rectify (see previous comment)

Issue #428, opened 2025-10-08
ByteAddressableBuffer: storing a Scalar and ScalarPair within the same codebase fails to compile


Firestar99:
### Compiletest

The following compiletest fails to compile:

```rust
// build-pass

use spirv_std::ByteAddressableBuffer;
use spirv_std::spirv;

#[derive(Copy, Clone, Debug)]
pub struct MyScalar(i32);

#[spirv(fragment)]
pub fn store_scalar(
    #[spirv(descriptor_set = 0, binding = 0, storage_buffer)] buf: &mut [u32],
    #[spirv(flat)] scalar: MyScalar,
) {
    unsafe {
        let mut buf = ByteAddressableBuffer::from_mut_slice(buf);
        buf.store(5, scalar);
    }
}

#[derive(Copy, Clone, Debug)]
pub struct MyScalarPair(i32, i32);

#[spirv(fragment)]
pub fn store_scalar_pair(
    #[spirv(descriptor_set = 0, binding = 0, storage_buffer)] buf: &mut [u32],
    #[spirv(flat)] pair0: i32,
    #[spirv(flat)] pair1: i32,
) {
    unsafe {
        let mut buf = ByteAddressableBuffer::from_mut_slice(buf);
        buf.store(5, MyScalarPair(pair0, pair1));
    }
}
```

### Error
```rust
error: buffer_store_intrinsic should have 4 args, it has 5
   --> /home/firestar99/workspace/rust-gpu/crates/spirv-std/src/byte_addressable_buffer.rs:155:13
    |
155 |             buffer_store_intrinsic(self.data, byte_index, value);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to 1 previous error
```

### Notes
* only stores are affected, loads work fine since they don't have any special handling for scalar pairs
* stores expect 5 args when passing a scalar pair, 4 when passing a scalar
* it seems like the presence of a scalar pair store converts the scalar store to a scalar pair one?

## System Info
 - rust-gpu main dbf373762d29cdcde621ed40de7d53f36a5d783f

PR #430, opened 2025-10-08
ByteAddressableBuffer: fix mixed scalar and scalar pair stores

Firestar99:
closes https://github.com/Rust-GPU/rust-gpu/issues/428

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -3247,7 +3247,7 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         &mut self,
         callee_ty: Self::Type,
         _fn_attrs: Option<&CodegenFnAttrs>,
-        _fn_abi: Option<&FnAbi<'tcx, Ty<'tcx>>>,
+        fn_abi: Option<&FnAbi<'tcx, Ty<'tcx>>>,
         callee: Self::Value,
         args: &[Self::Value],
         funclet: Option<&Self::Funclet>,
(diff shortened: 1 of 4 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -3247,7 +3247,7 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         &mut self,
         callee_ty: Self::Type,
         _fn_attrs: Option<&CodegenFnAttrs>,
-        _fn_abi: Option<&FnAbi<'tcx, Ty<'tcx>>>,
+        fn_abi: Option<&FnAbi<'tcx, Ty<'tcx>>>,
```
Firestar99, line 3250: I hope just using the one given to me is correct here, previously the `FnAbi` was computed from:
https://github.com/Rust-GPU/rust-gpu/blob/e38a78db6d79c1834fb0692c25b2126caa3ff171/crates/rustc_codegen_spirv/src/codegen_cx/declare.rs#L73
But it was in a function that wasn't given an `FnAbi` but a `rustc_middle::ty::Instance`, which I never interacted with.

crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
```diff
@@ -149,16 +149,10 @@ impl<'tcx> CodegenCx<'tcx> {
...
-            self.buffer_load_intrinsics
-                .borrow_mut()
-                .insert(def_id, mode);
+            self.buffer_load_intrinsics.borrow_mut().insert(def_id);
         }
         if attrs.buffer_store_intrinsic.is_some() {
-            let mode = &fn_abi.args.last().unwrap().mode;
-            self.buffer_store_intrinsics
-                .borrow_mut()
-                .insert(def_id, mode);
```
eddyb, line 161: Oh wow, so being keyed on the `DefId` of a _generic function_ meant that the last instance would win, and every call would be expected to conform to it.

Issue #424, opened 2025-10-05
Wrapping a `bitvec::slice::BitSlice` in a `struct` causes an ICE (attempt to divide by zero)


Steveplays28:
```toml
# Cargo.toml
spirv-std = { git = "https://github.com/Rust-GPU/rust-gpu.git", rev = "f79c4181a5dc2d37303947b113f190930c6c1ce6" }
bitvec = { git = "https://github.com/uglyoldbob/bitvec.git", default-features = false }
```

```rust
pub struct Wrapper<'v> {
    pub bit_slice: &'v mut BitSlice<u8>,
}
```

[rust-gpu-bitslice-rustc-ice-2025-10-03.txt](https://github.com/user-attachments/files/22710026/rust-gpu-bitslice-rustc-ice-2025-10-03.txt)

LegNeato:
https://github.com/Rust-GPU/rust-gpu/blob/dbf373762d29cdcde621ed40de7d53f36a5d783f/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L770

LegNeato:
This is what `BitSlice` looks like:

```rust
pub struct BitSlice<T = usize, O = Lsb0>
where
	T: BitStore,
	O: BitOrder,
{
	/// The ordering of bits within a `T` register.
	_ord: PhantomData<O>,
	/// The register type used for storage.
	_typ: PhantomData<[T]>,
	/// Indicate that this is a newtype wrapper over a wholly-untyped slice.
	///
	/// This is necessary in order for the Rust compiler to remove restrictions
	/// on the possible values of reference handles to this type. Any other
	/// slice type here (such as `[u8]` or `[T]`) would require that `&/mut
	/// BitSlice` handles have values that correctly describe the region, and
	/// the encoding *does not* do this. As such, reference handles to
	/// `BitSlice` must not be even implicitly dereferenceäble to real memory,
	/// and the slice must be a ZST.
	///
	/// References to a ZST have no restrictions about what the values can be,
	/// as they are never able to dereference real memory and thus both
	/// addresses and lengths are meaningless to the memory inspector.
	///
	/// See `ptr::span` for more information on the encoding scheme used in
	/// references to `BitSlice`.
	_mem: [()],
}
```

PR #431, opened 2025-10-09
Fix divide by zero ICE

LegNeato:
We assumed every array/vector/matrix element has a non-zero stride, but `[()]` inside `BitSlice` produced a zero stride which then divides by zero.

Fixes https://github.com/Rust-GPU/rust-gpu/issues/424.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -767,8 +767,18 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                     let stride = ty_kind.sizeof(self)?;
                     ty_size = MaybeSized::Sized(stride);
 
-                    indices.push((offset.bytes() / stride.bytes()).try_into().ok()?);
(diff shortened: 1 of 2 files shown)


Steveplays28:
Thank you for the fix!

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -767,8 +767,18 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
 
-                    indices.push((offset.bytes() / stride.bytes()).try_into().ok()?);
-                    offset = Size::from_bytes(offset.bytes() % stride.bytes());
+                    if stride == Size::ZERO {
+                        if offset != Size::ZERO {
+                            trace!("zero-sized element with non-zero offset: {:?}", offset);
+                            return None;
+                        }
+
+                        indices.push(0);
```
eddyb, line 776: Heh, got a conflict locally when rebasing on top of this PR, turns out I had a commit that included the `indices.push(0);` (but not the check - so I resolved the conflict by keeping the `main` version, i.e. the one from this PR), so that's probably an extra point in the favor of this being correct.

Issue #263, opened 2025-05-27
Create and publish docker images


LegNeato:
It would really help people get started if we had docker images with the appropriate setup (vulkan tools, correct rust version, etc) already ready to go to. lower the bar for contributors and users. We can look at `rust-cuda`'s to get an idea.

SWASTIC-7:
I worked on the issue and here’s what I’ve found:

- I was able to create a proper Docker image and successfully compile Rust-GPU within it.

- However, running WGPU shaders on the host presents several challenges: while compute shaders and CPU-based shaders work fine, GPU shaders fail due to Wayland display limitations — running them inside Docker isn’t feasible because of this.

- Additionally, while running binaries from host, there is a potential for glibc compatibility issues on the host, as shaders are compiled against the specific Ubuntu version used in the Docker image.

- The repeated error `Shader compiling failed: could not notify for changes: No path was found` arises from the file watcher mechanism and is related to path resolution, not the shader compilation itself.

Given these constraints, a better approach would be to use Nix,
I can open a draft pr of my work or else would love to switch to Nix
@LegNeato

Firestar99:
Oh this is a bit of an older issue. To be honest, I'm not sure how useful docker images are gonna be. In the mean time, we've made [cargo-gpu](https://github.com/Rust-GPU/cargo-gpu) to take care of toolchain installation, building the codegen backend and building the shaders with the codegen backend. 

About nix: Take note of https://github.com/Rust-GPU/rust-gpu/pull/463 proposing to use nix flakes instead of the current `nix-shell`. And [Graphite](https://github.com/GraphiteEditor/Graphite) have a [custom nix setup to build rust-gpu](https://github.com/GraphiteEditor/Graphite/blob/master/.nix/deps/rust-gpu.nix), which we may bring to cargo-gpu at some point. Adding an env var that overwrites codegen backend building with "just use that path and don't worry about it" would probably be sufficient for nix support.

SWASTIC-7:
[...]

PR #434, opened 2025-10-11
Dockerfile init

SenorGato:
In response to Issue #263 

This Dockerfile builds a working development environment for rust-gpu, and also builds the bin outputs.  
Instructions on use are included in the wiki

### DIFF
--- a/docs/src/building-rust-gpu.md
+++ b/docs/src/building-rust-gpu.md
@@ -39,6 +39,24 @@ cargo run \
 
 You should see `warning: use-installed-tools feature on, skipping compilation of C++ code` during the compilation, but otherwise the build will function just the same as if you compiled the C++ code, with the exception that it will fail if you don't have SPIRV-Tools installed correctly.
 
+## Docker Image
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM debian:bookworm-slim AS build
+
+RUN apt-get update && apt-get install -y \
+    curl ca-certificates build-essential pkg-config \
+    libssl-dev clang \
+    && rm -rf /var/lib/apt/lists/*
```
nazar-pc, line 6: There is nothing Vulkan-specific here though, just a typical boilerplate

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Install rustup
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+WORKDIR /app
+COPY . .
```
nazar-pc, line 14: Why would you copy the whole repo and include it into the container image?

PR #435, opened 2025-10-11
Nix flake init

SenorGato:
In addition to #434 
I also wrote a Nix Flake for my local system.  This will build the derivation and work as a flake input, as well as building a local devshell.  
Its fairly trivial to modify this flake to package rust-gpu upstream in nixpkgs if that is something you wanted to do.  

I left the existing .nix shells where they were although ideally they would be consolidated into the flake.  

Also added documentation explaining usage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ tests/Cargo.lock
 .github/install-spirv-tools/Cargo.lock
 rustc-ice-*.txt
 .idea
+result/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ tests/Cargo.lock
 .github/install-spirv-tools/Cargo.lock
 rustc-ice-*.txt
 .idea
+result/
```
nazar-pc, line 8: ?

PR #440, opened 2025-10-14
Scalar and Vector types refactor

Firestar99:
## Breaking changes
* rename `VectorOrScalar` to `ScalarOrVector` to match spec expression used in subgroup intrinsics
* changed publicly exported paths
  * `spirv_std::vector::VectorOrScalar` -> `spirv_std::ScalarOrVector`
  * `spirv_std::scalar::Scalar` -> `spirv_std::Scalar`
  * `spirv_std::vector::Vector` -> `spirv_std::Vector`
  * `spirv_std::number::Number` -> `spirv_std::Number`
  * `spirv_std::integer::Integer` -> `spirv_std::Integer`
  * `spirv_std::integer::SignedInteger` -> `spirv_std::SignedInteger`
  * `spirv_std::integer::UnsignedInteger` -> `spirv_std::UnsignedInteger`
  * `spirv_std::float::Float` -> `spirv_std::Float`
* these are probably barely used outside of `spirv-std` directly

## Changes
* move Scalar traits (`Number`, `Integer`, `SignedInteger`, `UnsignedInteger`, `Float`) to `mod scalar`
* simplify proc macro to declare Scalars and Vectors
* move `VectorOrScalar` and `Sealed` to new `mod vector_or_scalar`
* remove root-level mods `integer`, `number`, `sealed`
  * keep `float` since it has intrinsics

PR #441, opened 2025-10-14
add trait `ScalarComposite`

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/440

https://github.com/Rust-GPU/rust-gpu/blob/6db18940cfd473e181b0838cd8aaf5e3c69c0799/crates/spirv-std/src/scalar_or_vector.rs#L22-L49

Also see previous iteration at https://github.com/Rust-GPU/rust-gpu/pull/411

### todo
* [x] derive enums
* [x] `subgroup_all_equal()`
* [x] `subgroup_broadcast_const` added in https://github.com/Rust-GPU/rust-gpu/pull/443

close https://github.com/Rust-GPU/rust-gpu/issues/410

### DIFF
--- a/crates/spirv-std/src/scalar_or_vector.rs
+++ b/crates/spirv-std/src/scalar_or_vector.rs
@@ -19,32 +19,31 @@ pub unsafe trait ScalarOrVector: ScalarComposite + Default {
     const N: NonZeroUsize;
 }
 
-/// A `VectorOrScalarComposite` is a type that is either
+/// A `ScalarComposite` is a type that is either
 /// * a [`Scalar`]
-/// * a [`Vector`]
-/// * an array of `VectorOrScalarComposite`
-/// * a struct where all members are `VectorOrScalarComposite`
+/// * a [`Vector`] (since vectors are made from scalars)
+/// * an array of `ScalarComposite`
+/// * a struct where all members are `ScalarComposite`
 /// * an enum with a `repr` that is a [`Scalar`]
 ///
 /// By calling [`Self::transform`] you can visit all the individual [`Scalar`] and [`Vector`] values this composite is
 /// build out of and transform them into some other value. This is particularly useful for subgroup intrinsics sending
 /// data to other threads.
 ///
-/// To derive `#[derive(VectorOrScalarComposite)]` on a struct, all members must also implement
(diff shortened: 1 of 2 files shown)


Firestar99:
> assuming this is zero cost abstraction

Pretty much, otherwise you'd see the subgroup compiletests with disassembly fail due to different codegen.

> it can implicitly result in multiple subgroup operations

If you can't observe the difference between a single operation and multiple operations back-to-back, does that actually matter? 
Also, you can expect all modern hardware to decompose any kind of vector op into multiple scalar ops, so us emitting actual vector ops won't give us any real benefit. Apart from maybe saving a few bytes of shader IR.

LegNeato:
Going to let @eddyb review

Firestar99:
I'm surprised you say that, since this is only changing some ordinary rust code in `spirv-std`, defining that new trait and providing a derive for it. I didn't change any of the spirv intrinsics themselves.

nazar-pc:
I'd love for this to land, it'll make a few places in my code much less awkward

crates/spirv-std/src/scalar_or_vector.rs
```diff
@@ -11,12 +11,76 @@ pub(crate) mod sealed {
...
+/// crate. [`FromPrimitive`] requires that either the enum is exhaustive, or you provide it with a variant to default
+/// to, by either implementing [`Default`] or marking a variant with `#[num_enum(default)]`. Note to disable default
+/// features on the [`num_enum`] crate, or it won't compile on SPIR-V.
+///
+/// [`Integer`]: crate::Integer
+/// [subgroup operations]: crate::arch::subgroup_shuffle
+/// [`FromPrimitive`]: https://docs.rs/num_enum/latest/num_enum/derive.FromPrimitive.html
+/// [`IntoPrimitive`]: https://docs.rs/num_enum/latest/num_enum/derive.IntoPrimitive.html
+/// [`num_enum`]: https://crates.io/crates/num_enum
+pub trait ScalarOrVectorComposite: Copy + Send + Sync + 'static {
```
Firestar99: Thought: shorten the name to `ScalarComposite`. Vectors are made from scalars, so it is still correctly named.

PR #443, opened 2025-10-17
subgroup: fix docs and broadcast const variant

Firestar99:
* fix broadcast docs as reported by @nazar-pc in https://github.com/Rust-GPU/rust-gpu/pull/306/files/88eb6e3b2cfb30188aa9164d529738eddeaa77dd#r2413686865
* adds `subgroup_broadcast_const` variant
  * prior to `spv1.5` or `vulkan1.2`, the `OpGroupNonUniformBroadcast`'s `id` param must be a constant
  * you can't actually use our `subgroup_broadcast`, even if the `id` is a constant as we'll emit it as a dynamic param instead
  * `subgroup_broadcast_const` accepts `id` as a const generic and ensures it's emitted as a constant
* Question: In the newer spirv specs they refactored the docs to respect the new "tangle" definitions, should I also update the docs of our intrinsics?

### DIFF
--- a/crates/spirv-std/src/arch/subgroup.rs
+++ b/crates/spirv-std/src/arch/subgroup.rs
@@ -281,8 +281,9 @@ pub fn subgroup_all_equal<T: ScalarOrVector>(value: T) -> bool {
 ///
 /// # Safety
 /// * `id` must be dynamically uniform
-/// * before 1.5: `id` must be constant
 /// * Result is undefined if `id` is an inactive invocation or out of bounds
+/// * This variant with a dynamic `id` requires at least `spv1.5` or `vulkan1.2`. Alternatively, you can use
+/// [`subgroup_broadcast_const`] with a constant `id`.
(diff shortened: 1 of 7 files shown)


nazar-pc:
> Question: In the newer spirv specs they refactored the docs to respect the new "tangle" definitions, should I also update the docs of our intrinsics?

I'd prefer that, but when I saw it the first time, I had to search what they mean by "tangle" before I could understand what they're saying :sweat_smile: So make sure it is discoverable as well or else it becomes even less clear for newbies.

Firestar99:
The complexity of "tangle" (despite not really adding anything substantial over "subgroup size") is my main concern as well. I mean I could sit down and write some nice module level docs on subgroups, tangle and the different operations. But I'd honestly rather spent that time documenting the more fundamental aspects of rust-gpu first.

Firestar99:
Conflicts with https://github.com/Rust-GPU/rust-gpu/pull/440, so just waiting for that one to merge

PR #451, opened 2025-10-24
Remove unnecessary closures

c410-f3r:
Just wanted to provide some contributions to the project.

Looking at the issues alone didn't provide enough context so I shamelessly went ahead with `redundant_closure_for_method_calls`, which is trivial to review and accept :)

Feel free to indicate more things that are feasible/solvable for someone new.

Cheers

### DIFF
--- a/crates/rustc_codegen_spirv/build.rs
+++ b/crates/rustc_codegen_spirv/build.rs
@@ -8,6 +8,7 @@ use std::collections::VecDeque;
 use std::error::Error;
 use std::path::{Path, PathBuf};
 use std::process::{Command, ExitCode};
+use std::string::ToString;
 use std::{env, fs, mem};
 
 /// Current `rust-toolchain.toml` file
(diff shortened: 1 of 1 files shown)


crates/rustc_codegen_spirv/build.rs
```diff
@@ -33,15 +34,15 @@ fn get_rustc_commit_hash() -> Result<String, Box<dyn Error>> {
     rustc_output("-vV")?
         .lines()
         .find_map(|l| l.strip_prefix("commit-hash: "))
-        .map(|s| s.to_string())
+        .map(ToString::to_string)
```
eddyb, line 37: Apropos of nothing, Lean4 has a cute syntax for this I wish more languages had:
```lean
.map(·.to_string())
```
<sub>(based on [a quick GitHub search](https://github.com/search?q=repo%3Aleanprover%2Flean4+%2F%C2%B7%5C.toString%2F&type=code), they literally have `(·.toString)` for their equivalent of the Rust `|x| x.to_string()`)</sub>

There's a few others, I think Perl is one, but it's not as common as it could be.

Issue #452, opened 2025-10-27
ICE when using function as an argument


nazar-pc:
I need to reuse a sorting function to sort by different keys depending on use case, so I made the following changes to my project:
```diff
@@ -17,6 +17,7 @@
     bit_position: u32,
     block_size: usize,
     local_bucket: &mut [PositionR; ELEMENTS_PER_THREAD],
+    cmp_key: fn(&PositionR) -> u32,
 ) {
     // For local swaps within a thread's register data, iterate over half the elements to form pairs
     // `(a_offset, b_offset)` where indices differ only at `bit_position` and swaps them in-place
@@ -40,7 +41,7 @@
         let select_smaller =
             ((lane_id as usize * ELEMENTS_PER_THREAD + a_offset) & block_size) == 0;
 
-        let (final_a, final_b) = if (a.position <= b.position) == select_smaller {
+        let (final_a, final_b) = if (cmp_key(&a) <= cmp_key(&b)) == select_smaller {
             (a, b)
         } else {
             (b, a)
@@ -57,6 +58,7 @@
     bit_position: u32,
     block_size: usize,
     local_bucket: &mut [PositionR; ELEMENTS_PER_THREAD],
+    cmp_key: fn(&PositionR) -> u32,
 ) {
     // For cross-subgroup swaps, compute partner lane differing at `lane_bit_position`
     let lane_bit_position = bit_position - ELEMENTS_PER_THREAD.ilog2();
@@ -84,7 +86,7 @@
             PositionR { position, r }
         };
 
-        let selected_value = if (a.position <= b.position) == select_smaller {
+        let selected_value = if (cmp_key(&a) <= cmp_key(&b)) == select_smaller {
             a
         } else {
             b
@@ -121,6 +123,7 @@
 fn sort_local_bucket<const ELEMENTS_PER_THREAD: usize>(
     lane_id: u32,
     local_bucket: &mut [PositionR; ELEMENTS_PER_THREAD],
+    cmp_key: fn(&PositionR) -> u32,
 ) {
     // Iterate over merger stages, doubling block_size each time
     let mut block_size = 2;
@@ -132,10 +135,22 @@
         for bit_position in (0..merger_stage).rev() {
             if bit_position < ELEMENTS_PER_THREAD.ilog2() {
                 // Local swaps within thread's registers, no synchronization needed
-                perform_local_compare_swap(lane_id, bit_position, block_size, local_bucket);
+                perform_local_compare_swap(
[...]

PR #453, opened 2025-10-27
Do not ICE on indirect fn pointers

LegNeato:
Fixes ICE in https://github.com/Rust-GPU/rust-gpu/issues/452.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
+++ b/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
@@ -237,10 +237,18 @@ pub fn check_fragment_insts(sess: &Session, module: &Module) -> super::Result<()
         let mut any_err = None;
         for inst in module.functions[index].all_inst_iter() {
             if inst.class.opcode == Op::FunctionCall {
-                let callee = func_id_to_idx[&inst.operands[0].unwrap_id_ref()];
-                let callee_had_err =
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/linker/simple_passes.rs
```diff
@@ -237,10 +237,18 @@ pub fn check_fragment_insts(sess: &Session, module: &Module) -> super::Result<()
         let mut any_err = None;
         for inst in module.functions[index].all_inst_iter() {
             if inst.class.opcode == Op::FunctionCall {
-                let callee = func_id_to_idx[&inst.operands[0].unwrap_id_ref()];
-                let callee_had_err =
-                    visit(sess, module, visited, stack, names, callee, func_id_to_idx).err();
-                any_err = any_err.or(callee_had_err);
+                let callee_id = inst.operands[0].unwrap_id_ref();
```
eddyb, line 240: For the record, this code in `simple_passes` isn't incorrect.

The underlying bug¹ is that `OpFunctionCall` is still used with function pointers - the correct instruction is `OpFunctionPointerCallINTEL` (from the same SPIR-V extension that adds function pointers through a new `CodeSectionINTEL` Storage Class).

<sub>¹(which one of my branches fixes, been meaning to open a PR for that for a while, already landed most of its prerequisites earlier this year)</sub>

Issue #455, opened 2025-10-29
DCE for unused shared memory arguments


nazar-pc:
When debugging code, I often end up commenting-out pieces of code, which sometimes results in DCE that leaves shared memory arguments unused.

Unfortunately this results in validation error:
```
error: error:0:0 - [VUID-StandaloneSpirv-None-10684] Invalid explicit layout decorations on type for operand '290[%_ptr_Workgroup__struct_106]'
         %scratch_space = OpVariable %_ptr_Workgroup__struct_106 Workgroup
  |
  = note: spirv-val failed
  = note: module `file.spv`
```

It'd be really nice from DX perspective to also DCE this unused workgroup argument so that it doesn't end up in SPIR-V file and passes validation as the result.

Firestar99:
We have a linker pass called [explicit_layout](https://github.com/Rust-GPU/rust-gpu/blob/main/crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs) that duplicates the type layout declarations and removes the explicit layout from them. Primarily used for workgroup shared memory, which must explicitly *not* have an explicit layout since a few spirv-val versions, which seems to be failing here. 
Afaik all params in the entry point are considered roots for DCE, so it will never remove them on it's own. It's usually `spirv-opt` that may do so. Though I wonder if it's worth implementing a DCE now, when https://github.com/Rust-GPU/rust-gpu/issues/449 will likely refactor the entire entry point and require a new analysis for "used params" anyway. 
(We should still see if we can fix the linker pass though)

eddyb:
I ran into the `OpEntryPoint` list of globals recently, and for a while I was extremely confused (but the issue turned out to be an outdated set of live functions, being passed around, and accidentally "reviving" a dead function or something).

Because the solution was very simple (and entirely separate from that whole system), I didn't start poking at it, but I was considering adding some "used globals" tracking to `spirt::spv::lift`, so it can regenerate the list entirely on its own, and remove all hazards associated with it.

If anyone wants to test a quick fix (which I'm surprised I've forgotten I could "just" use):
https://github.com/Rust-GPU/rust-gpu/blob/2232fcb7b5f44415a573ba5ed3b2619f12559294/crates/rustc_codegen_spirv/src/linker/spirt_passes/explicit_layout.rs#L44-L49
That whole loop can just be `eraser.in_place_transform_module(module);`.

[...]

PR #456, opened 2025-11-03
Fix unused shared memory failing validation

Firestar99:
This fixes validation errors caused by unused shared memory and enables it to be DCE-ed away. 

Params of entry point interfaces will now be DCE-ed by rust-gpu as well, previously only `spirv-opt` does *some* DCE on them but it leaves unused shared memory params as-is. As with `spirv-opt`, `SpirvBuilder.optimizer.preserve_bindings = true` will disable any DCE of unused entry point params. 

On a vulkan1.1/spv1.3 or lower `preserve_bindings` does not work on any param that is not of class input or output, so shared memory will be DCE-ed anyway, due to there being no references to those instructions. Whereas vulkan1.2+/spv1.4+ added the requirement for all `OpVariable`s to be mentioned in the entry point instruction, allowing us to preserve them.

close https://github.com/Rust-GPU/rust-gpu/issues/455

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/entry_interface.rs
+++ b/crates/rustc_codegen_spirv/src/linker/entry_interface.rs
@@ -13,7 +13,7 @@ type Id = Word;
 ///
 /// This is needed for (arguably-not-interface) `Private` in SPIR-V >= 1.4,
 /// but also any interface variables declared "out of band" (e.g. via `asm!`).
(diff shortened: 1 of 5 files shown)


Firestar99:
@eddyb requesting rereview for the 2 new commits. I noticed that on `vulkan1.1` we did DCE the unused shared memory, so investigated a bit and found a way to DCE it on other targets as well (see desc).

LegNeato (approved):

Issue #449, opened 2025-10-22
RFC: entry point redesign ideation


Firestar99:
Roughly ordered from easiest to hardest to design / implement:
* builtins as functions in spirv-std, that any code can just call instead of relying on the callees to patch through the correct variables. Needs validation during linking, since buildins are often valid only in some shader types.
* wgpu "bind group layout" / vulkan "descriptor set layout" should have some kind of direct representation within rust-gpu directly. I'm thinking of a struct declaring all the bindings of *one* descriptor set, soft of like wgsl does it:
```rust
#[spirv(descriptor_set)]
struct BindGroupForSystemX {
  #[binding(0)] data: TypedBuffer<MyData>,
  #[binding(1)] base_color: Image2D,
  #[binding(2)] sampler: Sampler,
}
```
It would be amazing if you could pass them around within this struct (you currently can't), since you could do something like this:
```rust
#[spirv(fragment)]
fn my_entry(
    #[spirv(descriptor_set = 0)] system_x: &BindGroupForSystemX,
) {
  systemx::process(system_x)
}
```
Note how the descriptor_set / bind group id is declared at the entry point, so you could move it to whatever id you'd want and thus avoid conflicts if two systems use the same id? Also I pass it in as a reference, so you can't change eg. the buffer that's in `systemx.data` to another. We may want that as a feature, but assigning it non-uniformly makes it potentially problematic.
* small extra: Have a non-generic texture that can be unsafely cast to all the generic textures we have right now. For spirv, we need to automatically add a new texture binding of that texture type and the same binding id when casts happen. Super useful for bindless! See [rust-gpu zulip, research, slang `__DynamicResource`](https://rust-gpu.zulipchat.com/#narrow/channel/494264-research/topic/slang.20.60__DynamicResource.60/with/537723261)
* "pipeline layout" should probably also some some representation as well. It declares the ids of the descriptor sets and the push constants. I'd move it to a struct similarly to above. This also makes it trivial to ensure a vertex and fragment shader have the same pipeline layout.
```rust
#[spirv(pipeline_layout)]
struct MyPipelineLayout {
  #[descriptor_set(0)] system_x: BindGroupForSystemX,
  #[descriptor_set(1)] system_y: BindGroupForSystemY,
  #[push_constant] push: MyPushConstant,
}

#[spirv(vertex)]
fn my_fragment(pipeline: MyPipelineLayout) {...}
#[spirv(fragment)]
[...]

PR #459, opened 2025-11-03
spirv-std functions to query builtins

Firestar99:
This is a prototype for adding functions within spirv_std to query SPIR-V builtin read-only global variables. This is the very first part of https://github.com/Rust-GPU/rust-gpu/issues/449. 

Currently, the prototype in this PR adds the function `spirv_std::builtin::compute::local_invocation_id() -> UVec3` to query the `local_invocation_id` without having to pass it through from the entry point declaration. Interestingly, the linker infrastructure for this to work was already in place for other reasons (dynamically adding `OpVariable`s during codegen), so this PR just needs to adds the function with the correct inline assembly. 

More functions to come.

### DIFF
--- a/crates/spirv-std/src/builtin.rs
+++ b/crates/spirv-std/src/builtin.rs
@@ -1,0 +1,27 @@
+//! Symbols to query SPIR-V read-only global built-ins
+
+/// compute shader built-ins
+pub mod compute {
+    #[cfg(target_arch = "spirv")]
+    use core::arch::asm;
+    use glam::UVec3;
+
+    /// GLSL: `gl_LocalInvocationID()`
+    /// WGSL: `local_invocation_id`
+    #[doc(alias = "gl_LocalInvocationID")]
+    #[inline]
+    #[gpu_only]
+    pub fn local_invocation_id() -> UVec3 {
+        unsafe {
+            let result = UVec3::default();
+            asm! {
+                "%builtin = OpVariable typeof{result} Input",
+                "OpDecorate %builtin BuiltIn LocalInvocationId",
+                "%result = OpLoad typeof*{result} %builtin",
+                "OpStore {result} %result",
+                result = in(reg) &result,
+            }
+            result
+        }
+    }
+}
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -92,6 +92,7 @@ pub use macros::spirv;
 pub use macros::{debug_printf, debug_printfln};
 
 pub mod arch;
+pub mod builtin;
 pub mod byte_addressable_buffer;
 pub mod debug_printf;
 pub mod float;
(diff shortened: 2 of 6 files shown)


nazar-pc:
Will using such functions instead of arguments mean lower register usage overall, or how do those things interact with each other across function calls?

Firestar99:
It emits semantically the same spirv code as previously: define a builtin global variable and read it. Just that you can now do this anywhere in your codebase, rather than having to pass it down from the entry point and have to hope noone accidentally passes the wrong thing. 
Register usage heavily depends on the target platform. See [AMD RDNA3 ISA pdf](https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna3-shader-instruction-set-architecture-feb-2023_0.pdf) on "3.5. Initial Wave State", there you'll see compute shaders use 4 scalar registers (SGPR, once per workgroup) for `workgroup_id` plus some bitflags. For `local_invocation_id` they seem to have been changes in RDNA3, it seems like it only uses 1 vector register (VGPR, once per invocation / thread) now (even though the table in the pdf defines it for GS (eg. mesh shaders) not CS, I assume it's the same), whereas RDNA2 used 3 VGPRs, one for each axis (see ["3.12.2. SGPR Initialization"](https://www.amd.com/content/dam/amd/en/documents/radeon-tech-docs/instruction-set-architectures/rdna2-shader-instruction-set-architecture.pdf)).

fluffysquirrels:
I'm working on top this code in my fork at <https://github.com/fluffysquirrels/rust-gpu/tree/spirv-std-builtin-fn>. So far I've added the rest of the compute builtins I know of.

I don't recall what options there are on GitHub for me to push to the PR branch in this repo. I'm happy to push my work wherever you suggest.

fluffysquirrels:
I've only worked with compute shaders, not the various graphics shaders.

However I see the [list of supported builtins](https://github.com/Rust-GPU/rust-gpu/blob/8eecf61e0f3a0d9a822206673dab092985fe398b/crates/rustc_codegen_spirv/src/symbols.rs#L40-L155) in `rustc_codegen_spirv`, and the compile test [`all-builtins`](https://github.com/Rust-GPU/rust-gpu/blob/8eecf61e0f3a0d9a822206673dab092985fe398b/tests/compiletests/ui/spirv-attr/all-builtins.rs).

Question 1: Shall I try to add getter functions for each of these graphics builtins also?

---

Question 2: These compute builtins have very simple implementations, but do you think they still deserve more automated testing? If so, what do you think would be appropriate to test?

My first thought is to run a very simple compute grid with `wgpu` and have each invocation store its ID's to an output buffer, then the CPU can assert each one is correct.

fluffysquirrels:
I've created a new PR (#535), only to share my progress on this. If you want to merge that back into this one somehow, please go ahead.

Firestar99:
superseded by https://github.com/Rust-GPU/rust-gpu/pull/535

PR #463, opened 2025-11-15
Create nix flake

jla2000:
I have created a nix flake, that will fetch the required rust toolchain using [rust-overlay](https://github.com/oxalica/rust-overlay).
By doing this, the toolchain does not need to be installed into the user environment directly.
I also added an `.envrc`, this will allow to directly drop into a devshell if eg. [direnv](https://github.com/direnv/direnv) is used.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,26 @@
+{
+  description = "rust-gpu";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixpkgs-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { nixpkgs, rust-overlay, flake-utils, ... }:
+    flake-utils.lib.eachDefaultSystem (system:
+      let
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+          config.allowUnfree = true;
+        };
+        rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+      in
+      {
+        devShells = {
+          default = pkgs.callPackage ./shell.nix { inherit rustToolchain; };
+          android = pkgs.callPackage ./android.nix { inherit rustToolchain; };
+        };
+      });
+}
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ tests/Cargo.lock
 .github/install-spirv-tools/Cargo.lock
 rustc-ice-*.txt
 .idea
+.direnv
--- a/android.nix
+++ b/android.nix
@@ -9,9 +9,9 @@
 # (you can also replace `cargo apk build` with `cargo apk run` to launch it,
 # via `adb`, into either the Android Emulator, or a physical Android device)
 
-let
-  pkgs = import <nixpkgs> {};
-in with pkgs; mkShell rec {
+{ pkgs, rustToolchain, ... }:
+
+with pkgs; mkShell rec {
   # Workaround for https://github.com/NixOS/nixpkgs/issues/60919.
   # NOTE(eddyb) needed only in debug mode (warnings about needing optimizations
   # turn into errors due to `-Werror`, for at least `spirv-tools-sys`).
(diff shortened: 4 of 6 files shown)


LegNeato:
I don't know anything about nix, @eddyb ?

alphastrata:
not maintainer --but I think forcing everyone to pull in the android toolchain might be a bit of an ask no?

flake.nix
```diff
@@ -0,0 +1,26 @@
...
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+          config.allowUnfree = true;
+        };
+        rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
+      in
+      {
+        devShells = {
+          default = pkgs.callPackage ./shell.nix { inherit rustToolchain; };
+          android = pkgs.callPackage ./android.nix { inherit rustToolchain; };
```
Firestar99, line 23: @alphastrata This means you need to run `nix develop .#android` to get a dev shell with android, otherwise it'll just use default. (Though it wouldn't allow you to run the examples locally, since it's missing the deps that `shell.nix` provides... Maybe we should treat android as a purely additive thing?)

It looks good to me, but I'm not super familiar with nix myself, so I'll get @eddyb review it as well.

alphastrata: Ahh I see. ty.

jla2000: I'm not sure for what the android.nix was used for, I am assuming just to cross compile to android? If so, I think it's fine that it won't allow to run the examples per default as it was already like this before.

jla2000:
One thing I am not sure about yet is how the rust-overlay flake input can be automatically updated when upgrading to a newer toolchain without forcing someone to use nix. With these changes, every time the rust toolchain is being updated, it would be necessary to run "nix flake update rust-overlay" as well to make sure the new toolchain version is known to nix.

flake.nix
Firestar99: You can compile the `example-runner-wgpu` to Android. But it's not [documented in the example readme](https://github.com/Rust-GPU/rust-gpu/blob/d0f5c9e730cfd8516624c9e7dde61b1d8c22bfde/examples/README.md), could only find our [Android CI](https://github.com/Rust-GPU/rust-gpu/blob/3b1ad48caebfa5cd6b006c94d9364466d0b12a73/.github/workflows/ci.yaml#L108-L129) building it.

jla2000: What packages would be missing? I could just try add them to the android devshell

flake.nix
```diff
@@ -0,0 +1,26 @@
+{
+  description = "rust-gpu";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixpkgs-unstable";
```
eddyb, line 5: I'm not sure what the proper solution is but I will note that this approach has failed in the past:

https://github.com/Rust-GPU/rust-gpu-shadertoys/pull/43#issuecomment-3199776153

Keeping `default.nix` (as `shell.nix`, which I suppose is more appropriate), does make this less drastic, and I can explicitly avoid touching the flake path until it's usable for myself on NixOS etc.

But it's hard to tell what value of flakes really is outside of e.g. CI.

.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
eddyb, line 1: Is this common practice? I've never used a repo that came with an `.envrc`, and always did it locally, e.g. by running this (with a check that makes it idempotent, just in case I accidentally run it twice in the same directory):
```sh
git check-ignore -q .envrc || \
  echo /.envrc >> $(git rev-parse --git-path info/exclude) && echo use nix >> .envrc && direnv allow
```
<sup>(split onto two lines for readability on GitHub, it's just a oneliner in my shell history)</sup>

I use `fish` so typing any of `envrc`/`use nix`/`direnv allow`, and pressing :arrow_up:, is enough to recall it, but it is a weird thing I came up with when starting to use `direnv` (having already been using `.git/info/exclude` here and there whenever I didn't want to modify `.gitignore` etc.).

jla2000: Not sure if this is really common, I do it for most of my projects as I will almost every time want to use a nix devshell.
But I think it does not hurt much to have it

flake.nix
jla2000: Imo just the reproducibilty, using the same `default.nix` could fail on different systems, while a devshell coming from a flake would be more stable. But ofc this will require actively maintaining the `flake.lock`

Issue #468, opened 2025-11-26
CI: difftests on windows failing due to wgpu DXC errors


Firestar99:
CI on main is failing on windows due to DXC compiler errors stemming from within wgpu. Even though wgpu should abstract that sort of stuff. I suspect and env change, github updating the runners or something.

[Example failure](https://github.com/Rust-GPU/rust-gpu/actions/runs/19693098844/job/56412796220#step:11:931):
```rust
wgpu error: Validation Error

Caused by:
  In Device::create_compute_pipeline, label = 'Compute Pipeline'
    Internal error: DXC compile error: Compute Shader:138:14: error: expected identifier
            [8][8] _value3[2][8][8] = _e378;
             ^
Compute Shader:138:17: error: expected identifier
            [8][8] _value3[2][8][8] = _e378;
                ^
Compute Shader:138:20: error: use of undeclared identifier '_value3'
            [8][8] _value3[2][8][8] = _e378;
...
```

* Have not found any issues on wgpu's repo that look like this
* Maybe a simple `cargo update` does it? **no change**
* we're on wgpu v26, there's already wgpu v27 which may fix it?
* can't repo it locally, but I'm not on 25H2 yet

## Suspicion 1: wgpu is using DX12
* Why is it compiling for DX12 anyway? Wouldn't we want to run on Vulkan software renderers anyway?
  * Vulkan drivers may be missing?
* `workgroup_memory-ash` is failing as well with:
```
thread 'main' panicked at arch\workgroup_memory\workgroup_memory-ash\src\main.rs:49:10:
called `Result::unwrap()` on an `Err` value: Failed to create Vulkan instance

Caused by:
    Unable to find a Vulkan driver
```
* I suspect that because wgpu can't see the Vulkan driver, it is now picking some Windows Software DX12 driver. And thus transpiling the spirv to hlsl and compiling it with DXC, which it didn't do previously. So the DXC error may have been there since who knows how long, but switching from Vulkan to DX12 caused it to pop up.

# TLDR
* wgpu started using the DX12 backend instead of the Vulkan one
  * vulkan drivers are still present, no idea why that happened
* using DX12 means naga transpiles spirv to hlsl, for the DXC compiler
  * the transpiled hlsl is broken, always has been, but we never tested DX12 so noone noticed
  * this is already fixed in naga 27, we're still on wgpu 26
  * Broken code snippet, look at the types:
```hlsl
uint _e378[2][8][8] = phi_841_;
{
[...]

PR #469, opened 2025-11-26
difftest: force wgpu to use Vulkan, disable dx12

Firestar99:
ignore the branch name, just reusing the PR from a previous attempt

fixes CI on main, see https://github.com/Rust-GPU/rust-gpu/issues/468

close https://github.com/Rust-GPU/rust-gpu/issues/468

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,7 +770,6 @@ dependencies = [
  "ash",
  "bytemuck",
  "futures",
- "naga",
  "serde",
  "serde_json",
  "spirv-builder",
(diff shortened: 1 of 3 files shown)


schell (approved):

Firestar99:
* naga 27 has a fix for the incorrectly emitted hlsl, we're still on wgpu 26 (and thus naga 26)
* curiously, this pipeline ran but the merge queue pipeline failed due to not finding a Vulkan driver
  * just gonna retry for now, may need more investigation

PR #472, opened 2025-12-01
Ignore: debug log wgpu, displaying available drivers

Firestar99:


Firestar99:
debug log didn't work, and I won't bother trying any further

nazar-pc:
macOS runners have a functional paravirtualized GPU enabled, on Linux/Windows it is just LLVMpipe/SwiftShader if/when you enable them. Not sure if that is what you were looking for though.

Firestar99:
See https://github.com/Rust-GPU/rust-gpu/issues/468

Last week wgpu started preferencing DX12 over Vulkan for whatever reason. We assume runner updates may have added a DX12 driver on Windows, which previously wasn't there, so I wanted to enable debug logging to see what wgpu logs say. But I can't get it to log at all, even locally, with `env_logger` setup and everything, no idea why cause I got some logs last week. Since disabling DX12 has (mostly*) fixed our problems, won't bother spending any more time on this.

*we're seen some CI fail where supposedly no vulkan drivers were present, even though swiftshader installed successfully.

PR #474, opened 2025-12-01
Image texel write: allow writing scalar values into scalar images

Firestar99:
* Breaking: `Image::write()` no longer takes any vector component count (eg. for float images Vec2, Vec3 or Vec4), now type must match image's `SampleResult` exactly
* usage of this function is relatively rare, since it's only used in compute shader (not fragment!) write directly to images, which more advanced graphics pipelines often do for post-processing effects
* `Image!(type=f32)` aka. not specifying a specific image format (generic `format=Unknown`): `SampleResult=Vec4` instead of allowing any of Vec2, Vec3 or Vec4 (loads similarly always return a Vec4)
* for `Image!(format=rg32f)`:  `rg` -> Vec2, `rgb` -> Vec3, `rgba` -> Vec4
* for `Image!(format=r32f)`:  broken previously since you couldn't pass in a `f32` as it doesn't impl `Vector`, now allows passing f32 as expected

close https://github.com/Rust-GPU/rust-gpu/discussions/412


LegNeato:
I don't know enough about this area to review, @schell or @eddyb ?

PR #475, opened 2025-12-02
Minor fixes

Firestar99:
see individual commits

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -244,11 +244,11 @@ jobs:
           # FIXME(eddyb) consider using lavapipe instead, or even trying both.
           install_swiftshader: true
           # install_lavapipe: true
-      - name: Install rustup components
-        run: rustup component add rustfmt clippy
       # cargo version is a random command that forces the installation of rust-toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -244,11 +244,11 @@ jobs:
           # FIXME(eddyb) consider using lavapipe instead, or even trying both.
           install_swiftshader: true
           # install_lavapipe: true
-      - name: Install rustup components
-        run: rustup component add rustfmt clippy
       # cargo version is a random command that forces the installation of rust-toolchain
       - name: install rust-toolchain
         run: cargo version
+      - name: Install rustup components
+        run: rustup component add rustfmt clippy
```
nazar-pc, line 251: Shouldn't even need these, with rust-toolchain.toml they should be installed automatically

Firestar99: True, but it gives better accuracy how long each step takes, when you're comparing CI times like in https://github.com/Rust-GPU/rust-gpu/pull/334

(not that it really matters for lint, but why not make it match the others)

LegNeato (approved):
🚀

Issue #476, opened 2025-12-03
Remove the need for `crate-type = ["dylib"]`


nazar-pc:
Right now for shader crate to compile `crate-type = ["dylib"]` is needed to be specified in `Cargo.toml`.

This is annoying in case when a library is not just a shader, but also a regular dependency of other crates since this makes compiler generate extra `libshader_crate.so` shared objects in `target/{release,debug}`.

This can be worked around with `cargo rustc --crate-type dylib` when compiling normally, so maybe codegen backend or `cargo-gpu` could do something like that transparently and implicitly?

~~Feel free to move this into `cargo-gpu` repository if it makes more sense to be there.~~ Has nothing to do with cargo-gpu.

Firestar99:
FYI declaring it as `crate-type = ["lib", "dylib"]` will make the CPU build link in the static `lib` instead of using the `*.so`, but will still needlessly build the `*.so`.

Firestar99:
I would love to remove that requirement! 

`dylib` being used so rarely has [revealed some issues](https://github.com/rust-lang/cargo/issues/16000) in the past as well, causing weird failures. 

On `--crate-type dylib`: while the `cargo rustc` cmd does have a `--crate-type`, `cargo build` or `run` or `test` do not. And we primarily rely on [`cargo build` to build and link everything](https://github.com/Rust-GPU/rust-gpu/blob/6fb875068d0d35b1a81ae89b73cf7a464f8c60f7/crates/spirv-builder/src/lib.rs#L1000-L1012). So we'd either need `cargo` to add support for that, or figure out some other way to extract all the `rlib`s we need to link from cargo. 

Related: In [RFC: link-time capabilities](https://github.com/Rust-GPU/rust-gpu/issues/388), I've suggested this: 
> Changing link-time capabilities does not require a clean recompile, unlike the current spirv capabilities.

This would also need some kind of custom linking, though the information retrieval from cargo on what needs to get linked would likely be separate from however any variations of linking we do internally.

nazar-pc:
> FYI declaring it as `crate-type = ["lib", "dylib"]` will make the CPU build link in the static `lib` instead of using the `*.so`, but will still needlessly build the `*.so`.

I know, that is exactly what I do right now.

[...]

PR #477, opened 2025-12-04
No longer require `crate-type = ["dylib"]` on shader crates

Firestar99:
**With this PR, you no longer need to declare the `crate-type` in your shader crate.**

All shader crates can now remove the previously required declaration in `Cargo.toml` below:
```toml
[lib]
crate-type = ["dylib"]
```

### Changes
* no longer requires `crate-type` to be set in your shader crate
* switch spirv-builder from `cargo build` to `cargo rustc`
  * allows passing `--crate-type=dylib`
  * the difference between the commands is minor, `rustc` mostly just exposes additional rustc args for the target crate
* the `cargo <cmd>` is configurable
  * future investigation: may allow calling other cargo commands, such as `clippy` or `fmt`

close https://github.com/Rust-GPU/rust-gpu/issues/476

### cargo-gpu

https://github.com/Rust-GPU/cargo-gpu/pull/125

Needs to be updated after this PR is merged, and needs some fixups from other changes in master, so a simple cargo patch won't work.

### DIFF
--- a/examples/runners/wgpu/Cargo.toml
+++ b/examples/runners/wgpu/Cargo.toml
@@ -7,6 +7,10 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[lib]
+# required by cargo apk / Android
+crate-type = ["lib", "cdylib"]
+
 # See rustc_codegen_spirv/Cargo.toml for details on these features
 [features]
 default = ["use-compiled-tools"]


crates/spirv-builder/src/lib.rs
```diff
@@ -995,21 +1005,26 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
...
+        cargo_cmd,
         "--lib",
         "--message-format=json-render-diagnostics",
         "-Zbuild-std=core",
         "-Zbuild-std-features=compiler-builtins-mem",
         "--profile",
         profile,
     ]);
+    if cargo_cmd_like_rustc {
+        cargo.args(["--crate-type", "dylib"]);
```
LegNeato, line 1031: Could you include a comment for why we need `dylib` at all?

Firestar99:
Did another cleanup round, found some more crates that declare `crate-type=["dylib"]`  that I missed, and added a quick comment on crate types:

https://github.com/Rust-GPU/rust-gpu/blob/d7681df703dc135d370619d5dad65c5b907f04dc/crates/spirv-builder/src/lib.rs#L1026-L1030

Firestar99:
From [cargo apk](https://github.com/rust-mobile/cargo-apk/pull/28#issuecomment-1714502555) repository about using `cargo rustc` instead of `cargo build`:
> The unfortunate downside of this, as shown by [the last CI run](https://github.com/rust-mobile/cargo-apk/actions/runs/5398044617) is that we can no longer build multiple targets such as examples at once. Neither `--examples` nor explicitly specifying `--example hello_world --example jni_audio --example looper` is allowed, all result in `error: crate types to rustc can only be passed to one target, consider filtering`.

I don't think this affects us at all, since we're only ever building the `--lib` for shaders, never any examples or executables.

LegNeato:
I'll let @eddyb review as I am not sure of ramifications.

nazar-pc:
They already approved, but looks like the APIs used in the action to install Vulkan SDK on Windows is not using tokens, so it exceeds rate limits sometimes, which prevented merging: https://github.com/Rust-GPU/rust-gpu/actions/runs/19971068746/job/57275658889

Issue #481, opened 2025-12-07
Conflicting Output Location Assignment


marstaik:
### Expected Behaviour
I've used multiple vertex -> fragment outputs before with simpler data structures, but my attempt to pass two `&mut Struct` seem to be failing with `conflicting output locations`.

### Example & Steps To Reproduce

Relevant source code:

```rust
pub struct BlinnPhongMaterialData {
	pub albedo:            Vec4,
	pub specular_color:    Vec4,
	pub shininess:         f32,
	pub specular_strength: f32,
}

pub struct BlinnPhongVertexData {
	pub position: Vec4,
	pub normal:   Vec4,
	pub view_dir: Vec4,
}

#[spirv(vertex)]
#[scene_descriptors]
pub fn blinn_phong_material_vert(
	#[camera] cameras: &[Camera],
	#[instance_data] instance_data: &[InstanceData],
	#[material_data] material_data: &[BlinnPhongMaterialData],
	#[spirv(instance_index)] instance_index: usize,
	// vert
	vertex: PositionNormal,
	// outputs
	out_vertex: &mut BlinnPhongVertexData,
	out_material: &mut BlinnPhongMaterialData,
	#[spirv(position)] out_pos: &mut Vec4,
) {
	// Irrelevant body, occurs even when fully commented out
}

#[spirv(fragment)]
#[scene_descriptors]
pub fn blinn_phong_material_frag(
	#[framebuffer_info] framebuffer_info: &FramebufferInfo,
	#[ambient_lights] ambient_lights: &[AmbientLight],
	#[directional_lights] directional_lights: &[DirectionalLight],
	#[point_lights] point_lights: &[PointLight],
	#[spot_lights] spot_lights: &[SpotLight],
	// outputs
	vertex: BlinnPhongVertexData,
	#[spirv(flat)] material: BlinnPhongMaterialData,
	// frag
	frag_color: &mut Vec4,
) {
	// Irrelevant
}
```

## System Info
 - RustGpu - main `e97524f6b4816056b3edaa70c3e0e0c656392c05`
 - Rust: `rustc 1.90.0-nightly (35f603652 2025-06-29)`
 - OS: `fedora`
 - GPU: `NVIDIA RTX 4090`
 - SPIR-V: `SPIRV-Tools v2025.4 v2025.4-0-g7f2d9ee9`

## Backtrace

```
[...]

PR #482, opened 2025-12-08
Fix location assignment

Firestar99:
Locations for Input, Output (and UniformConstant, where I have no idea how they work *without* explicit descriptor_set and binding) were assigned one per declaration, whereas they should have been assigned ~~one per 16 bytes~~ following the [Vulkan spec on 16.1.4. Location and Component Assignment](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#interfaces-iointerfaces-locations). 

close https://github.com/Rust-GPU/rust-gpu/issues/481.

Also adds an explicit location assignment, that keeps on counting from the last assigned value, just like glsl does:
https://github.com/Rust-GPU/rust-gpu/blob/d00ba029586d0b983320e94836671ecdedd950e3/tests/compiletests/ui/spirv-attr/location_assignment_explicit.rs#L25-L32

We **don't** check for overlaps when locations are manually assigned, leaving that for spirv-val:
https://github.com/Rust-GPU/rust-gpu/blob/d00ba029586d0b983320e94836671ecdedd950e3/tests/compiletests/ui/spirv-attr/location_assignment_explicit_overlap.stderr#L44-L48

See vulkan spec on [16.1.4. Location and Component Assignment](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#interfaces-iointerfaces-locations) and on [16.9.4. Offset and Stride Assignment](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#interfaces-resources-layout) of structs. spirv spec is irrelevant here, it simply declares the Location decoration exists, and leaves all further spec for the client APIs to handle.

### DIFF
--- a/crates/rustc_codegen_spirv/src/spirv_type.rs
+++ b/crates/rustc_codegen_spirv/src/spirv_type.rs
@@ -293,11 +293,19 @@ impl SpirvType<'_> {
                 }
                 locations
             }
-            Self::Matrix { element, count } => cx.lookup_type(element).location_size(cx)? * count,
+            Self::Matrix { element, count } => cx
+                .lookup_type(element)
+                .location_size(cx)?
+                .checked_mul(count)
+                .expect("overflow"),
             Self::Array { element, count } => {
-                let count = cx.builder.lookup_const_scalar(count).unwrap();
-                let count: u32 = count.try_into().unwrap();
-                cx.lookup_type(element).location_size(cx)? * count
+                let element = cx.lookup_type(element).location_size(cx)?;
+                let count = cx
+                    .builder
+                    .lookup_const_scalar(count)
+                    .and_then(|c| u32::try_from(c).ok())
+                    .expect("SpirvType::Array.count to be a u32 constant");
+                element.checked_mul(count).expect("overflow")
             }
             _ => return None,
         };
@@ -313,11 +321,19 @@ impl SpirvType<'_> {
             Self::Integer(width, _) | Self::Float(width) => Size::from_bits(width),
             Self::Adt { size, .. } => size?,
             Self::Vector { size, .. } => size,
-            Self::Matrix { element, count } => cx.lookup_type(element).sizeof(cx)? * count as u64,
+            Self::Matrix { element, count } => cx
+                .lookup_type(element)
+                .sizeof(cx)?
+                .checked_mul(count as u64, cx)
+                .expect("overflow"),
             Self::Array { element, count } => {
-                let count = cx.builder.lookup_const_scalar(count).unwrap();
-                let count = count.try_into().unwrap();
-                cx.lookup_type(element).sizeof(cx)? * count
+                let element = cx.lookup_type(element).sizeof(cx)?;
+                let count = cx
+                    .builder
+                    .lookup_const_scalar(count)
+                    .and_then(|c| u64::try_from(c).ok())
+                    .expect("SpirvType::Array.count to be a u32 constant");
+                element.checked_mul(count, cx).expect("overflow")
             }
             Self::Pointer { .. } => cx.tcx.data_layout.pointer_size,
             Self::Image { .. }


marstaik:
Trying this branch out with this code:
```rust

#[shader_type]
pub struct BlinnPhongMaterialData {
	pub albedo:            Vec3,
	pub specular_color:    Vec3, // specular tint (often white for dielectrics)
	pub specular_strength: f32,  // controls the intensity of the specular highlight
	pub shininess:         f32,  // controls the size of the specular highlight
}

#[shader_type]
pub struct BlinnPhongFragData {
	pub position: Vec3, // surface position in world space
	pub normal:   Vec3, // surface normal in world space
	pub view_dir: Vec3, // direction from surface to camera
}

#[spirv(vertex)]
#[scene_descriptors]
pub fn blinn_phong_material_vert(
	#[camera] cameras: &[Camera],
	#[instance_data] instance_data: &[InstanceData],
	#[material_data] material_data: &[BlinnPhongMaterialData],
	#[spirv(instance_index)] instance_index: usize,
	// vert
	vertex: PositionNormal,
	// outputs
	out_material: &mut BlinnPhongMaterialData,
	out_frag: &mut BlinnPhongFragData,
	#[spirv(position)] out_pos: &mut Vec4,
) {
	/// 
}

#[spirv(fragment)]
#[scene_descriptors]
pub fn blinn_phong_material_frag(
	#[framebuffer_info] framebuffer_info: &FramebufferInfo,
	#[ambient_lights] ambient_lights: &[AmbientLight],
	#[directional_lights] directional_lights: &[DirectionalLight],
	#[point_lights] point_lights: &[PointLight],
	#[spot_lights] spot_lights: &[SpotLight],
	// outputs
	#[spirv(flat)] material: BlinnPhongMaterialData,
	frag: BlinnPhongFragData,
	// frag
	frag_color: &mut Vec4,
) {
	///
}
```

Still results in error:

```
  --- stderr
     Compiling rtd_scene_shaders v0.1.0 (/home/marios/proj/rtd/packages/rtd_scene_shaders)
  error: [VUID-StandaloneSpirv-OpEntryPoint-08722] Entry-point has conflicting output location assignment at location 2, component 0
    |
    = note: module `/home/marios/proj/rtd/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/rtd_scene_shaders.spv`

  warning: an unknown error occurred
    |
    = note: spirv-opt failed, leaving as unoptimized
    = note: module `/home/marios/proj/rtd/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/rtd_scene_shaders.spv`

  error: error:0:0 - [VUID-StandaloneSpirv-OpEntryPoint-08722] Entry-point has conflicting output location assignment at location 2, component 0
           OpEntryPoint Vertex %2 "materials::blinn_phong_material::blinn_phong_material_vert" %cameras %instance_data %material_data %instance_index %vertex %out_material %out_frag %out_pos
    |
    = note: spirv-val failed
    = note: module `/home/marios/proj/rtd/target/spirv-builder/spirv-unknown-vulkan1.2/release/deps/rtd_scene_shaders.spv`
  ```
  
  Double checked to make sure it was from the right branch:
  
image

**Comment by marstaik** [comment]
Note: I do use scalar-block-layouts if that matters - both enabled in Vulkan1.2, and I pass the command line argument for spirv-builder

**Comment by Firestar99** [comment]
We don't care about scalar block layout or std430, we're just using whatever layout rustc gives us to ensure it's the same if you're using the same struct on shader and CPU side (at least since https://github.com/Rust-GPU/rust-gpu/pull/380). If someone needs higher alignment (for silly wgpu uniform buffers or so), they should just `#[repr(align(16))]` the struct.

I'm not actually too sure how scalar block layout and spirv-val interact precisely. Even though it's not enabled by default, we've never experienced any issues wrt layout with it being disabled. Even though `SpirvBuilder.validator.scalar_bock_layout` exists.

Note that you are technically required to `#[repr(C)]` any struct shared between CPU and GPU, since rustc makes no guarantees about the rust layout. In theory, even two different compile invocations could result in different layouts, though in practice we haven't seen the layout algorithm change, like *ever*.

**Comment by Firestar99** [comment]
Even more minified code:

```rust
#[repr(C)]
#[derive(Copy, Clone, Default)]
pub struct ManyFloats {
    a: f32,
    b: f32,
    c: f32,
}

#[spirv(vertex)]
pub fn main(out1: &mut ManyFloats, #[spirv(location = 3)] out2: &mut f32) {
    const {
        assert!(size_of::<ManyFloats>() == 12);
    }
    *out1 = Default::default();
    *out2 = Default::default();
}
```

This fails as well, as according to spec array elements can't share slots, each array element has to get their own slot.
```rust
#[spirv(vertex)]
pub fn main(out1: &mut [f32; 3], out2: &mut f32) {
    *out1 = Default::default();
    *out2 = Default::default();
}
```

Firestar99:
I've reread the [Vulkan spec on location assignment](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#interfaces-iointerfaces-locations) and noticed that locations are not assigned for every 16 bytes, but there's a totally custom algorithm on how locations are assigned. Implemented the layout algorithm in `SpirvType::location_size()`, so our locations are hopefully correct now.

marstaik:
Success! :partying_face: 

> I'm not actually too sure how scalar block layout and spirv-val interact precisely. Even though it's not enabled by default, we've never experienced any issues wrt layout with it being disabled. Even though SpirvBuilder.validator.scalar_bock_layout exists.

Fun fact: I actually had it enabled only on the Vulkan12 feature side, and the spirv compiler threw an error and I had to enable it.

LegNeato:
I'll let @eddyb review this one as I am not qualified.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -757,6 +792,35 @@ impl<'tcx> CodegenCx<'tcx> {
...
+            if attrs.binding.is_some() {
+                self.tcx.dcx().span_fatal(
+                    location.span,
+                    "`#[spirv(location = ...)]` cannot be combined with `#[spirv(binding = ...)]`",
+                );
+            }
+            if !has_location {
+                self.tcx.dcx().span_fatal(
+                    location.span,
+                    "`#[spirv(location = ...)]` can only be used on Inputs (declared as plain values, eg. `Vec4`)\
```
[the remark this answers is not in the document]
Firestar99: squashed in

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -276,6 +276,34 @@ impl SpirvType<'_> {
...
+                for f in field_types {
+                    locations += cx.lookup_type(*f).location_size(cx)?;
+                }
+                locations
+            }
+            Self::Matrix { element, count } => cx.lookup_type(element).location_size(cx)? * count,
+            Self::Array { element, count } => {
+                let count = cx.builder.lookup_const_scalar(count).unwrap();
+                let count: u32 = count.try_into().unwrap();
+                cx.lookup_type(element).location_size(cx)? * count
```
schell: I'm not sure of what the numeric types are, here, but if we're going to `unwrap` we should mention why it's safe (or why we _want_ to panic) by using `expect`. Also - should we be using checked multiplication here to catch overflow in a very large array of very large types?

Firestar99: Sure, why not :D

But also: I hate our internal type system, *especially* the `size_of` below where we compute our own sizes. Imagine the terrifying amount of bugs that could stem from our `size_of` computing size differently from the rustc layout. (I remember showing this to eddyb and getting a scared response)

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -276,6 +276,42 @@ impl SpirvType<'_> {
...
+                for f in field_types {
+                    locations += cx.lookup_type(*f).location_size(cx)?;
+                }
+                locations
+            }
+            Self::Matrix { element, count } => cx
+                .lookup_type(element)
+                .location_size(cx)?
+                .checked_mul(count)
+                .expect("overflow"),
```
nazar-pc, line 300: This is not really more helpful than `.unwrap()`. Neither in runtime not during code review.

The best approach so far I have found in blockchain space:
* write the code in a way that expects no panics at all
* only use `.expect()` where panics never happen and in its message you write a proof that reviewer can verify

For example:
```rust
let n = NonZeroUsize::new(1).expect("Not zero; qed");
```

Firestar99: I very much agree with you, it would be a lot better to remodel the entire internal type system, cause it's a mess. I've already been floating a few suggestions, to express some of the complexity in standard library traits. Like a `trait ExplicitLayout` that is only implemented by types that have a reasonable layout in SPIR-V and can be read from or written to buffers. Specifically, they do not contain references.

PR #483, opened 2025-12-08
cargo-release setup

Firestar99:
Setup releases so that a simple `cargo release minor` (or `prerelease`) in the workspace root *should* do it. 

Note that it's been quite difficult to test this configuration, like *at all*. Test releases (without `--execute`) fail to verify the packaged crates due to using the "latest" version from crates.io instead of local changes. With the many breaking changes that requires upgrading more than one crate at once, basically all crates fail verification. Not even `--no-verify` works, since the generated `Cargo.toml` wants a new feature from one of our crates, that doesn't exist on `v0.9.0`, so it won't even package it. Only when we actually do the full release, package verification should work, since we're actively publishing the `v0.10.0` (or `*-alpha.0`) crates and redownloading them. So for the next release, we just release immediately and hope nothing breaks :skull: 

TLDR: Only the next release will show whether everything will just work or not, so I'll take the responsibility to do the next release and fix any issues as I encounter them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -->
 
+<!-- next-header -->
 ## [Unreleased]
 
 ### Changed 🛠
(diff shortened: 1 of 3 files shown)


LegNeato:
`cargo smartrelease` has a dry run mode that is super useful (https://github.com/GitoxideLabs/cargo-smart-release, should we use that instead? it is used for `gitoxide`)

examples/run-wasm/Cargo.toml
```diff
@@ -10,3 +10,6 @@ repository.workspace = true
 
 [dependencies]
 cargo-run-wasm = "0.3.2"
+
+[package.metadata.release]
+release = false
```
LegNeato, line 15: These shouldn't be needed as we already have `publish = false`?

Firestar99: `cargo release` doesn't respect `publish = false` unfortunately
(which feels like an oversight on their part?)

LegNeato: That seems...weird? I could have sworn it respects it when I used it in the past. This seems to imply they are not published? https://github.com/crate-ci/cargo-release/issues/877

LegNeato: Ok, looks like this is counterintuitive:

https://github.com/crate-ci/cargo-release/issues/923

Best to just stick with release. Ugh

Firestar99: Yea it won't release (at least I think it won't), but cargo release will still bump the version and such, which is just totally unnecessary for internal crates versioned at "0.0.0".

PR #484, opened 2025-12-11
Target specs v3

Firestar99:
Wanted to simply target specs even more:
* remove crate `rustc_codegen_spirv-target-specs`
  * this removes all the `spirv-unknown-*.json` target spec files in the repo
* generate target specs on the fly in spirv-builder, as originally [suggested by eddyb here](https://github.com/Rust-GPU/rust-gpu/pull/311#discussion_r2178318854)
* **NEW** spirv-builder accepts both full targets (`spirv-unknown-vulkan1.3`) and target envs (`vulkan1.3`)
* spirv-builder no longer checks that target env is a known valid one:
https://github.com/Rust-GPU/rust-gpu/blob/823b0c275e3eb8ab44ada5a6d989bdb57e8f9409/crates/rustc_codegen_spirv-types/src/target.rs#L8-L11
* target validation within the compiler is untouched, to be done the followup PR https://github.com/Rust-GPU/rust-gpu/pull/491
  * Learning from https://github.com/Rust-GPU/rust-gpu/pull/311 having had a lot of discussions on how to represent targets properly
* cargo-gpu see https://github.com/Rust-GPU/cargo-gpu/pull/128
  * this [cargo-gpu action](https://github.com/Rust-GPU/cargo-gpu/actions/runs/20230926648) shows this PR being tested against various rust-gpu releases and old revs
  * rust-gpu versions after [this PR] will still work with older cargo-gpu versions, until we update target specs in rust-gpu again (not planned anytime soon)
* in total with cargo-gpu, ~600 lines have been removed (including json)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,7 +54,7 @@ jobs:
         run: cargo test -p rustc_codegen_spirv --release --no-default-features --features "use-installed-tools"
 
       - name: workspace test (excluding examples & difftest)
-        run: cargo test --release --workspace --exclude "example-runner-*" --exclude "difftest*" --no-default-features --features "use-installed-tools"
+        run: cargo test --release --workspace --exclude "example-runner-*" --exclude "difftest*" --no-default-features --features "use-installed-tools,clap"
 
       # Examples
       - name: cargo check examples
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -73,6 +73,8 @@
 
 pub mod cargo_cmd;
 mod depfile;
+#[cfg(test)]
+mod tests;
 #[cfg(feature = "watch")]
 mod watch;
 
--- a/crates/spirv-builder/src/tests.rs
+++ b/crates/spirv-builder/src/tests.rs
@@ -1,0 +1,20 @@
+#[cfg(feature = "clap")]
+mod clap {
+    pub use crate::*;
+    use clap::Parser;
+
+    /// look at the output of this test to see what `--help` prints
+    #[test]
+    fn test_clap_help() {
+        // any malformed clap declarations should panic within clap's parse
+        match SpirvBuilder::try_parse_from(["", "--help"]) {
+            Ok(_) => panic!("`--help` should make clap return an error"),
+            Err(e) => {
+                let e = e.to_string();
+                println!("{}", e);
+                // sanity check help
+                assert!(e.contains("--target"));
+            }
+        };
+    }
+}


crates/rustc_codegen_spirv-types/src/target.rs
```diff
@@ -0,0 +1,73 @@
...
+        })?;
+
+        // used only to split the full list into groups.
+        #[allow(clippy::match_same_arms)]
+        match target_env {
+            // HACK(firestar99) this hardcoded list should be replaced with patterns (eg. `vulkan{}.{}`),
+            // which would allow us to support new target versions without updating this list.
+            "spv1.0" | "spv1.1" | "spv1.2" | "spv1.3" | "spv1.4" | "spv1.5" | "spv1.6" => {}
+            "opengl4.0" | "opengl4.1" | "opengl4.2" | "opengl4.3" | "opengl4.5" => {}
+            "vulkan1.0" | "vulkan1.1" | "vulkan1.1spv1.4" | "vulkan1.2" | "vulkan1.3"
```
nazar-pc: Can someone explain to me what is even the point of supporting OpenGL here? Is someone really expected to use it? If not, a bunch of APIs can be cleaned up that are OpenGL-specific in this repo.

Firestar99: You can also use SPIR-V with OpenGL, instead of giving it glsl source code directly. But some concepts like descriptor sets don't exist in OpenGL, they only have bindings. So validation needs to be configured differently, to allow OpenGL rules and not Vulkan rules, which would require specifying a descriptor set. It's mostly small details like that.

In reality, noone is testing that code path, and it's most likely broken *somewhere*. But if the effort is literally just allowing the target and forwarding it to `spirv-val`, why remove it?

Whereas SPIR-V for OpenCL shares some basics, its effectively an entirely different bytecode, which would be a lot more effort to support, so we don't bother.

nazar-pc: I know it is technically possible, but is it really worth supporting OpenGL/OpenCL?

There are currently capabilities in this repo that are not usable in Vulkan, so not having them would clean things up and reduce confusion if no one is really testing it or care about it anyway.

Firestar99: Which ones? Cause I'm not aware of any

nazar-pc: I was lazy to check myself, so I asked Grok: https://grok.com/share/bGVnYWN5LWNvcHk_237f5279-6772-431c-ae89-376a5ab95bda

Also there are memory semantics variants that are not usable or do not make sense in Vulkan: https://grok.com/share/bGVnYWN5LWNvcHk_b4879980-f9fa-496e-a6ea-5df6343f5384

All of those blow up the number of variants presented to the user. There are probably more examples of this in the codebase that I'm not familiar with yet.

Firestar99: It is true that there are some enum variants that are illegal in Vulkan, like `SequentiallyConsistent` and Rect Images, but that's pretty much it. Which I'd argue is not a lot of support nor effort to maintain.

All the OpenCL capabilities Grok mentioned we don't support anyway, and all the OpenGL image types are just the basic "Image" type in Vulkan that doesn't need any extra capabilities.

nazar-pc: `AtomicCounterMemory` also doesn't really work when used with `wgpu` and there are probably other things like that, which are technically acceptable in Vulkan, but not really useful. I just see no point in having all that noise in the API.

Firestar99: I feel like we should move this to a new issue, since it's rather unrelated to this PR. Also feel free to research which ones are actually useless or put up a PR for it.

LegNeato:
@eddyb thoughts?

eddyb (approved):

Firestar99:
merge failed due to lockfile conflicts, rebasing

Issue #346, opened 2025-07-23
Reporting warnings when building in build script


Firestar99:
If warnings are emitted during shader compile, they are often hidden by cargo only showing the output of a build scripts if it fails. But there's also `cargo::warning=MESSAGE` to emit warnings visible to the end user. We should add a system to forward these warnings, given that we already have `print_metadata`...

https://github.com/Rust-GPU/rust-gpu/blob/53c627bf7db77dfd92ab6301783f90aa73f6f24f/crates/spirv-builder/src/lib.rs#L375-L378

PR #488, opened 2025-12-15
Forward rustc warnings

Firestar99:
* replace `enum MetadataPrintout` with `struct BuildScriptConfig` for more flexibility
  * Primarily: `spirv_builder.build_script.defaults` allows us to add new options like below in the future
  * `MetadataPrintout::DependencyOnly` is now `spirv_builder.build_script.dependency_info = Some(true)` and defaults to `spirv_builder.build_script.defaults` (property above)
  * `MetadataPrintout::Full` is now `spirv_builder.build_script.env_shader_spv_path = Some(true)` and defaults to `false` (preventing conflicts with multimodule)
* forward rustc warnings and errors as build script warnings (via `cargo::warning={}`)
  * includes support for ansi colors!

### Example error

Image cause github doesn't support ansi in code blocks. Also note how in the shader warning, the final "`space-engine-shader` (lib) generated 1 warning" is missing. This is due to `cargo --quiet` also suppressing those, next to the other status messages. 
image

close https://github.com/Rust-GPU/rust-gpu/issues/346
close https://github.com/Rust-GPU/cargo-gpu/issues/103

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -392,7 +392,7 @@ pub struct BuildScriptConfig {
 
     /// Whether to emit an env var pointing to the shader module file  (via `cargo::rustc-env={}`). The name of the env
     /// var is the crate name with `.spv` appended, e.g. `sky_shader.spv`.
-    /// Not supported together with `multimodule=true` or `.watch()`.
+    /// Not supported together with `multimodule=true`.
     ///
     /// Some examples on how to include the shader module in the source code:
     /// * wgpu:
--- a/crates/spirv-builder/src/watch.rs
+++ b/crates/spirv-builder/src/watch.rs
@@ -55,9 +55,6 @@ impl SpirvWatcher {
             .as_ref()
             .ok_or(SpirvBuilderError::MissingCratePath)?
             .clone();
-        if builder.build_script.get_env_shader_spv_path() {
-            return Err(SpirvWatcherError::WatchWithPrintMetadata.into());
-        }
 
         let (tx, rx) = sync_channel(1);
         let watcher =
@@ -170,8 +167,6 @@ impl SpirvWatcher {
 
 #[derive(Debug, thiserror::Error)]
 pub enum SpirvWatcherError {
-    #[error("watching within build scripts will prevent build completion")]
-    WatchWithPrintMetadata,
     #[error("could not notify for changes: {0}")]
     NotifyFailed(#[from] notify::Error),
     #[error("watcher died and closed channel")]


crates/spirv-builder/src/lib.rs
```diff
@@ -731,17 +770,11 @@ impl SpirvBuilder {
...
-                        .strip_suffix(ARTIFACT_SUFFIX)
-                        .unwrap()
-                );
-                if self.print_metadata == MetadataPrintout::Full {
-                    println!("cargo:rustc-env={}={}", env_var, spirv_module.display());
+                if self.build_script.env_shader_spv_path() {
+                    println!(
+                        "cargo::rustc-env=SHADER_SPV_PATH={}",
+                        spirv_module.display()
+                    );
```
Firestar99, line 758: Since I'm already changing this part, I've considered adding some more ways to "send" the shader binaries from the build script into the source code, like multimodule support for env vars or better source code gen. But I feel like exploring that would be better in a followup PR.

Firestar99: Note: the code is outdated, I decided to go back on renaming the env var to `SHADER_SPV_PATH` since it breaks the android build

LegNeato:
Would it make sense to split builder into *two* public crates, one for using in build.rs and one for using from the CLI / other tools?

Firestar99:
I don't think we should split `spirv-builder` like that. `cargo-gpu` being built on top may also be used from different use-cases, and would suddenly need to use two different libraries. 

But what does make sense is splitting `cargo-gpu` into two crates, but *differently*: One `cargo-gpu` with all the cli goodness with all it's dependencies, specifically clap. The other `cargo-gpu-install` that just has the install action, since that's the only thing really needed for build scripts and `watch()`-ing binaries, and no cli deps. See [this issue](https://github.com/Rust-GPU/cargo-gpu/issues/105) and the rejected [PR1](https://github.com/Rust-GPU/cargo-gpu/pull/112) and [PR2](https://github.com/Rust-GPU/cargo-gpu/pull/117). EDIT: here's my attempt at it: https://github.com/Rust-GPU/cargo-gpu/pull/132

Setting a single bool when used from a build script is far simpler, just use whatever method you found in another project, and make sure you adjust this one value to switch between build script and watch in your binary mode. Though I do agree slightly error-prone since it's easy to miss.

crates/spirv-builder/src/lib.rs
```diff
@@ -1041,10 +1071,20 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
...
-    cargo.stderr(Stdio::inherit()).current_dir(path_to_crate);
+    if !builder.build_script.get_forward_rustc_warnings() {
+        cargo.stderr(Stdio::inherit());
+    }
+    cargo.current_dir(path_to_crate);
     log::debug!("building shaders with `{cargo:?}`");
     let build = cargo.output().expect("failed to execute cargo build");
 
+    if builder.build_script.get_forward_rustc_warnings() {
+        let stderr = String::from_utf8(build.stderr).unwrap();
```
fee1-dead: I don't like this unwrap because I don't know if rustc always prints in utf-8. I would suggest using `from_utf8_lossy`

crates/spirv-builder/src/watch.rs
```diff
@@ -55,7 +55,7 @@ impl SpirvWatcher {
             .as_ref()
             .ok_or(SpirvBuilderError::MissingCratePath)?
             .clone();
-        if !matches!(builder.print_metadata, crate::MetadataPrintout::None) {
+        if builder.build_script.get_env_shader_spv_path() {
             return Err(SpirvWatcherError::WatchWithPrintMetadata.into());
```
fee1-dead: Error should be renamed

Firestar99: Thanks for catching that one. 

I actually ended up removing this check entirely, as there's no technical reason anymore why you couldn't turn on this option and watch. Though it is completely nonsensical to watch within build scripts, the only place where this option makes sense to use, but maybe someone is reusing the same code for build scripts *and* watching within their application for hot code reload.

PR #489, opened 2025-12-16
ci: switch vk software impl to lavapipe

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/506

Switches from swiftshader to lavapipe for software rasterization, since swiftshader doesn't support int64 *for whatever reason*. Needed for int64 difftests in https://github.com/Rust-GPU/rust-gpu/pull/502.

### Old description, please ignore

swiftshader download has been giving us a lot of download timeouts, from https://github.com/Rust-GPU/rust-gpu/issues/407:
> 🚀 Installing SwiftShader library...
Error: HttpClientError: API rate limit exceeded for 52.159.244.80. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

The [install-vulkan-sdk-action](https://github.com/jakoch/install-vulkan-sdk-action) doesn't offer the option to forward credentials either. And I don't think you can use [retry actions](https://github.com/marketplace/actions/retry-step) to retry actual actions, they only support commands.

~~So by far the most low effort "fix" would be to switch to lavapipe and hope their servers don't have weird throttling. And our CI still works.~~ This won't work, both swiftshader and lavapipe artifacts are downloaded from [github releases](https://github.com/jakoch/rasterizers/releases).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -163,10 +163,7 @@ jobs:
           install_runtime: true
           cache: true
           stripdown: true
-
-          # FIXME(eddyb) consider using lavapipe instead, or even trying both.
-          install_swiftshader: true
-          # install_lavapipe: true
+          install_lavapipe: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - if: ${{ runner.os == 'Windows' }}


nazar-pc (commented):
To avoid throttling you should implement caching (or the action you're using).

I'm installing Lavapipe from https://github.com/mmozeiko/build-mesa releases, which are specifically designed for downloads and there is always a fresh Mesa version there.

Here is a PowerShell script I wrote that includes x86-64 and aarch64 support and does hash verification: https://github.com/nazar-pc/abundance/blob/ee1071d4a6002664b292230e216cfc5ec14bd667/.github/workflows/rust.yml#L271-L332

On Linux it is much simpler and no separate action is needed: https://github.com/nazar-pc/abundance/blob/ee1071d4a6002664b292230e216cfc5ec14bd667/.github/workflows/rust.yml#L265-L268

I hope it is helpful.

Firestar99:
On Linux we also just install mesa drivers, and Mac uses wgpu's metal backend (and we disable the one Vulkan test for now). So this is primarily about Windows.

The action has caching build in, it doesn't seem work on windows. Also it just downloads Swiftshader / Lavapipe without actually adding the registry key / env vars for it to work, we fix it ourselves. We're previously had an action that worked but stopped updating the vulkan sdk at some point. So yea, the state of currently available actions is underwhelming, and I have considered just rolling my own that can be shared by the wider rust-gpu ecosystem and hopefully be more stable.

This has literally been a "low effort" attempt at fixing it, but for some reason Windows really doesn't want to load that lavapipe driver, and the loader debug envs are not forwarded to the difftest for some reason. But if it continues being this unstable, I may really start rolling my own.

LegNeato (approved):

PR #490, opened 2025-12-16
make `image::fetch()` implicitly declare lod 0

Firestar99:
* glsl [`texelFetch(sampler, coordinate, lod)`](https://docs.gl/sl4/texelFetch) takes the lod as an explicit parameter, so you can't emit a texel fetch in glsl without explicit lod
* naga doesn't understand texel fetch without explicit lod

### DIFF
--- a/crates/spirv-std/src/image.rs
+++ b/crates/spirv-std/src/image.rs
@@ -144,21 +144,7 @@ impl<
     where
         I: Integer,
     {
-        let mut result = SampledType::Vec4::default();
-        unsafe {
-            asm! {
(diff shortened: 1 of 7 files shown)


LegNeato:
@schell you know more graphics stuff, can you review this?

Firestar99:
I don't have a good compiletest to offer, other than naga chokes on `image.fetch()` but works with `image.fetch_with(lod(0))` or our alias `image.fetch_with_lod(0)`. Also see this past PR https://github.com/Rust-GPU/rust-gpu/pull/361

PR #493, opened 2025-12-17
Add `naga-wgsl` target v2

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/491
Supersedes https://github.com/Rust-GPU/rust-gpu/pull/280

Adds `spirv-unknown-naga-wgsl` target for built-in naga transpilation to wgsl. Further naga targets could be added easily.

Quite a lot compiletest fail with naga, common reasons include:
* unsupported extensions & capabilities (mesh shader, ray tracing etc.)
* no support for `u8` & `i8`
* unknown instructions
* > Vertex shaders must return a `@builtin(position)` output value

For now, I'd just accept that any naga targets won't pass our compiletest suite and not add a CI step for that.

### DIFF
--- a/crates/rustc_codegen_spirv/Cargo.toml
+++ b/crates/rustc_codegen_spirv/Cargo.toml
@@ -20,10 +20,10 @@ crate-type = ["dylib"]
 default = ["use-compiled-tools"]
 # If enabled, uses spirv-tools binaries installed in PATH, instead of
 # compiling and linking the spirv-tools C++ code
-use-installed-tools = ["spirv-tools/use-installed-tools"]
+use-installed-tools = ["spirv-tools/use-installed-tools", "naga"]
 # If enabled will compile and link the C++ code for the spirv tools, the compiled
 # version is preferred if both this and `use-installed-tools` are enabled
-use-compiled-tools = ["spirv-tools/use-compiled-tools"]
+use-compiled-tools = ["spirv-tools/use-compiled-tools", "naga"]
 # If enabled, this will not check whether the current rustc version is set to the
 # appropriate channel. rustc_cogeden_spirv requires a specific nightly version,
 # and will likely produce compile errors when built against a different toolchain.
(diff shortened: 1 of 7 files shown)


crates/rustc_codegen_spirv/Cargo.toml
```diff
@@ -20,15 +20,16 @@ crate-type = ["dylib"]
 default = ["use-compiled-tools"]
 # If enabled, uses spirv-tools binaries installed in PATH, instead of
 # compiling and linking the spirv-tools C++ code
-use-installed-tools = ["spirv-tools/use-installed-tools"]
+use-installed-tools = ["spirv-tools/use-installed-tools", "naga"]
 # If enabled will compile and link the C++ code for the spirv tools, the compiled
 # version is preferred if both this and `use-installed-tools` are enabled
-use-compiled-tools = ["spirv-tools/use-compiled-tools"]
+use-compiled-tools = ["spirv-tools/use-compiled-tools", "naga"]
```
nazar-pc, line 26: Hm, this effectively makes `naga` non-optional if I'm reading it correctly. Is it intentional?

Firestar99: That's from the old PR, added with comment
> [wgsl: enable naga feature by default, cargo-gpu can't handle it](https://github.com/Rust-GPU/rust-gpu/pull/493/commits/9fe446cf0fdf72842b580f88db353689b15da8f6)

So this was more a hack than anything else. I'll have a look what the compile time impact is, and if it's actually significant (which next to `spirv-tools-sys` may not be), look into making it optional and supported by cargo-gpu.

nazar-pc: I'm wondering if `naga` should really be a dependency of either of those two features. Why not let user specify it explicitly instead?

Firestar99: > cargo-gpu can't handle it

we'd need to build the infra in cargo-gpu first

Firestar99:
* upgraded naga 27 -> 29
* remove CI checking compiletests with naga target, too many failures due to transpilation to bother atm

eddyb (approved):

PR #495, opened 2025-12-19
wgpu 27

Firestar99:
* **most of the diff is lockfiles** from `cargo update`
* update to wgpu 27
  * required for https://github.com/Rust-GPU/rust-gpu/pull/493, it seems naga 27 and wgpu 26 are biting themselves if they're in your dep tree
* wgpu 28 just released, but requires rustc 1.92 (current stable), which we don't support yet
  * (cargo gpu does through decoupling the CPU and shader toolchains)


LegNeato (approved):
Looks reasonable, thanks!

PR #497, opened 2025-12-22
Spirv builder cleanup

Firestar99:
# Requires https://github.com/Rust-GPU/rust-gpu/pull/484 https://github.com/Rust-GPU/rust-gpu/pull/488 https://github.com/Rust-GPU/rust-gpu/pull/496

* review commit by commit after the merge commits
* just some general cleanup
* were originally created for https://github.com/Rust-GPU/rust-gpu/pull/496, but replaced it with a far simpler implementation, but these cleanups are still good

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -758,24 +758,22 @@ impl SpirvBuilder {
         let metadata: CompileResult =
             rustc_codegen_spirv_types::serde_json::from_reader(BufReader::new(metadata_contents))
                 .map_err(SpirvBuilderError::MetadataFileMalformed)?;
-        match &metadata.module {
-            ModuleResult::SingleModule(spirv_module) => {
-                assert!(!self.multimodule);
-                if self.build_script.get_env_shader_spv_path() {
-                    let env_var = format!(
-                        "{}.spv",
-                        at.file_name()
-                            .unwrap()
-                            .to_str()
-                            .unwrap()
-                            .strip_suffix(ARTIFACT_SUFFIX)
-                            .unwrap()
-                    );
+
+        let is_multimodule = match &metadata.module {
+            ModuleResult::SingleModule(_) => false,
+            ModuleResult::MultiModule(_) => true,
+        };
+        assert_eq!(self.multimodule, is_multimodule);
+
+        if self.build_script.get_env_shader_spv_path() {
+            match &metadata.module {
+                ModuleResult::SingleModule(spirv_module) => {
+                    let env_var = spirv_module.file_name().unwrap().to_str().unwrap();
                     println!("cargo::rustc-env={}={}", env_var, spirv_module.display());
                 }
-            }
-            ModuleResult::MultiModule(_) => {
-                assert!(self.multimodule);
+                ModuleResult::MultiModule(_) => {
+                    Err(SpirvBuilderError::MultiModuleWithEnvShaderSpvPath)?;
+                }
             }
         }
         Ok(metadata)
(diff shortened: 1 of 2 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -117,9 +117,9 @@ pub enum SpirvBuilderError {
         "`multimodule: true` build cannot be used together with `build_script.env_shader_spv_path: true`"
     )]
     MultiModuleWithEnvShaderSpvPath,
-    #[error("multi-module metadata file missing")]
+    #[error("some metadata file is missing")]
     MetadataFileMissing(#[from] std::io::Error),
-    #[error("unable to parse multi-module metadata file")]
+    #[error("unable to parse some metadata file")]
```
LegNeato: It would be better to have two variants here for the single and multi-module case to make the message more explicit and the message better?

LegNeato (changes requested):

crates/spirv-builder/src/lib.rs
Firestar99: The error message on main is completely wrong: There are no "multi-module" metadata files. The metadata file is needed regardless of the module being a multi-module or single module, as it contains all entry point names and `*.spv` paths which are returned to you after building.

In fact there are four cases `MetadataFileMissing` will be emitted *on main*:
* `parse_metadata_file`: The metadata file is missing, stays `MetadataFileMissing`. Can also fail with `MetadataFileMalformed`.
* `leaf_deps` fails to read rustc dependency information files, now `DepFileParseError`. 
* `query_rustc_version` fails to query the rustc version, now generic `IoError`.
* `TargetSpecVersion::target_arg` fails to write the target spec file, now generic `IoError`.

The new error messages:
https://github.com/Rust-GPU/rust-gpu/blob/a292dc3312d280b82e790776706e9791c8db7703/crates/spirv-builder/src/lib.rs#L120-L125
https://github.com/Rust-GPU/rust-gpu/blob/a292dc3312d280b82e790776706e9791c8db7703/crates/spirv-builder/src/lib.rs#L135-L136

crates/spirv-builder/src/lib.rs
```diff
@@ -888,297 +883,309 @@ fn join_checking_for_separators(strings: Vec<impl Borrow<str>>, sep: &str) -> St
...
+                toolchain_rustc_version.clone()
+            } else {
+                query_rustc_version(self.toolchain_overwrite.as_deref())?
+            };
+
+        // Okay, this is a little bonkers: in a normal world, we'd have the user clone
+        // rustc_codegen_spirv and pass in the path to it, and then we'd invoke cargo to build it, grab
+        // the resulting .so, and pass it into -Z codegen-backend. But that's really gross: the user
+        // needs to clone rustc_codegen_spirv and tell us its path! So instead, we *directly reference
+        // rustc_codegen_spirv in spirv-self's Cargo.toml*, which means that it will get built
```
eddyb:
```suggestion
        // rustc_codegen_spirv in spirv-builder's Cargo.toml*, which means that it will get built
```

crates/spirv-builder/src/lib.rs
```diff
@@ -888,297 +883,309 @@ fn join_checking_for_separators(strings: Vec<impl Borrow<str>>, sep: &str) -> St
...
-        // we're emitting regular rust libraries as is expected. For `dylib` or `cdylib`, we're linking all `rlib`s
-        // together, legalize them in many passes and emit a final `*.spv` file. Quirk: If you depend on a crate
-        // that has crate-type `dylib`, we also link it, and it will fail if it has no shaders, which may not be
-        // desired. (Gathered from reading source code and experimenting, @firestar99)
-        cargo.args(["--crate-type", "dylib"]);
-    }
+        let target_dir_path = self
+            .target_dir_path
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("spirv-self"));
```
eddyb:
```suggestion
            .unwrap_or_else(|| PathBuf::from("spirv-builder"));
```

eddyb (changes requested):

Firestar99:
Oops, find replace was a little too aggressive there... squashed it in

eddyb (approved):

Firestar99:
Force-merged since review from LegNeato is blocking it

Issue #501, opened 2025-12-30
u64 with `trailing_zeros` returns incorrect values


voxelbee:
### Expected Behaviour

We expect the `trailing_zeros()` function on u64 values to match the behavior of Rust's standard library implementation, returning the count of trailing zero bits.

For example:

- `0x0000000000000001_u64.trailing_zeros()` should return 0
- `0x0000000100000000_u64.trailing_zeros()` should return 32
- `0x8000000000000000_u64.trailing_zeros()` should return 63

### Example & Steps To Reproduce

- `0x0000000000000001` Expected: 0, Result: 32
- `0x0000000100000000` Expected: 32, Result: 0
- `0x8000000000000000` Expected: 63, Result: 31

Run a shader that calculates trailing zeros on a u64. I am creating a pull request that contains test to validate the function and a fix for this issue.

```rust
#![no_std]
use spirv_std::spirv;

#[spirv(compute(threads(1)))]
pub fn main_cs(
    #[spirv(storage_buffer, descriptor_set = 0, binding = 0)] output: &mut [u32],
) {
    // Expected: 0, Actual: 32
    output[0] = 0x0000000000000001_u64.trailing_zeros();
    
    // Expected: 32, Actual: 0  
    output[1] = 0x0000000100000000_u64.trailing_zeros();
}
```

PR #502, opened 2025-12-30
Added tests for 64bit trailing_zeros

voxelbee:
Fix for issue #501

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/intrinsics.rs
+++ b/crates/rustc_codegen_spirv/src/builder/intrinsics.rs
@@ -582,8 +582,8 @@ impl Builder<'_, '_> {
 
                         if trailing {
                             let use_lower = self.emit().i_equal(bool, None, lower, u32_0).unwrap();
-                            let lower_bits = find_xsb(self, lower, 32);
-                            let higher_bits = find_xsb(self, higher, 0);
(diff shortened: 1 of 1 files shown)


Firestar99:
1. Looks like swiftshader doesn't support int64... I was looking into [switching to lavapipe](https://github.com/Rust-GPU/rust-gpu/pull/489) anyway for unrelated reasons, guess I got to finish that one.

2. wgpu has a native-only feature called [`SHADER_INT64`](https://docs.rs/wgpu/latest/wgpu/struct.Features.html#associatedconstant.SHADER_INT64), so you don't necessarily need to use the ash runner

3. I wouldn't be surprised if there's more integer intrinsics that have issues. These in particular have been added before we had difftests, and I wanted to some for quite a while already.

Firestar99:
I just rebased it and split out the fix from the test.

I started working on a difftest that would test all the various functions with wgpu, only to notice that our wgpu difftest runners are a mess and I need to clean them up first before we can use wgpu for 64bit testing. So I'm fine with keeping the ash tests as is, and I may clean them up later. (There's still https://github.com/Rust-GPU/rust-gpu/pull/334 waiting for a review, which should speed up difftests significantly...)

Anyway, I think it's better if we land it now and take care of testing the remaining ops some other time. I may even have found an alternative way of testing them even better. Thanks for the patch :heart:

PR #505, opened 2026-01-05
spirv-builder: check and clippy shortcuts

Firestar99:
@nazar-pc ashed in https://github.com/Rust-GPU/cargo-gpu/pull/131#issuecomment-3691984086:

> I'd appreciate some simple [to run `cargo gpu clippy` and `check`] way through `cargo-gpu-install`.

So I added these shortcuts, so you can run clippy in your build script, and afterwards still build your shaders without having to clone the `SpirvBuilder`.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -719,6 +719,24 @@ impl SpirvBuilder {
         self
     }
 
+    /// Shortcut for `cargo check`
+    pub fn check(&mut self) -> Result<CompileResult, SpirvBuilderError> {
+        self.run_cargo_cmd("check")
+    }
+
+    /// Shortcut for `cargo clippy`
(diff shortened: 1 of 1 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -675,6 +675,24 @@ impl SpirvBuilder {
         self
     }
 
+    /// Shortcut for `cargo check`
+    pub fn check(&mut self) -> Result<CompileResult, SpirvBuilderError> {
+        self.run_cargo_cmd("check")
+    }
+
+    /// Shortcut for `cargo clippy`
+    pub fn clippy(&mut self) -> Result<CompileResult, SpirvBuilderError> {
```
Firestar99, line 728: I'm not a huge fan how these are `&mut self` but `build()` is only `&self`. But I also don't want to clone the entire `SpirvBuilder` just to change that one property, though it probably wouldn't matter performance-wise. Suggestions?

PR #460, opened 2025-11-06
Update toolchain to latest nightly (Nov 13)

dvdplm:
Based on @nnethercote's work in #420, but with a more recent toolchain.

The toolchain update from `nightly-2025-06-30` to `nightly-2025-11-13` uncovered a slew of minor incompatibilities and a major(?) ABI issue: `rustc_target::callconv::PassMode` handling in function ABIs changed upstream. The bulk of that work happened in [#144529](https://github.com/rust-lang/rust/pull/144529) with the introduction of the new internal `#[rustc_pass_indirectly_in_non_rustic_abis]` attribute.

This change causes an ICE when we run the compiletests from `rust-gpu` with, e.g. `cargo run -p compiletests --no-default-features --features "use-installed-tools" typed_buffer.rs -- --target-env vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4,spv1.3`. The assertion we hit when compiling `core` is [here](https://github.com/rust-lang/rust/blob/0b329f801a09004dacb19aaf09d5cb8b4c51d3f8/compiler/rustc_ty_utils/src/abi.rs#L392-L397):

```
            assert_matches!(
                arg.mode,
                PassMode::Indirect { on_stack: false, .. },
                "the {spec_abi} ABI does not implement `#[rustc_pass_indirectly_in_non_rustic_abis]`"
            );
```

The assertion fires in our query hook (abi.rs:104) when calling the default provider's `fn_abi_of_instance`, which now runs the new sanity check from PR #144529. `rust-gpu` converts `extern "C"` to `extern "unadjusted"` (line 47 in abi.rs), then attempts to force all arguments to `PassMode::Direct` via `readjust_fn_abi()` (line 68). However, PR #144529 introduced a new invariant that requires certain types marked with `#[rustc_pass_indirectly_in_non_rustic_abis]` to be passed indirectly, even in unadjusted ABIs.

<details>
<summary>Example ICE output (Click to view)</summary>

```
$ cargo run -p compiletests --no-default-features --features "use-installed-tools" typed_buffer.rs -- --target-env vulkan1.1,vulkan1.2,vulkan1.3,vulkan1.4,spv1.3
   Compiling rustc_codegen_spirv v0.9.0 (/Users/aggron/dev/play/rust-gpu/crates/rustc_codegen_spirv)
   Compiling compiletests v0.0.0 (/Users/aggron/dev/play/rust-gpu/tests/compiletests)
[...]

PR #508, opened 2026-01-08
Update to nightly-2025-08-04.

fee1-dead:
based on #420

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -85,8 +85,6 @@ extern crate rustc_arena;
 #[cfg(rustc_codegen_spirv_disable_pqp_cg_ssa)]
 extern crate rustc_ast;
 #[cfg(rustc_codegen_spirv_disable_pqp_cg_ssa)]
-extern crate rustc_attr_data_structures;
-#[cfg(rustc_codegen_spirv_disable_pqp_cg_ssa)]
 extern crate rustc_attr_parsing;
 #[cfg(rustc_codegen_spirv_disable_pqp_cg_ssa)]
 extern crate rustc_codegen_ssa;


fee1-dead:
The `core` ICE seems to be that the provided immediate has type `fn(f32) -> i8` while `from_scalar` has a type of `*void`.

fee1-dead:
Apparently it was codegenning for https://doc.rust-lang.org/src/core/ptr/const_ptr.rs.html#153

EDIT: more precisely, `<*const u8>::addr`

Firestar99:
@fee1-dead I'll review it in a moment, but I've got a question for you: This PR is based on https://github.com/Rust-GPU/rust-gpu/pull/420, but we've also had someone else attempt to upgrade the toolchain to 2025-11-13 in https://github.com/Rust-GPU/rust-gpu/pull/460. Now their branch doesn't work as of right now, and their commit list somehow reverts some merged PRs for whatever reason. But still, which one should we go forward with? 
Also bjorn3 added some comments that may help you out as well?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -3389,6 +3384,15 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
         }
 
+        // HACK(fee1-dead): `MaybeUninit` uses a union which we don't have very good support yet. Replacing all calls to it
+        // with an `Undef` serves the same purpose and fixes compiler errors
+        if instance_def_id.is_some_and(|did| {
+            self.tcx
+                .is_diagnostic_item(rustc_span::sym::maybe_uninit_uninit, did)
+        }) {
+            return self.undef(result_type);
+        }
```
Firestar99, line 3394: What an interesting solution, I like it!

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -3397,6 +3401,19 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+        _llty: Self::Type,
+        _fn_attrs: Option<&CodegenFnAttrs>,
+        _fn_abi: &FnAbi<'tcx, Ty<'tcx>>,
+        _llfn: Self::Value,
+        _args: &[Self::Value],
+        _funclet: Option<&Self::Funclet>,
+        _instance: Option<ty::Instance<'tcx>>,
+    ) {
+        todo!()
+    }
```
Firestar99, line 3415: This is from @nnethercote 's commits. CI is passing, so I assume this is never called, but why not make `tail_call` call `call`?

Firestar99 (approved):
Looks good to me, then again, I'm not a rustc guy. I just see CI passing and nothing in the changes that jumps out to me. Just one question :D

About https://github.com/Rust-GPU/rust-gpu/pull/460: it looks like @dvdplm jumped immediately to 2025-11-02 and later to 2025-11-13, so we should be able to rebase their work on top of this and work from there. (I'll try to rebase it rn)

Firestar99:
I've rebased https://github.com/Rust-GPU/rust-gpu/pull/460 onto this PR and split it into:
* working https://github.com/Rust-GPU/rust-gpu/pull/541
* still needs work: https://github.com/Rust-GPU/rust-gpu/pull/542

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
LegNeato: Could that break semantic expectations when someone expects TCO? I guess it is moot as there is no support for `become` or whatever yet...

crates/rustc_codegen_spirv/build.rs
```diff
@@ -237,6 +237,14 @@ pub(super) fn elf_e_flags(architecture: Architecture, sess: &Session) -> u32 {",
                 src = src.replace("alloca(field.size,", "typed_alloca(llfield_ty,");
             }
 
+            // HACK(fee1-dead): our backend type number doesn't always match the type of the value. Should fix?
+            if relative_path == Path::new("src/mir/rvalue.rs") {
+                src = src.replace(
+                    "debug_assert_eq!(bx.cx().val_ty(imm), from_backend_ty);",
+                    "",
+                );
+            }
```
Firestar99, line 246: We've discussed this change in a meeting and figured out it's cause. The debug assert got added late Juli (between [2025-07-28](https://github.com/Rust-GPU/rust-gpu/pull/508/changes/9c61be21b2bd813e690f95a2ffd5210c6f16a447) and [2025-08-04](https://github.com/Rust-GPU/rust-gpu/pull/508/changes/5796ebceb2901c796bdd827b07d93f65b7470afc)). It's failing on function pointers specifically, as it asserts that `val_ty` returning a Word of type `SpirvType::Pointer` -> `SpirvType::Function { .. }` is equal to [`type_ptr_ext()`](https://github.com/Rust-GPU/rust-gpu/blob/7f5fd7cb611d8a54f2497397134ccf7764124eea/crates/rustc_codegen_spirv/src/codegen_cx/type_.rs#L216-L218) (after some indirection), which has a dummy impl that returns `SpirvType::Pointer` -> `SpirvType::Void`. Those are ofc not equal.

PR #512, opened 2026-01-09
Fold OpLoad from Private variables with constant initializers

LegNeato:
Disclosure: I had AI write this and I only superficially understand the code.

Replace OpLoad from Private/Function
storage class variables with their constant initializers. This eliminates pointer-to-pointer patterns like `&&123` which would otherwise generate invalid SPIR-V in Logical addressing mode (pointer-to-pointer is only allowed for StorageBuffer/Workgroup, not Private).

The optimization runs after inlining (which may expose such patterns) and before DCE (so the now-unused variables can be removed).

Also extend DCE to treat Private storage class variables as pure, allowing them to be removed when unused.

Newer versions of spirv-val catch this, causing our compiletests to fail.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/dce.rs
+++ b/crates/rustc_codegen_spirv/src/linker/dce.rs
@@ -282,7 +282,15 @@ fn instruction_is_pure(inst: &Instruction) -> bool {
         | PtrEqual
         | PtrNotEqual
         | PtrDiff => true,
-        Variable => inst.operands.first() == Some(&Operand::StorageClass(StorageClass::Function)),
+        // Variables with Function or Private storage class are pure and can be DCE'd if unused.
+        // Other storage classes (Input, Output, Uniform, etc.) are part of the shader interface
+        // and must be kept.
+        Variable => matches!(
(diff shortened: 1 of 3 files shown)


Firestar99 (approved):
Code looks good!

And indeed spirv-tools v2025.5 (pulled from nix unstable) introduces a new validation rule that is failing `nested-ref`:
```
error: error:0:0 - In Logical addressing, variables can only allocate a pointer to the StorageBuffer or Workgroup storage classes
         %3 = OpVariable %_ptr_Private__ptr_Private_uint Private %4
   |
   = note: spirv-val failed
   = note: module `$TEST_BUILD_DIR/lang/consts/nested-ref.vulkan1.2`
```

Firestar99:
For a moment I was considering if this transformation may be unsound. There could be a situation where an `OpVariable` has some initial contents, but in the code something `OpStore`'s some data to that variable before it is being loaded again. Now rust-gpu itself only emits `OpVariables` for entry points and constants, which won't ever have anything stored to them, but I'm not sure if maybe spirt transforms code into `OpStore`s. @eddyb ?

Issue #293, opened 2025-06-25
"No OpEntryPoint instruction was found" for public function


nazar-pc:
### Expected Behaviour

Non-public submodules should be supported, but instead I'm getting this:
```
     Compiling ab-proof-of-space-gpu v0.1.0 (ab-proof-of-space-gpu)
  warning: function `main_cs` is never used
    --> src/shaders.rs:10:8
     |
  10 | pub fn main_cs(
     |        ^^^^^^^
     |
     = note: `#[warn(dead_code)]` on by default

  error: No OpEntryPoint instruction was found. This is only allowed if the Linkage capability is being used.
    |
    = note: module `target/debug/build/ab-proof-of-space-gpu-dc89fe269d0c81ad/out/spirv-unknown-vulkan1.1/release/deps/ab_proof_of_space_gpu.spv`

  warning: an unknown error occurred
    |
    = note: spirv-opt failed, leaving as unoptimized
    = note: module `target/debug/build/ab-proof-of-space-gpu-dc89fe269d0c81ad/out/spirv-unknown-vulkan1.1/release/deps/ab_proof_of_space_gpu.spv`

  error: error:0:0 - No OpEntryPoint instruction was found. This is only allowed if the Linkage capability is being used.
    |
    = note: spirv-val failed
    = note: module `target/debug/build/ab-proof-of-space-gpu-dc89fe269d0c81ad/out/spirv-unknown-vulkan1.1/release/deps/ab_proof_of_space_gpu.spv`
```

Slapping `#[unsafe(no_mangle)]` is sufficient to bypass this error, making submodule publicly reachable too.

I think `#[spirv(..)]` macro should be able to do something about it automatically to ensure it is exposed properly.

### Example & Steps To Reproduce

1. Have a shader in submodule that is not publicly reachable

Firestar99:
This is a known issue that entry points must be reachable via a `pub` chain. I think this has to do with rust-gpu only emitting entry points which are [monomorphized](https://rustc-dev-guide.rust-lang.org/backend/monomorph.html), and rustc only monomorphizing symbols that are considered reachable, which a non-generic publicly reachable function is. And `no_mangle` seem to imply that too.

LegNeato:
Related but not exactly the same: https://github.com/Rust-GPU/rust-gpu/issues/155

Firestar99:
[...]

PR #514, opened 2026-01-15
allow ZST entry params in input and output positions

Firestar99:
With ZST entry params, you can do some cursed generic entry point stuff to make certain inputs and outputs optional:

https://github.com/Rust-GPU/rust-gpu/blob/5f1d97fc624d7ccf919a12d44a04c3fca1135091/tests/compiletests/ui/spirv-attr/entry_generic.rs#L19-L28

### DIFF
--- a/tests/compiletests/ui/spirv-attr/entry_generic.rs
+++ b/tests/compiletests/ui/spirv-attr/entry_generic.rs
@@ -1,0 +1,28 @@
+// build-pass
+// compile-flags: -C llvm-args=--disassemble
+// normalize-stderr-test "OpSource .*\n" -> ""
+// normalize-stderr-test "OpLine .*\n" -> ""
+// normalize-stderr-test "%\d+ = OpString .*\n" -> ""
+// normalize-stderr-test "(^|\n); .*" -> ""
+// normalize-stderr-test "OpCapability VulkanMemoryModel\n" -> ""
+// normalize-stderr-test "OpMemoryModel Logical Vulkan" -> "OpMemoryModel Logical Simple"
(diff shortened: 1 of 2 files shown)


Ralith:
As a prospective user, I think having entry points appear as a side-effect of monomorphization is potentially very difficult to reason about. I'd rather just write out the entry points I want.

LegNeato:
Oh this is fun

Firestar99:
> I think having entry points appear as a side-effect of monomorphization is potentially very difficult to reason about. I'd rather just write out the entry points I want.

I can get behind that monomorpization generating entry points is very cheesy, but I wouldn't outright ban people from making generic entry points. Afterall, it's an active choice you make as a developer to declare your entry point generic, and you can always resort to macros to generate all the variants you want.

Using generics on entry points has always been possible, even if I don't know anyone using it atm. This PR *only* allows entry point params that are ZST to be ignored instead of erroring, which would even help both with generic and macro case by you not having to explicitly remove args but being able to just ZST them away.

A bigger issue with monomorpization is that rustc doesn't monomorphize functions that are not "exported" aka visible from an outside crate. So the entry point must be pub, all modules leading to it must be pub, or you need to pub reexport it somehow. Which is also why `fn dummy()` must be `pub` in the example. See https://github.com/Rust-GPU/rust-gpu/issues/293 and https://github.com/Rust-GPU/rust-gpu/issues/155

Ralith:
> Using generics on entry points has always been possible

Fair, I guess the discussion is out of scope for this PR then.

LegNeato:
@eddyb see any issues with this?

Issue #516, opened 2026-01-20
powi is slow


voidentente:
# Premise

I was wondering whether I could replace expressions like `x * x` with `x.powi(2)` and went ahead and compared the generated SPIR-V in both cases. Turns out, Rust-GPU fails to optimize even the most obvious opportunities.

# Method

I inspected the SPIR-V using https://www.khronos.org/spirv/visualizer/ and the artifact was generated in release mode using rust-gpu 08f98a7335ae7e2ac4ec9423a9b1acf9270f2a65 (a commit from Mon Jan 5 12:45:27 2026 +0100), which uses Rust nightly-2025-06-30.

# Case: Manual Multiply

This Rust code:

```rs
#[spirv(fragment)]
pub fn fragment(out: &mut Vec4, #[spirv(frag_coord)] frag_coord: Vec4) {
    *out = Vec4::splat(frag_coord.x * frag_coord.x);
}
```

Becomes the following SPIR-V:

```
OpCapability Shader
OpCapability VulkanMemoryModel
OpMemoryModel Logical Vulkan
OpEntryPoint Fragment %1 "fragment" %2 %3
OpExecutionMode %1 OriginUpperLeft
OpDecorate %2 BuiltIn FragCoord
OpDecorate %3 Location 0
%6 = OpTypeFloat 32
%7 = OpTypeVector %6 4
%8 = OpTypePointer Output %7
%9 = OpTypePointer Input %7
%10 = OpTypeVoid
%11 = OpTypeFunction %10
%2 = OpVariable %9 Input
%3 = OpVariable %8 Output
%1 = OpFunction %10 None %11
%12 = OpLabel
%13 = OpLoad %7 %2
%14 = OpCompositeExtract %6 %13 0
%16 = OpFMul %6 %14 %14
%17 = OpCompositeConstruct %7 %16 %16 %16 %16
OpStore %3 %17
OpReturn
OpFunctionEnd
```

# Case: Powi

The following Rust code:

```rs
#[spirv(fragment)]
pub fn fragment(out: &mut Vec4, #[spirv(frag_coord)] frag_coord: Vec4) {
    *out = Vec4::splat(frag_coord.x.powi(2));
}
```

Becomes the following SPIR-V:

```
OpCapability Shader
OpCapability VulkanMemoryModel
OpMemoryModel Logical Vulkan
OpEntryPoint Fragment %1 "fragment" %2 %3
OpExecutionMode %1 OriginUpperLeft
[...]

PR #518, opened 2026-01-21
Optimize num_traits::Float::powi to use GLSL.std.450 Pow

LegNeato:
Fixes https://github.com/Rust-GPU/rust-gpu/issues/516.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/libm_intrinsics.rs
+++ b/crates/rustc_codegen_spirv/src/builder/libm_intrinsics.rs
@@ -43,7 +43,7 @@ pub enum LibmIntrinsic {
     Custom(LibmCustomIntrinsic),
 }
 
-pub const TABLE: &[(&str, LibmIntrinsic)] = &[
+pub const LIBM_TABLE: &[(&str, LibmIntrinsic)] = &[
     ("acos", LibmIntrinsic::GLOp(GLOp::Acos)),
     ("acosf", LibmIntrinsic::GLOp(GLOp::Acos)),
     ("acosh", LibmIntrinsic::GLOp(GLOp::Acosh)),
@@ -158,7 +158,6 @@ pub const TABLE: &[(&str, LibmIntrinsic)] = &[
     ),
     ("pow", LibmIntrinsic::GLOp(GLOp::Pow)),
     ("powf", LibmIntrinsic::GLOp(GLOp::Pow)),
-    ("powi", LibmIntrinsic::Custom(LibmCustomIntrinsic::Powi)),
     (
         "remainder",
         LibmIntrinsic::Custom(LibmCustomIntrinsic::Remainder),
@@ -199,6 +198,9 @@ pub const TABLE: &[(&str, LibmIntrinsic)] = &[
     ("trunc", LibmIntrinsic::GLOp(GLOp::Trunc)),
     ("truncf", LibmIntrinsic::GLOp(GLOp::Trunc)),
 ];
+
+pub const NUM_TRAITS_TABLE: &[(&str, LibmIntrinsic)] =
+    &[("powi", LibmIntrinsic::Custom(LibmCustomIntrinsic::Powi))];
 
 impl Builder<'_, '_> {
     pub fn call_libm_intrinsic(
--- a/crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
@@ -169,8 +169,10 @@ impl<'tcx> CodegenCx<'tcx> {
         // Check for usage of `num_traits` intrinsics (like Float::powi) that we can optimize
         if self.tcx.crate_name(def_id.krate) == self.sym.num_traits && !def_id.is_local() {
             let item_name = self.tcx.item_name(def_id);
-            if let Some(&intrinsic) = self.sym.libm_intrinsics.get(&item_name) {
-                self.libm_intrinsics.borrow_mut().insert(def_id, intrinsic);
+            if let Some(&intrinsic) = self.sym.num_traits_intrinsics.get(&item_name) {
+                self.num_traits_intrinsics
+                    .borrow_mut()
+                    .insert(def_id, intrinsic);
             }
         }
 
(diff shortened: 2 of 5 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/declare.rs
```diff
@@ -166,6 +166,14 @@ impl<'tcx> CodegenCx<'tcx> {
             }
         }
 
+        // Check for usage of `num_traits` intrinsics (like Float::powi) that we can optimize
+        if self.tcx.crate_name(def_id.krate) == self.sym.num_traits && !def_id.is_local() {
+            let item_name = self.tcx.item_name(def_id);
+            if let Some(&intrinsic) = self.sym.libm_intrinsics.get(&item_name) {
+                self.libm_intrinsics.borrow_mut().insert(def_id, intrinsic);
+            }
+        }
```
nazar-pc, line 177: I have no idea how things work more generally, but why are things like this even necessary? I'd expect the generic machinery of the compiler to optimize it to some canonical form that the target would like the most.

So it shouldn't make any difference if one writes `x.powi(2)` or `x * x` or uses a few layers of zero-cost abstractions in the process, should all result in identical SPIR-V, just like it does with LLVM. What am I missing?

LegNeato: The problem is we are using `num_traits` to get the APIs we need off primitives, and `num_traits` knows nothing about the intrinsics available on the platform. It is using a loop as impl https://github.com/rust-num/num-traits/blob/master/src/pow.rs#L179. Pretty much no compiler would be able to tell that code can be replaced with a cast and a `powi`, and `rustc` doesn't. Instead somewhere needs to map the call / the semantics to the underlying fast impl, which is what the backend is _supposed_ to do. It already does it for many ops, we just missed a spot (I guess? This all predates me).

One can argue we shouldn't use `num_traits` and instead should just hang our own impls off the types, I have no clue why it was chosen vs that route. Probably a make it work first then make it fast tradeoff (as the `num_traits` impl _does_ work on spirv).

I have a new optimizer where we can control things much more rather than relying on `spirv-opt`, but this replacement happens _before_ the optimizer.

nazar-pc: Ouch, that definitely looks like a problem. Maybe worth creating an issue to look into getting rid of `num_traits`? Doing things like looping instead of native exponentiation is definitely going to hurt and there might be more examples like it.

Firestar99: We can't get rid of `num_traits` for silly rust reasons: 
[`core::f32::powf`](https://doc.rust-lang.org/std/primitive.f32.html#method.powi) is marked as `#[rustc_allow_incoherent_impl]`, meaning it doesn't actually "exist" unless some other crate declares it. I'm not too familiar on the details, but it seems like when you have `std` it exists and if you only have core it's declared but doesn't exist and using it will fail with:
```rust
  error[E0599]: no function or associated item named `powf` found for type `f32` in the current scope
     --> crates/restir-shader/src/material/pbr/eval.rs:159:25
      |
  159 |     f0 + (1.0 - f0) * f32::powf(f32::clamp(1.0 - cos_theta, 0.0, 1.0), 5.0)
      |                            ^^^^ function or associated item not found in `f32`
      |
      = help: items from traits can only be used if the trait is in scope
```
see https://github.com/rust-lang/rust/issues/149347 on this sillyness

The workaround has usually been this `use` statement:
```rust
#[cfg(target_arch = "spirv")]
use spirv_std::num_traits::Float;
```

`num_traits` with `std` feature will call `std` functions and with `libm` feature / in `#[no_std]` environments call the appropriate `libm` function. For SPIR-V, our [`spirv-std` configures the feature for you](https://github.com/Rust-GPU/rust-gpu/blob/af29a38586e3000a6699dcd394c81c7fb9091ba3/crates/spirv-std/Cargo.toml#L19-L26) and we then intercept those libm calls and replace them with intrinsics.

While researching this, I've noticed this has been discussed [in 2021 on the embark repo already](https://github.com/EmbarkStudios/rust-gpu/issues/795), and their conclusion was:
> I think this is a fully external issue, then - rust-gpu is working "as intended", faithfully compiling the code that it is given, which doesn't use an intrinsic. The way to fix this, then, would be to change upstream code to be less branchy (e.g. requesting that libm adds a powi intrinsic, and having num-traits call it).

I don't mind if we want to merge this anyway, so it's not a trap to end users @LegNeato. 

But `num-traits` isn't a perfect solution either: *Technically,* it doesn't actually reimplement all f32 functions. For example, it excludes `euclid`, for which you need to use the `Euclid` trait, [which takes args by reference instead of by value](https://github.com/rust-num/num-traits/issues/349), so you'll get [awful code like this](https://github.com/GraphiteEditor/Graphite/blob/2be7790d4daa2b46b4f94c7064ef7cc208756f32/node-graph/libraries/no-std-types/src/color/color_types.rs#L452-L455) (which initially sparked my investigation into this last summer):
```rust
#[cfg(feature = "std")]
let rem = |x: f32| x.rem_euclid(1.);
#[cfg(not(feature = "std"))]
let rem = |x: f32| x.rem_euclid(&1.);
```

LegNeato: @Firestar99 created an issue for this, let's just merge for now as the current state is obviously bad and this makes it (slightly) better?

LegNeato:
@Firestar99 ping

crates/rustc_codegen_spirv/src/builder/libm_intrinsics.rs
```diff
@@ -306,6 +310,12 @@ impl Builder<'_, '_> {
             LibmIntrinsic::Custom(LibmCustomIntrinsic::NextAfter) => {
                 self.undef_zombie(result_type, "NextAfter not supported yet")
             }
+            LibmIntrinsic::Custom(LibmCustomIntrinsic::Powi) => {
+                assert_eq!(args.len(), 2);
+                // Convert integer exponent to float, then use GLOp::Pow
+                let float_exp = self.sitofp(args[1], args[0].ty);
+                self.gl_op(GLOp::Pow, result_type, [args[0], float_exp])
+            }
```
foxtran, line 318: I think for known low powers it would be better to produce explicit sequence of multiplications (or division + sequence of multiplications for negate powers)

Firestar99: While generally correct, SPIR-V isn't final machine code. Most drivers will spin up an entire LLVM compiler to turn that SPIR-V into whatever machine code they actually run, which are going to be much better at emitting properly optimized Pow. So the best thing we can do is tell them directly it's a Pow, and not needlessly complicating it.

Firestar99:
@LegNeato since you put a :+1: on my comment, I assume you're fine with my changes but forgot to merge? I'll just hit the merge button...

Issue #522, opened 2026-02-03
Image store parameters are invalid - Not a global variable


pyranota:
### Expected Behaviour
Naga validation to succeed, shader to be executed by bevy

### Example & Steps To Reproduce

1. Set up simple shader using texture store:
```rust
#[inline(never)]
#[spirv(compute(threads(8, 8, 1)))]
pub fn main(
    #[spirv(global_invocation_id)] id: UVec3,
    #[spirv(descriptor_set = 0, binding = 1)] image2d: &Image!(
        2D,
        format = rgba32f,
        sampled = false
    ),
) {
    unsafe {
        image2d.write(ivec2(1, 1), vec4(3., 2., 1., 0.));
    }
}
```
2. Compile
3. `naga /../shader/path.spv` (or use it in bevy_render)

I'm using rust-gpu `main` branch, `spirv-unknown-vulkan1.4` as a target in build.rs, but the error is the same for any target as far as I can tell.

Not sure if bevy reproduction set up is needed, as it fails with same error naga validation does. Something makes me think it is related to incorrect spv gen

## System Info
 - Rust: rustc 1.90.0-nightly (35f603652 2025-06-29)
 - OS: NixOS
 - GPU: AMD Radeon 890M
 - SPIR-V: SPIRV-Tools v2025.5 unknown hash

## Backtrace
```
[2026-02-03T15:08:12Z INFO  naga::front::spv] Generated by 983040 version 10600
[2026-02-03T15:08:12Z INFO  naga::front::spv] Patching...
Function [0] '' is invalid:
	Image store parameters are invalid
	Not a global variable

```

Firestar99:
Since `spirv-val` succeeds, it's more likely that naga is just failing to parse that spirv. Which is not surprising, there's quite a lot of spirv naga can't parse. (See pipelines in https://github.com/Rust-GPU/rust-gpu/pull/493 failing cause naga can't understand some spirv)

You can try to use `type=f32` instead of `format = rgba32f`, like in this alias?
https://github.com/Rust-GPU/rust-gpu/blob/ba15d4c65aa8845428ab6e9efef4ff499dc1e3ae/crates/spirv-std/src/image.rs#L68

You may also need these extra capabilities, set them via `SpirvBuilder.capability`:
```rust
Capability::StorageImageExtendedFormats,
Capability::StorageImageReadWithoutFormat,
Capability::StorageImageWriteWithoutFormat,
```

pyranota:
[...]

PR #523, opened 2026-02-04
Inline image intrinsics

Firestar99:
Inline all image intrinsics, helps with naga transpiling. Currently, all texel writes fail to transpile as the `OpImageWrite` is in a separate function that is called, inlining that call makes naga happy.

Also cargo update to fix cargo deny vuln reporting, insignificant for us, it's just the Android example.

Close https://github.com/Rust-GPU/rust-gpu/issues/522

### Additional context
See https://github.com/Rust-GPU/rust-gpu/issues/522#issuecomment-3846641840:
> While I generally agree that this is _technically_ a naga problem of not parsing valid spirv, I also like to look at the spirv we're emitting vs spirv emitted by glslc, to figure out what weird thing we're doing that may be breaking naga. And indeed, we are kinda weird.
> 
> Here I've transpiled the output of a compiletest just writing an image to glsl with `spirv-cross -V`:
> ```glsl
> #version 450
> layout(local_size_x = 8, local_size_y = 8, local_size_z = 1) in;
> 
> layout(set = 0, binding = 0, std430) readonly buffer fill_color
> {
>     vec4 _m0;
> } fill_color_1;
> 
> layout(set = 0, binding = 1) uniform readonly writeonly image2D image;
> 
> void _spirv_std_image_Image_f32_1_2_0_0_2_0_4_write_u32_glam_u32_uvec2_UVec2_(writeonly image2D _32, uvec2 _33, vec4 _34)
> {
>     imageStore(_32, ivec2(uvec2(_33)), _34);
> }
> 
> void main()
> {
>     vec4 _29 = fill_color_1._m0;
>     _spirv_std_image_Image_f32_1_2_0_0_2_0_4_write_u32_glam_u32_uvec2_UVec2_(image, uvec2(gl_GlobalInvocationID.xy), _29);
> }
> ```
> First, it'll fail to compile cause our image is both `read_only` and `write_only`. Removing `read_only`, you'll notice that our `imageStore` does indeed not operate on a global variable but on a parameter passed to the function. If we inline that function call, it works.
> 
> So yea... the fix is to add `#[inline]` to all the image instrinsics on our side

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -282,9 +282,9 @@ checksum = "175812e0be2bccb6abe50bb8d566126198344f707e304f45c648fd8f2cc0365e"
 
 [[package]]
 name = "bytemuck"
-version = "1.24.0"
+version = "1.25.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1fbdf580320f38b612e485521afda1ee26d10cc9884efaaa750d383e13e3c5f4"
+checksum = "c8efb64bd706a16a1bdde310ae86b351e4d21550d98d056f22f8a7f7a2183fec"
 dependencies = [
  "bytemuck_derive",
 ]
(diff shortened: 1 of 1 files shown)


pyranota:
🚀

Issue #388, opened 2025-09-16
RFC: link-time capabilities


Firestar99:
Link-time capabilities are capabilities that are not exposed to codegen, but only set at link time. Importantly, changing link-time capabilities does not require a clean recompile, unlike the current spirv capabilities. 

I good first experiment ground would likely be [Large Integer polyfills](https://github.com/Rust-GPU/rust-gpu/issues/307), which I think they are better suited at link time. A really common use-case would likely be to emit one shader module with Int64 Capabilitiy enabled, and another with u32 polyfill for all u64 integers. So implementing this at link time would allow you to compile the shader crate only once and link it twice.

There's a few variants I could imagine to specify these link-time capabilities:
* per entry point in the `#[spirv(capability(polyfill_u64))]`, mapping well with multimodule
* per shader crate using `#![spirv(capability(polyfill_u64))]` at root level, mapping well to single module
* some new system that allows specifying how entry points are grouped, potentially allowing entry points to be emitted in a variety of modules with different link-time capabilities. Maybe not with the first implementation.

Importantly, this should work well with rust features, so you can `#[cfg_attr(feature = "u64_emulation", spirv(capability(polyfill_u64)))]` and compile *only* the final shader crate twice, while reusing the remaining compiled crate tree.

I could also see a similar system being employed for usual spirv capabilities in the future, replacing the current pre-declaration of used features in spirv-builder. Similarly to the zombie system, you could attach an `OpCapability` to a certain instruction. If that instruction is linked into a shader module, add the `OpCapability` to it. But if it's unused, DCE will kill the instruction thus the capability. 

This reverses how you'd enable code that depends on some capability: Currently, you'd declare the capability in spirv-builder and have a `#[cfg(target_feature = "")]` in your code. With link-time capabilities, you use a normal rust feature like `#[cfg(feature = "")]`to gate your code requiring some capability, and if the end user enables that feature, it implicitly enables the required capability. The final shader crate binary can itself have features enabled though spirv-builder, that enable features of any such library gating required capabilities like you'd normally do in `Cargo.toml`.

PR #531, opened 2026-02-18
remove `has_capability()` checks

Firestar99:
* remove `BuilderSpirv::has_capability(&self, capability: Capability) -> bool`, you can no longer check whether some capability is enabled. Progress towards https://github.com/Rust-GPU/rust-gpu/issues/388
* add compiletests for non_uniform removal by post-link pass (previously *partially* gated behind `has_capability` check)

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -11,7 +11,7 @@ use crate::custom_insts::CustomInst;
 use crate::spirv_type::SpirvType;
 use itertools::Itertools;
 use rspirv::dr::{InsertPoint, Instruction, Operand};
-use rspirv::spirv::{Capability, MemoryModel, MemorySemantics, Op, Scope, StorageClass, Word};
+use rspirv::spirv::{MemoryModel, MemorySemantics, Op, Scope, StorageClass, Word};
 use rustc_abi::{Align, BackendRepr, Scalar, Size, WrappingRange};
 use rustc_apfloat::{Float, Round, Status, ieee};
 use rustc_codegen_ssa::MemFlags;
@@ -569,12 +569,10 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
 
     #[instrument(level = "trace", skip(self))]
     fn zombie_ptr_equal(&self, def: Word, inst: &str) {
-        if !self.builder.has_capability(Capability::VariablePointers) {
-            self.zombie(
-                def,
-                &format!("{inst} without OpCapability VariablePointers"),
-            );
-        }
+        self.zombie(
+            def,
+            &format!("{inst} without OpCapability VariablePointers"),
+        );
     }
 
     /// Convenience wrapper for `adjust_pointer_for_sized_access`, falling back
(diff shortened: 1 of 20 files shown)


crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
```diff
@@ -50,13 +49,11 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-            self.emit()
-                .decorate(ptr.def(self), Decoration::NonUniform, []);
-            self.emit()
-                .decorate(actual_index, Decoration::NonUniform, []);
-        }
+        // apply NonUniform to the operation and the index
+        self.emit()
+            .decorate(ptr.def(self), Decoration::NonUniform, []);
+        self.emit()
+            .decorate(actual_index, Decoration::NonUniform, []);
```
eddyb, line 56: Can you add a test using this feature, but not enabling `ShaderNonUniform`?

It should trigger a validation error from the new SPIR-T-based validation pass, if I'm not mistaken.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -569,12 +569,10 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
-        if !self.builder.has_capability(Capability::VariablePointers) {
-            self.zombie(
-                def,
-                &format!("{inst} without OpCapability VariablePointers"),
-            );
-        }
+        self.zombie(
+            def,
+            &format!("{inst} without OpCapability VariablePointers"),
+        );
```
eddyb: This one is weird. I would almost want to say that the instruction is not yet supported, rather than claiming any connection to `VariablePointers` (which doesn't give you pointer equality in general, just in very narrow cases).

eddyb (changes requested):

crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
Firestar99: Done, works without any issues. I seem to have added tests for everything *except* ByteAddressableBuffer.

eddyb: To be clear, I am talking about a test that lacks the `ShaderNonUniform` capability, and therefore results in an error. If it compiles successfully and contains `NonUniform` decorations, without the necessary capability, that's a bug IMO.

Firestar99: If you look at the tests this PR adds, there's a `ByteAddressableBuffer` variant with `NonUniform` capability, and a `ByteAddressableBufferUniform` variant without, all disassembled so you can diff that and see the the uniform variant missing the `OpDecoration %x NonUniform`

eddyb: But you removed the condition that gates `NonUniform` on `ShaderNonUniform`, and the decoration should always be emitted now... is *something else* removing the `NonUniform` decorations?

Firestar99: Yea, [here a linker pass that removes all NonUniform decorations](https://github.com/Rust-GPU/rust-gpu/blob/2cc79f16ca53f8c9b042250e7ef7313f1a4b3458/crates/rustc_codegen_spirv/src/linker/simple_passes.rs#L353-L367), which has been added in https://github.com/Rust-GPU/rust-gpu/pull/177 1.5 years ago. I wanted it so I could write just one image access asm! block with all the needed NonUniform decorations, since you can't conditionally exclude an asm line.

eddyb (approved):

PR #547, opened 2026-03-18
batch mem2reg to process all variables in a single pass

LegNeato:
mem2reg processed each variable independently.
This O(V*N) complexity caused multi-minute hangs on large post-inline functions.

Batch all operations into single passes over the instructions.

Fixes #546. Measured on the proof-of-space-gpu shader: mem2reg drops from 228s to 18s, total link from 236s to 26s.

Also adds unit tests.

Disclosure: largely AI.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/mem2reg.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mem2reg.rs
@@ -15,7 +15,6 @@ use rspirv::dr::{Block, Function, Instruction, ModuleHeader, Operand};
 use rspirv::spirv::{Op, Word};
 use rustc_data_structures::fx::{FxHashMap, FxHashSet, FxIndexMap};
 use rustc_middle::bug;
-use std::collections::hash_map;
 
 // HACK(eddyb) newtype instead of type alias to avoid mistakes.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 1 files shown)


LegNeato:
@eddyb can you review this one?

crates/rustc_codegen_spirv/src/linker/mem2reg.rs
```diff
@@ -433,46 +487,60 @@ fn top_stack_or_undef(
...
+            self.var_data[var_idx].base_var_type,
         );
+
+        // Search for an existing phi belonging to this variable.
+        let phi_defs = &self.var_data[var_idx].phi_defs;
+        let existing_phi = self.blocks[block]
+            .instructions
+            .iter_mut()
+            .take_while(|inst| inst.class.opcode == Op::Phi)
+            .find(|inst| phi_defs.contains(&inst.result_id.unwrap()));
```
Firestar99: You added the `take_while()`, is it guaranteed that Blocks always declare all their `OpPhi`s at the beginning of the block? I think yes, but just want to double check this.

crates/rustc_codegen_spirv/src/linker/mem2reg.rs
```diff
@@ -612,3 +710,565 @@ fn remove_old_variables(
...
+                Op::TypeInt
+                    if inst.operands[0].unwrap_literal_bit32() == 32
+                        && inst.operands[1].unwrap_literal_bit32() == 0 =>
+                {
+                    u32_type = Some(inst.result_id.unwrap());
+                }
+                Op::Constant if u32_type.is_some() && inst.result_type == u32_type => {
+                    let value = inst.operands[0].unwrap_literal_bit32();
+                    constants.insert(inst.result_id.unwrap(), value);
+                }
```
Firestar99, line 774: You're separating out u32 constants, yet a lot of your tests are using f32 constants?

Firestar99 (commented):
Looks sane to me, then again, I never looked at our mem2reg pass before. What's really interesting to me are the access patters of this pass, and how it reads and writes instructions, surely useful info for implementing some more efficient data structures for storing a module.

crates/rustc_codegen_spirv/src/linker/mem2reg.rs
LegNeato: Good catch! They are required to be at the beginning but apparently `OpLine` and `OpNoLine` can be interleaved and break this `take_while()`

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpPhi

> Within a block, this instruction must appear before all non-OpPhi instructions (except for [OpLine](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpLine) and [OpNoLine](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpNoLine), which can be mixed with OpPhi).

crates/rustc_codegen_spirv/src/linker/mem2reg.rs
LegNeato: The map only collects `u32` constants because it's used for resolving `OpAccessChain` indices. The tests use the `f32` store/load values. From https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpAccessChain, access chain indices must be scalar integers, and AFAICT rustc always emits u32 for these (and the previous code assumed `u32`). I added a doc comment.

nazar-pc:
As big improvement as it is, 18 seconds for `mem2reg` still seems way too long in absolute terms

LegNeato:
Yeah, I didn't want to do any major changes and just moved multiple loops to batches. I can take a closer look after we get to a steady state.

LegNeato:
@Firestar99 @eddyb any stamp here?

eddyb:
I'm getting back into catching up with Rust-GPU this week (ideally unblocking a release soon), and I'd want to wait for my own PR (last touched ~5 months ago) that both tracks the progress made on https://github.com/Rust-GPU/rust-gpu/issues/63 and includes a couple of additional optimizations (mostly doing more `mem2reg` and other clean-up, *during* inlining).

More specifically, I would like to see the remaining improvement from this PR (after rebasing it on my branch), on the "worst offenders" shaders we have (my old benchmarks + this newer one) - there's a solid chance there will still be some improvement left, but I've been scared of soundness subtleties in the past.

(The idea behind this PR seems fine, and the future SPIR-T "propagate values of locals" pass does handle all local variables together, instead of one at a time, but I haven't reviewed this yet)

LegNeato:
@eddyb review please.

LegNeato:
@eddyb ping

eddyb:
The changes (and measurements) I had mentioned are now up at:
- #552

<sub>(ran into complications with old code vs Rust-GPU `main` - ended up abandoning one of the projects being measured, but thankfully it was the fastest-to-compile one anyway)</sub>

There's two blockers I see with this PR:
- I can't repro the really long `ab-proof-of-space-gpu` compile times
  <sub>(do we know what _really_ happened? was it some kind of temporary regression that got fixed by the time I rebased on `main`?)</sub>
- there are a lot of changes, for what I suspect is likely equivalent to the commit titled:
  "_[2024] linker/mem2reg: apply rewrite rules only once per `insert_phis_all` invocation._"
  from #552
  <sub>(i.e. moving the `apply_rewrite_rules`+`remove_nops` calls out of a loop, and running them only once per `insert_phis_all` call, with everything else seeming to be AI causing non-functional refactors)</sub>

nazar-pc:
Very interesting. When I reported an issue, I was on 6a67e7b5954f37989ad540a555b5d6969073592e, just re-tested with it and `build.rs` execution takes roughly 340 seconds.

I since upgraded to 877bd8697a15f3e6d09446a5e1807e6237ca1dac and now `build.rs` takes roughly 20-30 seconds, which is a massive difference.

No idea how to explain it, but things are not nearly as bad as they were before.

Firestar99:
Just before the problematic 6a67e7b5954f37989ad540a555b5d6969073592e we upgraded toolchain to2025-10-28 (rustc 1.92.0), and upgraded toolchain multiple times right afterwards as well, so could very well be some upstream "bug" that got fixed. Could totally imagine something generated too much code and LLVM just optimized it away so noone noticed?

LegNeato:
@eddyb should we still land this or no?

PR #549, opened 2026-04-02
spirv-builder: fall back to hashed codegen dylibs

LegNeato:
`spirv-builder` only looks for the exact `librustc_codegen_spirv.{so,dylib,dll}` filename in the dylib search path.

That works in the `rust-gpu` workspace, but it misses external `build.rs` consumers where Cargo only exposes hashed deps artifacts such as `librustc_codegen_spirv-(hash).so`. This showed up in a downstream workspace update to current `rust-gpu` main where the consumer path only had hashed backend artifacts available.

Keep preferring the exact filename when present, then fall back to the newest hashed `rustc_codegen_spirv` dylib in the same search paths. Add tests for exact matches, hashed fallback, and exact-over-hashed precedence.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -83,10 +83,12 @@ use serde::Deserialize;
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::env;
+use std::ffi::OsStr;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
+use std::time::SystemTime;
 use thiserror::Error;
 
 #[cfg(feature = "watch")]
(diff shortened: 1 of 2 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -805,19 +807,70 @@ fn dylib_path() -> Vec<PathBuf> {
...
-            path.push(&filename);
-            if path.is_file() {
-                return Ok(path);
-            }
+        let filename = rustc_codegen_spirv_dylib_name();
+        if let Some(path) = find_rustc_codegen_spirv_in_paths(dylib_path()) {
+            return Ok(path);
         }
         panic!("Could not find {filename} in library path");
     } else {
```
Firestar99: nit: `filename` could be moved downwards right before the panic, to not allocate a string that's only used when panic-ing

Firestar99 (approved):
I assume this is something newer nightlies introduced? Cause I'm not aware of cargo adding a hash to the `.so`. 

If it is, it may finally resolve the many warnings some projects get on windows about conflicting debug symbol files from `.dll` and `.lib` files being named the same.

crates/spirv-builder/src/lib.rs
LegNeato: Good call, done.

PR #554, opened 2026-04-08
update toolchain to nightly-2026-04-08

Firestar99:


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -3466,4 +3458,8 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
     fn apply_attrs_to_cleanup_callsite(&mut self, _llret: Self::Value) {
         // Ignore
     }
+
+    fn alloca_with_ty(&mut self, _layout: TyAndLayout<'tcx>) -> Self::Value {
+        bug!("scalable alloca is not supported in SPIR-V backend")
+    }
```
Firestar99, line 3464: this is the replacement for `scalable_alloca`, otherwise this is never called, although judging by the name who knows what'll be used for in the future

nnethercote (approved):

crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -338,16 +340,15 @@ impl WriteBackendMethods for SpirvCodegenBackend {
...
         _exported_symbols_for_lto: &[String],
-        _each_linked_rlib_for_lto: &[PathBuf], // njn: ?
-        modules: Vec<(String, Self::ModuleBuffer)>,
-        cached_modules: Vec<(SerializedModule<Self::ModuleBuffer>, WorkProduct)>,
+        _each_linked_rlib_for_lto: &[PathBuf],
+        _modules: Vec<ThinLtoInput<Self>>,
     ) -> (Vec<ThinModule<Self>>, Vec<WorkProduct>) {
-        link::run_thin(cgcx, modules, cached_modules)
+        // Note(@firestar99): gcc impl this as unreachable as well
+        unreachable!()
```
LegNeato, line 351: Are we sure this is unreachable? What if someone sets LTO to thin?

https://doc.rust-lang.org/cargo/reference/profiles.html#lto

Wouldn't what we did before, just pass-through, not be better?

Firestar99: The gcc impl is just `unreachable!()`, as gcc doesn't seem to have ThinLTO: [rust-lang](https://github.com/rust-lang/rust/blob/1fe72d35998dea48aeecaf7fc07783b0b553f24f/compiler/rustc_codegen_gcc/src/lib.rs#L445-L455)

I personally feel like we've got quite a few things implemented "in case they are used" without actually knowing when they're used. So we can't test them either, so there's no way to know whether our "mock" implementations actually work. To me, removing these dead code paths by a hard error and seeing it never trigger is a good thing, as it means less code to maintain. 

With the ThinLTO interface having gotten a refactor the past week, I don't exactly know how to actually implement this pass-through correctly, and even worse, how to even test it.

LegNeato: sgtm.

nazar-pc:
Looking forward to newer nightly, apparently the few recent ones (including one used here) had a regression, which results in ICE on repeated compilation: https://github.com/rust-lang/rust/issues/154878

Firestar99:
Just double checked, https://github.com/Rust-GPU/rust-gpu/pull/563 has that fix / revert included

PR #560, opened 2026-04-13
Add `--dump-spirt` for dumping only the final SPIR-T module (unlike `--dump-spirt-passes`).

eddyb:
Small convenience flag for peeking at the final SPIR-T (useful when `--dump-spirt-passes` gets unwieldly, or its level of detail is entirely unnecessary).


LegNeato (approved):
Looks nice and straightforward

PR #562, opened 2026-04-13
builder: allow unsigned integers in `neg` and `exactsdiv`.

eddyb:
`rustc_codegen_ssa` relies on signed operations on unsigned integers for e.g. `size_of_val`.

---

**TODO**: move this from the compute shader (see diff) to a proper test:
```rust
core::mem::size_of_val(
    const {
        struct S<T: ?Sized>(T);
        &S([]) as &S<[()]>
    },
)
```

### DIFF
--- a/tests/compiletests/ui/lang/core/mem/size_of_val_unsized.stderr
+++ b/tests/compiletests/ui/lang/core/mem/size_of_val_unsized.stderr
@@ -1,0 +1,23 @@
+error: unsupported unsized `[struct () {  }]` constant
+   |
+   = note: used by unnamed global (%68)
+   = note: used by unnamed global (%69)
+note: used from within `size_of_val_unsized::main`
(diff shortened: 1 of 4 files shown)


tests/compiletests/ui/lang/core/mem/size_of_val_unsized.stderr
```diff
@@ -0,0 +1,23 @@
...
+LL | |     );
+   | |_____^
+note: called by `main`
+  --> <$DIR/size_of_val_unsized.rs>:4:8
+   |
+LL | pub fn main(out: &mut usize) {
+   |        ^^^^
+
+error: aborting due to 1 previous error
+
```
Firestar99, line 23: I think the patch is fine, saw the todo, moved it to a compiletest and it resulted in this interesting error

PR #564, opened 2026-04-13
setup release-plz and release v0.10.0-alpha.1

Firestar99:
> [!CAUTION]  
> Merging this PR will trigger the release!

* setup release-plz
  * changelogs must be manually written due to our commits not following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
  * configured to only do things once and *not* for every crate:
    * only one tag `v0.10.0-alpha.1`
    * only one changelog in `/CHANGELOG.log`
  * trusted publishing is set up
    * published a dummy crate to `cargo-gpu-install` to set that up, since trusted publishing only works on existing crates
* release `v0.10.0-alpha.1`
  * manually written release, will be published by release-plz once this PR merges (due to branch name matching `release-plz*`, according to [these docs](https://release-plz.dev/docs/config#the-release_always-field))
  * changelog contents were initially extracted by AI, I cleaned it up *a lot* and grouped them semantically
  * Do we want a highlights section in the changelog or as a blog post?
* see https://github.com/Rust-GPU/spirv-tools-rs/pull/29 as reference

## Problem
The release-pr updating the version number updates `/Cargo.lock` but not `/tests/difftests/tests/Cargo.lock`, leading to CI failure *when merging future releases*. Not yet sure how to fix, needs to be fixed so the release PRs of `alpha.2` work, so we could go ahead with this release and fix this later.

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,57 @@
+name: Release-plz
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
+    if: ${{ github.repository_owner == 'rust-gpu' }}
+    runs-on: ubuntu-latest
+    environment: release
+    permissions:
+      contents: write
+      pull-requests: read
+      id-token: write
+    steps:
+      - &checkout
+        name: Checkout repository
+        uses: actions/checkout@v6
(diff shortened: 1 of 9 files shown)


LegNeato:
Let's just release alphas daily to make sure everything is working in the release process end-to-end...we don't need to wait, that's the point of alpha.

LegNeato (approved):

PR #565, opened 2026-04-13
fix No OpEntryPoint for non-public entry points

39ali:
fixes the confusing `No OpEntryPoint instruction was found` you get when entry point is private 
#293

### DIFF
--- a/crates/spirv-std/macros/src/lib.rs
+++ b/crates/spirv-std/macros/src/lib.rs
@@ -151,6 +151,19 @@ pub fn spirv(attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut tokens = quote! { #[cfg_attr(target_arch="spirv", rust_gpu::#spirv(#attr))] };
 
     let item: proc_macro2::TokenStream = item.into();
+    // If the annotated item is a function without `pub`, automatically add it.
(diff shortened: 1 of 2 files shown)


LegNeato:
Ohhh, that is a good idea! I was be concerned about automatic `pub` breaking encapsulation, but putting it as part of the macro is extremely smart as (AFAICT) it solves that....any code not targeting spirv will have the macro stripped away and still see the private function.

@Firestar99 do you see any potential issues I am missing?

Firestar99 (approved):
It is an improvement, but doesn't solve it for the following code:
```rust
mod private {
    #[spirv(fragment)]
    pub fn main() {}
}
```

I think we can merge this anyway, but do note that when I do get back to the `asm!`-based entry point system the next weeks, it shouldn't suffer from this problem at all, since `global_asm!` is always accessible (doesn't need to be "monomorphized" by being pub and accessible from potentially external crates).

39ali:
@Firestar99 we can `#[export_name]` and have some unique hash for functions, that should fix it but if entry point system is changing then this should be okish for now

LegNeato:
Thank you for the PR! 🍻

Firestar99:
`#[unsafe(no_mangle)]` also works to make it globally visible

PR #567, opened 2026-04-13
difftest: fix 5 difftests silently passing on zeroed inputs

LegNeato:
Also fixes `trig_ops-rust` clamping asin's output instead of its input.


Firestar99 (approved):
Had a quick scroll through, looking good!

Won't verify every single comment against the shader since it's just tests

PR #569, opened 2026-04-15
add support for unions

39ali:
Rust unions require reinterpreting the bits of one type as another — e.g. reading `[f32; 5]` as `struct { a: i32, b: [f32; 3], c: u32 }.` In SPIR-V this would naively lower to an `OpBitcast ` between two composite types, which is illegal by the spec: `OpBitcast` only accepts scalar, vector, or pointer operands, never structs or arrays.

previously, any union field access involving layout-compatible but differently-typed SPIR-V composites would either produce invalid SPIR-V or fail to compile.

solution:

implement two new codegen paths in `builder_methods.rs` that handle type-mismatched copies without relying on composite OpBitcast:

`memcpy_const_size` — used by the memcpy / OpCopyMemory path (pointer-to-pointer copies).

`try_store_composite_scalar_leaves` — used by the store path (value-to-pointer writes).

Both work by:

decomposing both the source and destination types recursively into scalar leaves (`collect_scalar_leaves`), collecting the OpAccessChain index path, SPIR-V type, and bit width for each leaf.

grouping leaves by equal total bit widths (`group_leaves_by_bits`), this handles both 1:1 (same scalar count) and N:M (different counts, same total bits) cases.

emitting element-wise copies per group:

1→1: OpLoad + optional OpBitcast + OpStore
N→1 (pack): load N scalars, OpCompositeConstruct into an intermediate uN×k vector, OpBitcast to the wider dst type, OpStore
1→N (unpack): OpLoad, OpBitcast to an intermediate uN×k vector, OpCompositeExtract each component, OpStore each

solves #241


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1278,6 +1293,533 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
         result_id.with_type(ptr_ty)
     }
+
+    /// attempt to copy memory between two pointers of different but layout-compatible types
+    /// by recursively decomposing both into scalar leaves and emitting element-wise
+    /// `OpLoad` + `OpStore` pairs (with `OpBitcast` for type mismatches). This is used to
+    /// support union field access, e.g. `[f32; 5]` -> `struct { a: f64, b: [f32; 3] }`.
+    /// handles both 1:1 (same scalar count) and N:M (different counts but same total bits)
+    /// returns `false` if the types are not compatible.
+    fn memcpy_const_size(&mut self, dst: SpirvValue, src: SpirvValue) -> bool {
```
Firestar99, line 1314: What's the difference between `memcpy_const_size()` and `try_store_composite_scalar_leaves()`? Both functions seem to be almost the exact same, apart from memcpy's arg being `src: SpirvValue` as a pointer and the other one `val: SpirvValue` as a value.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1278,6 +1293,533 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+                            .emit()
+                            .composite_extract(vec_elem_ty, None, vec_val.def(self), [i as u32])
+                            .unwrap()
+                            .with_type(vec_elem_ty);
+                        let casted = self.bitcast_leaf_if_needed(component, vec_elem_ty, *dst_ty);
+                        self.store_scalar_leaf(dest_ptr, dst_indices, *dst_ty, casted);
+                    }
+                }
+
+                _ => return false,
```
Firestar99, line 1630: I see an implementation for 1 => 1, 1 => N, N => 1 but no N => M?

This compiletest fails:
```rust
// build-pass
// compile-flags: -C target-feature=+Int64

use spirv_std::spirv;

union MyUnion {
    a: (u32, u64),
    b: (u64, u32),
}

#[spirv(fragment)]
pub fn union_mixed_types(
    out: &mut u64,
) {
    let bla = MyUnion { a: (42, 123) };
    *out = unsafe { bla.b }.0;
}
```

Firestar99 (changes requested):
I like the simplicity of the implementation, this was pretty easy to review, but it does ring "AI impl" bells.

PR #579, opened 2026-04-15
upgrade examples to wgpu 30

Firestar99:


examples/runners/wgpu/src/graphics.rs
```diff
@@ -137,58 +131,10 @@ async fn run(
...
-            .as_ref()
-            .err()
-            .map(|(_, layout, _)| layout)
-            .as_slice(),
-        push_constant_ranges: push_constants_or_rossbo_emulation
-            .as_ref()
-            .map_or(&[], slice::from_ref),
+        bind_group_layouts: &[],
+        immediate_size: size_of::<ShaderConstants>() as u32,
     });
```
eddyb, line 181: I think you could've kept the emulation (which is needed for e.g. running on actual WebGPU via wasm - unless `wgpu` does its own emulation internally? if that's true, you can remove the `emulate_push_constants_with_storage_buffer` option entirely!).

As per https://github.com/Rust-GPU/rust-gpu/pull/387, you should be able to test the wasm build (and its use of WebGPU) via:
```
rustup target add wasm32-unknown-unknown
cargo run-wasm -p example-runner-wgpu
```
(you may need Chrome w/ e.g. `--enable-features=Vulkan --enable-unsafe-webgpu` - not sure if Firefox has gotten around to enabling WebGPU at all, outside nightly)

---

The main difference seems to be (assuming this emulation needs to be kept, as per above) that instead of `wgpu::PushConstantRange`, `immediate_size` is just an `u32`, so the non-emulation case would just become `Ok(PUSH_CONSTANTS_SIZE)`.

eddyb (changes requested):
It's been long enough that wgpu 30 is already out, oops!

examples/runners/wgpu/src/graphics.rs
Firestar99: Rereading the documentation:

```
/// Supported platforms:
/// - DX12
/// - Vulkan
/// - Metal
/// - OpenGL (emulated with uniforms)
/// - WebGPU
```

So I thought it's supported for every platform, but:
> WebGPU support is currently a proposal and will be available in browsers in the future.

So I guess we'll keep our emulation...

Firestar99:
Updated it all the way to wgpu30 without removing our push constant emulation. Moved the removal to branch [`wgpu_remove_push_constant_emu`](https://github.com/Rust-GPU/rust-gpu/tree/wgpu_remove_push_constant_emu)

eddyb (approved):

PR #580, opened 2026-04-15
add slice cast support 

### LINKED ISSUE #465: Using an array as a slice causes a compiler error
I'm surprised I haven't hit this earlier, but it's not possible to pass an array as a slice. Is this a known limitation? Is there an easy workaround?

### Expected Behaviour

I would expect to be able to pass an array like `[u32; 3]` as a parameter to a function that takes a slice.
 
Here is a minimal reproducible example:

```rust
#![no_std]
#![allow(unexpected_cfgs)]

fn do_work(data: &[u32], slab: &mut [u32]) {
    slab[0] = data[0];
    slab[1] = data[1];
    slab[2] = data[2];
}

#[spirv(compute(threads(16)))]
pub fn compute_shader(
    #[spirv(storage_buffer, descriptor_set = 0, binding = 0)] slab: &mut [u32],
    #[spirv(global_invocation_id)] global_id: glam::UVec3,
) {
    let data = [global_id.x, global_id.y, global_id.z];
    do_work(&data, slab);
}
```

## System Info

- Rustc version: `1.90`
- spirv-std: from the latest `main`, `b3eda4df9814b6176d3c0844eb67f70a22ebb378` as of this writing
- SPIR-V version: 
  ```
  ❯ spirv-val --version
  SPIRV-Tools v2025.3 unknown hash, 2025-06-23T19:59:33+00:00
  Targets:
    SPIR-V 1.0
    SPIR-V 1.1
    SPIR-V 1.2
    SPIR-V 1.3
    SPIR-V 1.4
    SPIR-V 1.5
    SPIR-V 1.6
    SPIR-V 1.2 (under OpenCL 2.2 Full Profile semantics)
    SPIR-V 1.0 (under Vulkan 1.0 semantics)
    SPIR-V 1.3 (under Vulkan 1.1 semantics)
    SPIR-V 1.4 (under Vulkan 1.1 semantics)
    SPIR-V 1.5 (under Vulkan 1.2 semantics)
    SPIR-V 1.6 (under Vulkan 1.3 semantics) SPIR-V 1.6 (under Vulkan 1.4 semantics)
  ```

39ali:
fixes casting array as a slice , or array to ptr :

 *[T; N] -> *[T] 
 
  *[T; N] -> *T

fixes #465

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2423,7 +2423,28 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
                 "pointercast called on non-pointer dest type: {other:?}"
             )),
         };
-        let dest_pointee_size = self.lookup_type(dest_pointee).sizeof(self);
+
+        let dst_pointee_ty = self.lookup_type(dest_pointee);
+        let dest_pointee_size = dst_pointee_ty.sizeof(self);
+        let src_pointee_ty = self.lookup_type(ptr_pointee);
+
+        // *[T; N] -> *RuntimeArray<T>
+        if let SpirvType::Array { element: elem_ty, .. } = src_pointee_ty
+            && let SpirvType::RuntimeArray { element: rt_elem_ty } = dst_pointee_ty
+            && elem_ty == rt_elem_ty
+        {
+            let zero = self.constant_u32(self.span(), 0).def(self);
+            let elem_ptr_ty = self.type_ptr_to(elem_ty);
+            let elem_ptr = self
+                .emit()
+                .in_bounds_access_chain(elem_ptr_ty, None, ptr.def(self), [zero])
+                .unwrap();
+            return self
+                .emit()
+                .bitcast(dest_ty, None, elem_ptr)
+                .unwrap()
+                .with_type(dest_ty);
+        }
 
         if let Some((indices, _)) = self.recover_access_chain_from_offset(
             ptr_pointee,
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2429,8 +2429,12 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         let src_pointee_ty = self.lookup_type(ptr_pointee);
 
         // *[T; N] -> *RuntimeArray<T>
-        if let SpirvType::Array { element: elem_ty, .. } = src_pointee_ty
-            && let SpirvType::RuntimeArray { element: rt_elem_ty } = dst_pointee_ty
+        if let SpirvType::Array {
+            element: elem_ty, ..
+        } = src_pointee_ty
+            && let SpirvType::RuntimeArray {
+                element: rt_elem_ty,
+            } = dst_pointee_ty
             && elem_ty == rt_elem_ty
         {
             let zero = self.constant_u32(self.span(), 0).def(self);
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -660,6 +660,7 @@ pub fn link(
             peephole_opts::composite_construct(&types, func);
             peephole_opts::vector_ops(output.header.as_mut().unwrap(), &types, func);
             peephole_opts::bool_fusion(output.header.as_mut().unwrap(), &types, func);
+            peephole_opts::fold_array_bitcast_access_chain(&types, func);
         }
     }
 
--- a/crates/rustc_codegen_spirv/src/linker/peephole_opts.rs
+++ b/crates/rustc_codegen_spirv/src/linker/peephole_opts.rs
@@ -680,3 +680,94 @@ pub fn fold_load_from_constant_variable(module: &mut Module) {
         }
     }
 }
+
+/// Eliminate the `OpBitcast` that arises from `*[T; N] → *RuntimeArray<T>` pointer casts.
+///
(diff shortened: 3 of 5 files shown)


Firestar99 (approved):
Rebased it on main, squashed the fmt fixes and added disassembly and arrays from consts to the compiletest.

Overall, I'm super impressed how simple this has turned out to be. I haven't looked at our existing pointer emulation too deeply before, but this seems quite logical.

@LegNeato But this also shows that I've been wrong about `RuntimeArray` only being used via `spirv_std::RuntimeArray`, it is also used to represent slices. Just that those slices get removed in the post-link legalization, since they contain a pointer, so you don't actually see them in the disassembly of the compiletest.
See:
https://github.com/Rust-GPU/rust-gpu/blob/bdf8bae42c50ea0690687d926ccb77cf03b8ced7/crates/rustc_codegen_spirv/src/abi.rs#L637-L646

39ali:
@Firestar99  we should probably use `LogicalPtrCast` in the future and resolve it in `qptr`, but the issue it'll get zombied since i think `qptr` is not on by default

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -2423,7 +2423,46 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+        } = src_pointee_ty
+            && elem_ty == dest_pointee
+        {
+            let zero = self.constant_u32(self.span(), 0).def(self);
+            return self
+                .emit()
+                .in_bounds_access_chain(dest_ty, None, ptr.def(self), [zero])
+                .unwrap()
+                .with_type(dest_ty);
+        }
```
eddyb: How come this is needed, won't `recover_access_chain_from_offset` produce the same effect?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -2423,7 +2423,46 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+
+        // array -> RuntimeArray: *[T; N] -> *[T]
```
eddyb: Looks like this is doing `*[T; N] -> *T -> *[T]`, using an `OpInBoundsAccessChain` for the first step, and `OpBitcast` for the second step, but:
1. that `OpBitcast` is illegal, but won't be seen by the same zombie-creating logic that would be triggered by using `self.pointercast(elem_ptr, dest_ty)`
2. creating `SpirvValueKind::LogicalPtrCast` (like calling `pointercast` recursively would do) allows later stripping the cast, so it should be more helpful?

I can't see how this helps, other than hiding an error?

eddyb (changes requested):
I'm not sure why this change even helps - is it hiding errors that don't end up mattering in the end?

One way you could check is compiling (using some Rust-GPU version without this change) with `RUSTGPU_CODEGEN_ARGS=--no-early-report-zombies` (which moves the "zombie" reporting much later, near the end of the SPIR-T part of the linker).

For the test, I believe this is equivalent:
```rust
// compile-flags: -C llvm-args=--no-early-report-zombies
```

eddyb:
The test from this PR passes on `main` with only this change:
```patch
--- a/tests/compiletests/ui/lang/core/array-slice-cast.rs
+++ b/tests/compiletests/ui/lang/core/array-slice-cast.rs
@@ -1,5 +1,5 @@
@@ -1,5 +1,5 @@
 // build-pass
-// compile-flags: -C llvm-args=--disassemble
+// compile-flags: -C llvm-args=--disassemble -C llvm-args=--no-early-report-zombies
```

I've been meaning for a while to replace the old zombie system:
- https://github.com/EmbarkStudios/rust-gpu/issues/1095
  (TIL it didn't get migrated: https://github.com/Rust-GPU/rust-gpu/issues/581)

---

In this specific case, I am guessing inlining removes the need for a `*[T]`-typed pointer, but I don't recall us having a pass to actually remove `OpBitcast`s, so I'm not sure how this ends up working.

39ali:
thanks @eddyb for the clarifications and you're right, 
take 2 : 
with main the only way it'll work with the new test is if you use : `// -C llvm-args=--spirt-passes=qptr -C llvm-args=--no-early-report-zombies` which is not ideal , the ultimate solution will be to make `qptr` default  and fix the zombie system.

the solution that i just pushed and would work right away is to do the  illegal `OpBitcast` and fix/fold it once i have more info in the linking stage and this approach would also work with `qptr` enabled.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
39ali: oh yes this is redundant, i'll remove it

39ali:
@eddyb  i noticed that each slice index `data[i]` and `slab[i]` generates a Rust bounds check. Rust emits something like:
`if i < slice.len() { /* ok */ } else { panic!() }`

The data checks (0 < 3, 1 < 3, 2 < 3) are constant comparisons ,so a better way would be to fold them to true and eliminate those branches entirely. is this something that is supported currently and it's just not working or do we need to add support ?

eddyb:
> The data checks (0 < 3, 1 < 3, 2 < 3) are constant comparisons ,so a better way would be to fold them to true and eliminate those branches entirely. is this something that is supported currently and it's just not working or do we need to add support ?

This should fold comparisons away, and there's a check in `rustc_codegen_ssa::mir` which should optimize the conditional branch away (but I'm not sure if there's some other indirection that is blocking any of this):
https://github.com/Rust-GPU/rust-gpu/blob/c1acf6e4b2d338b0f822457350ea91db21757f1a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L2507-L2560

irisfalke:
any info on this problem? it makes working with buffers annoying.

Issue #307, opened 2025-06-30
Integer polyfills


Firestar99:
We should have polyfills to downlevel `u64` and `u128` to many `u32` instructions, enabling the use of these types without concern for hardware capabilities on eg. mobile ARM platforms. Also consider polyfill for u8 and u16, which could enable things like using `enum Ordering` from std, see https://github.com/Rust-GPU/rust-gpu/issues/147.

`libm` only handles float math, no integer downlevel.

### Discussed in https://github.com/Rust-GPU/rust-gpu/discussions/301

<div type='discussions-op-text'>

<sup>Originally posted by **nazar-pc** June 27, 2025</sup>
I faced a small challenge: my CPU code I'm trying to run on a GPU naturally maps onto `u64` and `u128`, but even `u64` apparently requires `OpCapability Int64`, let alone `u128`.

Now writing polyfills that work around that feels like a job that compiler should do for me, though I appreciate that it at least tells me that there isn't a native way to run that and I should think about what to do with it.

I'm wondering it it'd be possible to either have a compilation flag that automatically inserts polyfills for such types. It will be an explicit opt-in feature, but otherwise require the least amount of effort for those who do need it. I believe LLVM is capable of this.

I'm really not looking forward to writing polyfills for `u128` using a bunch of `u32`s.</div>

Firestar99:
### Implementation Notes

For anyone investigating this, I can at least paste these relevant code locations:
* where all basic math functions are implemented
https://github.com/Rust-GPU/rust-gpu/blob/2f915a09f1cc8cbdced277dc6ccf4c5bcb5de6ee/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L1538-L1588
* with this proc macro just emitting the correct instruction plus some const eval, nothing wrt emulation to be found
https://github.com/Rust-GPU/rust-gpu/blob/2f915a09f1cc8cbdced277dc6ccf4c5bcb5de6ee/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L79-L82
* I already made some downlevels for `count_ones`, `bit_reverse`, `count_trailing_zeros` and `count_leading_zeros` here, since spirv requires these instructions to be u32 exclusive.
https://github.com/Rust-GPU/rust-gpu/blob/2f915a09f1cc8cbdced277dc6ccf4c5bcb5de6ee/crates/rustc_codegen_spirv/src/builder/intrinsics.rs#L361-L592

[...]

PR #585, opened 2026-04-21
promote int8 to int32 if int8 is not supported

### LINKED ISSUE #409: `PartialOrd` doesn't work with new type
I have the following trivial new type:
```rust
#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, From, Into)]
pub struct CustomPosition(u32);
```

When trying to compare two values of this type, I'm getting an error:
```
  error: `i8` type used without `OpCapability Int8`
       |
  note: used from within `<u32 as core::cmp::PartialOrd>::partial_cmp`
      --> /home/nazar-pc/.rustup/toolchains/nightly-2025-06-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs:1989:5
       |
  1989 |     ord_impl! { char usize u8 u16 u32 u64 u128 isize i8 i16 i32 i64 i128 }
       |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  note: called by `<ab_proof_of_space_gpu::shader::sort_buckets::CustomPosition as core::cmp::PartialOrd>::partial_cmp`
      --> crates/farmer/ab-proof-of-space-gpu/src/shader/sort_buckets.rs:17:35
       |
  17   | #[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, From, Into)]
       |                                   ^^^^^^^^^^
  note: called by `<ab_proof_of_space_gpu::shader::sort_buckets::CustomPosition as core::cmp::PartialOrd>::le`
      --> /home/nazar-pc/.rustup/toolchains/nightly-2025-06-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs:1403:14
       |
  1403 |         self.partial_cmp(other).is_some_and(Ordering::is_le)
       |              ^^^^^^^^^^^^^^^^^^
  note: called by `ab_proof_of_space_gpu::shader::sort_buckets::perform_compare_swap`
      --> crates/farmer/ab-proof-of-space-gpu/src/shader/sort_buckets.rs:80:32
       |
  80   |     let (smaller, larger) = if a <= b { (a, b) } else { (b, a) };
       |                                ^^^^^^
```

At the same time, comparison of `u32` directly works fine and doesn't require `Int8` capability.

While this is likely related to discussion in https://github.com/Rust-GPU/rust-gpu/issues/307, I think it is distinct enough to be reported separately.

39ali:
fixes #409

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
+++ b/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
@@ -1,5 +1,5 @@
 use super::{get_name, get_names};
-use rspirv::dr::{Block, Function, Module};
+use rspirv::dr::{Block, Function, Module, Operand};
 use rspirv::spirv::{Decoration, ExecutionModel, Op, Word};
 use rustc_codegen_spirv_types::Capability;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/linker/simple_passes.rs
```diff
@@ -365,3 +365,79 @@ pub fn remove_non_uniform_decorations(_sess: &Session, module: &mut Module) -> s
...
+    for inst in &mut module.types_global_values {
+        // widen each 8-bit OpTypeInt to 32 bits
+        if inst.class.opcode == Op::TypeInt
+            && let Some(id) = inst.result_id
+            && narrow_types.contains_key(&id)
+        {
+            inst.operands[0] = Operand::LiteralBit32(32);
+        }
+
+        // fix OpConstant values: sign-extend signed 8-bit constants to 32 bits.
```
Firestar99, line 429: *Technically*, you are *only* sign-extending signed integers and zero-extending unsigned ones

Firestar99 (changes requested):
This is incredibly dangerous, as you're promoting *every single* u8 to u32, regardless of where it is used. Imagine you're using `OpIAddCarry` and promoting that one, studdenly you're getting wrong results.

And if the single `u8` delcaration is used *anywhere* in an `OpTypePointer`, even in another entry point unrelated to yours, you just back off and do nothing, meaning that unrelated code could affect whether this rewriting goes through or not. 

Instead, you need to have a whitelist about what can be promoted, which is a huge rabbit hole. Feel free to read the discussion in https://github.com/Rust-GPU/rust-gpu/issues/307 (which is linked from the issue you're trying to close).

39ali:
oh wow thanks for the clarification, it's indeed a rabbit hole, i'll take a shot at addressing it

PR #586, opened 2026-04-22
Fix slow compilation of large zero-initialized arrays; warn on large non-zero fills

39ali:
`let arr = [0; 32 * 1024]` compiled very slowly. The cause: Rvalue::Repeat with a zero element routes through `memset`, while will create a huge list of repeated store instructions, the fix: use `OpConstantNull` instead when `fill_byte == 0`, valid for any SPIR-V composite type, single instruction regardless of size.

SPIR-V has no equivalent of `OpConstantNull` for non-zero values , so the best we can do is tell the user early by emitting a warning

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -37,6 +37,8 @@ use std::ops::{BitAnd, BitOr, BitXor, Not, RangeInclusive};
 use tracing::{Level, instrument, span};
 use tracing::{trace, warn};
 
+const LARGE_ARRAY_MEMSET_WARN_THRESHOLD: usize = 1024;
+
 enum ConstValue {
     Unsigned(u128),
     Signed(i128),
@@ -425,8 +427,18 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                 .def(self)
             }
             SpirvType::Array { element, count } => {
+                if fill_byte == 0 {
+                    return self.constant_null(ty.def(self.span(), self)).def(self);
+                }
                 let elem_pat = self.memset_const_pattern(&self.lookup_type(element), fill_byte);
                 let count = self.builder.lookup_const_scalar(count).unwrap() as usize;
+                if count > LARGE_ARRAY_MEMSET_WARN_THRESHOLD {
+                    self.warn(format!(
+                        "large array of {count} elements with a non-zero fill will generate a \
+                         SPIR-V constant with {count} operands, which may be slow to compile; \
+                         consider using a storage buffer for large data instead"
+                    ));
+                }
                 self.constant_composite(ty.def(self.span(), self), iter::repeat_n(elem_pat, count))
                     .def(self)
             }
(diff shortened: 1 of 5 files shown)


Firestar99 (changes requested):
> Fix slow compilation

First, could you explain what kind of slowdown you're seeing?

> use OpConstantNull instead when fill_byte == 0

I like that idea, although I'm unsure if spirt supports that command @eddyb?

> will create a huge list of repeated store instructions

Isn't the bug then that the emitted `OpConstantComposite` was broken up into a bunch of individual `OpStore` instructions / not being reassembled into an `OpCompositeConstruct` / `OpConstantComposite`? (Our [`destructure_composites`](https://github.com/Rust-GPU/rust-gpu/blob/main/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs) post-link pass *may* be related to this?)

39ali:
i needed to clarify that for `let arr = [0; 32 * 1024]`  it'll translate to `OpConstantComposite` 
but for `arr = [1; 32 * 1024]` it'll translate to `OpInBoundsAccessChain` + `OpStore`

##### First, could you explain what kind of slowdown you're seeing?
compile time is slow, i believe its because of the long `OpConstantComposite` instruction  

i'll attach the .spv file for this shader :
```
#![no_std]
use spirv_std::glam::Vec4;
use spirv_std::spirv;

#[inline(never)]
#[spirv(compute(threads(1)))]
pub fn self_assignment(
    #[spirv(storage_buffer, descriptor_set = 0, binding = 1)] y: &mut [f32],
) {
    let arr = [0f32; 32 * 1024];
    for i in 0..arr.len() {
        y[i] = arr[i];
    }
} 
```

[spirv.txt](https://github.com/user-attachments/files/27069158/spirv.txt)

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -425,8 +427,18 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                 .def(self)
             }
             SpirvType::Array { element, count } => {
+                if fill_byte == 0 {
+                    return self.constant_null(ty.def(self.span(), self)).def(self);
```
eddyb, line 431: #### Context (SPIR-T)

(Future) SPIR-T supports `OpConstantNull` (e.g. in https://github.com/Rust-GPU/spirt/pull/45) so this is definitely a good catch and won't cause any issues!

I haven't implemented [`SPV_EXT_replicated_composites`](https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_replicated_composites.html) (and [`VK_EXT_shader_replicated_composites`](https://docs.vulkan.org/refpages/latest/refpages/source/VK_EXT_shader_replicated_composites.html) has pretty bad adoption - according to https://vulkan.gpuinfo.org/listextensions.php - not to mention we'd need user opt-in anyway).

*However*, SPIR-T could take `OpConstantCompositeReplicateEXT` as input, and specifically https://github.com/Rust-GPU/spirt/pull/45 would map it to either undoing the optimization, or optionally re-emitting `OpConstantCompositeReplicateEXT` on the output side if the extension is enabled etc. (and internally SPIR-T could optimize for this situation, but it doesn't really depend on how SPIR-V does any of this).

---

#### What else you should/could do here:

- because `OpConstantNull` supports all of these types, you can go as far as making `memset_const_pattern` return `OpConstantNull` regardless of the type, and not keeping it specific to arrays (and it would also support e.g. `struct`s)
- (optional) for full `SpirvType::Adt` support in `memset_const_pattern` you just need to iterate over field types
- (optional) I'd almost want to say that `constant_composite` itself should check for this situation

eddyb (changes requested):

PR #561, opened 2026-04-13
Fix/implement `checked_{add,sub,mul}` and `saturating_{add,sub}` (using SPIR-V instructions were possible).

eddyb:
After this PR:
- signed `checked_{add,sub}` still have no direct SPIR-V equivalents
  - our custom implementation used to have bugs around e.g. `0 + 0` / `0 - 0`
    (fixed in this PR, and verified via Alive2: [broken](https://alive2.llvm.org/ce/z/Q3Pchi) vs [correct](https://alive2.llvm.org/ce/z/aWvThi))
- unsigned `checked_{add,sub}` and signed/unsigned `checked_mul` map to SPIR-V instructions
  (`OpIAddCarry`, `OpISubBorrow`, `OpSMulExtended`/`OpUMulExtended`)
- `saturating_{add,sub}` use an Intel extension instead of plainly erroring
  (`OpIAddSatINTEL`/`OpISubSatINTEL`)
  - in time, we might choose to emulate these instructions (later in the pipeline)

---

**TODO**: add tests for the edge cases, consider the need for `OpUAddSatINTEL`/`OpUSubSatINTEL`.

PR #587, opened 2026-04-24
difftest: cleanup wgpu runner

Firestar99:
Extracted from https://github.com/Rust-GPU/rust-gpu/pull/334


tests/difftests/tests/lib/src/scaffold/compute/wgpu_runner.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    pub fn with_feature(self, feature: wgpu::Features) -> Self {
+        Self {
+            features: self.features | feature,
+            ..self
+        }
+    }
+
+    pub fn with_push_constant<T: NoUninit>(self, data: &T) -> Self {
```
eddyb: This could also imply `.with_feature(Features::PUSH_CONSTANTS)`, right?
(or `IMMEDIATES`, after the update, rather - should the method also be renamed?)

tests/difftests/tests/lib/src/scaffold/compute/wgpu_runner.rs
```diff
@@ -0,0 +1,275 @@
...
+                                        ty: wgpu::BufferBindingType::Storage { read_only: true },
+                                        has_dynamic_offset: false,
+                                        min_binding_size: None,
+                                    },
+                                    BufferUsage::Uniform => wgpu::BindingType::Buffer {
+                                        ty: wgpu::BufferBindingType::Uniform,
+                                        has_dynamic_offset: false,
+                                        min_binding_size: None,
+                                    },
+                                },
```
eddyb: Looks like the mapping could be `BufferUsage`->`wgpu::BufferBindingType` (only the `ty` field differs).

tests/difftests/tests/lib/src/scaffold/compute/wgpu_runner.rs
```diff
@@ -0,0 +1,275 @@
...
+                    label: Some(&format!("Buffer {i}")),
+                    size: buffer_config.size,
+                    usage,
+                    mapped_at_creation: true,
+                });
+                {
+                    // Zero the buffer.
+                    let initial_data = vec![0u8; buffer_config.size as usize];
+                    let mut mapping = buffer.slice(..).get_mapped_range_mut();
+                    mapping.copy_from_slice(&initial_data);
```
eddyb: `mapping.fill(0)` would also work, assuming it implements `DerefMut<Target = [u8]>`.

tests/difftests/tests/lib/src/scaffold/compute/wgpu_runner.rs
```diff
@@ -0,0 +1,275 @@
...
+            if let Some(buffer) = staging_buffer {
+                let buffer_slice = buffer.slice(..);
+                let (sender, receiver) = futures::channel::oneshot::channel();
+                buffer_slice.map_async(wgpu::MapMode::Read, move |res| {
+                    let _ = sender.send(res);
+                });
+                device.poll(wgpu::PollType::wait_indefinitely())?;
+                block_on(receiver)
+                    .context("mapping canceled")?
+                    .context("mapping failed")?;
```
eddyb: Does the `wgpu` example runner do it this way? Or does it just unwrap the `Result` in the closure passed to `map_async`?

eddyb (changes requested):
LGTM modulo a few small things, anything else blocking this?

tests/difftests/tests/lib/src/scaffold/compute/wgpu_runner.rs
Firestar99: It actually does, that said, I'm not a huge fan of the download code AI wrote so let me refactor it a little. `encoder.map_buffer_on_submit()` is way cleaner as a `map_async()` replacement.

Also noticed it was only ever writing out the first writable buffer, if you had two the second one would just be ignored... I'll just append all the buffers. But I also don't want to spend too much time on this and rather completely refactor the difftest framework *yet again*, will propose my idea on zulip today.

eddyb (approved):
We should've landed this already - feel free to bump wgpu to 30 if it's straightforward.

Firestar99:
Updated it to wgpu 30, will merge when ready.

I also found the branch [`eddyb/checked-saturating-spv-difftest`](https://github.com/Rust-GPU/rust-gpu/tree/eddyb/checked-saturating-spv-difftest) from I guess the discussions of PR https://github.com/Rust-GPU/rust-gpu/pull/561, adding a new test for checked and saturating add, which is failing on current main.

PR #593, opened 2026-05-04
add memset const and dynamic support for f16

39ali:
now you can do something like : ` let arr = [0f16; 64];`


Firestar99 (approved):
Looks good, added a quick test, I know cargo-deny is failing again...

PR #595, opened 2026-05-05
support memset on 128-bit integers

### LINKED ISSUE #594: 128-bit array memset fill
## what - examples

```rs
pub fn new_array_zero_fill_u128() -> [u128; 4] {
    return [0; 4];
}
```

or for *signed* (`i128`) array:

```rs
pub fn new_array_zero_fill_i128() -> [i128; 4] {
    return [0; 4];
}
```

## why

As an experiment I have been able to *build* a "shader" that imports & uses `chacha20` from RustCrypto, with many (many) local workarounds that I would like to discuss in the near future. The first build issue that I encountered was this:

```
   Compiling cipher v0.5.1
error: memset on integer width 128 not implemented yet
   --> /Users/brody-dev/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/crypto-common-0.2.1/src/hazmat.rs:347:1
    |
347 | / impl_serializable_state_u128_array! {
348 | |     U1,
349 | |     U2,
350 | |     U3,
...   |
362 | |     U512
363 | | }
    | |_^
```

While it would be possible for me to continue building by removing or feature-gating the `impl_serializable_state_u128!` usage in `crypto-common`, I think it would be much easier if I could continue building without this kind of hackage.

While I would love to see this resolved for both signed & unsigned 128-bit arrays, the unsigned case is more important for me at this point.

## related issues

- #359
- #307

## potential workaround

- #595

brody2consult:
__resolves #594__

This is a quick solution to support the case I described in #594:

```rs
pub fn new_array_zero_fill_u128() -> [u128; 4] {
    return [0; 4];
}
```

This proposal would help me get past the first build issue I have encountered with RustCrypto.

I would love to support non-zero memset fill for both `u128` & `i128` as discussed in #594, but this would involve some more work in testing to hopefully ensure that this is done correctly & not likely broken by potential future refactoring.

_I would be happy to address any possible feedback on my first contribution to Rust-GPU._

### TODO

- [x] ~~squash commits before merging - Reviewers please let me know if you want me to do this or if you want to do this yourself~~

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -308,6 +308,11 @@ fn memset_fill_u64(b: u8) -> u64 {
         | ((b as u64) << 56)
 }
 
+fn memset_fill_u128(b: u8) -> u128 {
+    let b64 = memset_fill_u64(b) as u128;
+    b64 | b64 >> 64
+}
+
 fn memset_dynamic_scalar(
     builder: &mut Builder<'_, '_>,
     fill_var: Word,
@@ -387,6 +392,9 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                 64 => self
                     .constant_u64(self.span(), memset_fill_u64(fill_byte))
                     .def(self),
+                128 => self
+                    .constant_u128(self.span(), memset_fill_u128(fill_byte))
+                    .def(self),
                 _ => self.fatal(format!(
                     "memset on integer width {width} not implemented yet"
                 )),
@@ -401,6 +409,9 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                     .def(self),
                 64 => self
                     .constant_i64(self.span(), memset_fill_u64(fill_byte) as i64)
+                    .def(self),
+                128 => self
+                    .constant_u128(self.span(), memset_fill_u128(fill_byte))
                     .def(self),
                 _ => self.fatal(format!(
                     "memset on integer width {width} not implemented yet"
--- a/crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
@@ -50,6 +50,10 @@ impl<'tcx> CodegenCx<'tcx> {
         self.constant_int_from_native_unsigned(span, val)
     }
 
+    pub fn constant_u128(&self, span: Span, val: u128) -> SpirvValue {
+        self.constant_int_from_native_unsigned(span, val)
+    }
+
     fn constant_int_from_native_unsigned(&self, span: Span, val: impl Into<u128>) -> SpirvValue {
         let size = Size::from_bytes(std::mem::size_of_val(&val));
         let ty = SpirvType::Integer(size.bits() as u32, false).def(span, self);
--- a/tests/compiletests/ui/lang/core/array/init_array_i128.rs
+++ b/tests/compiletests/ui/lang/core/array/init_array_i128.rs
@@ -1,0 +1,14 @@
+// build-pass
+
+use spirv_std::spirv;
+
+pub fn u128_constant() -> [u128; 4] {
+    return [0x123456789ABCDEF0123456789ABCDEF; 4];
+}
+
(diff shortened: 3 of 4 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -308,6 +308,10 @@ fn memset_fill_u64(b: u8) -> u64 {
         | ((b as u64) << 56)
 }
 
+fn zero_fill_u128() -> u128 {
+    0 as u128
```
brody2consult: - need to do `0_u128` to keep Clippy happy
- I think there is no need to keep an extra function for this hopefully temporary hack

tests/compiletests/ui/lang/core/array/new_array_zero_fill_u128.rs
brody2consult, line 1: I would like to rename this file due to the amount of difference from the other init_array compile tests

brody2consult (commented):
A couple nits - I am already testing a resolution in personal PR: https://github.com/brody2consult/rust-gpu/pull/1 - now resolved ✅

I will push the updates to *this* PR if I do not encounter any failures in https://github.com/brody2consult/rust-gpu/pull/1 - now done ✅

Firestar99:
I've generalized it to work with arbitrary bitpatterns and on signed and unsigned 128 bit integers. If you could confirm this patch works for you, I'd be happy to merge it.

I wrote this into the compiletest:
> `crypto-common` has a proc macro to implement code for u8, u16, u32, u16 and u128, which means we need our codegen
> to handle the u128 case for the crate to even compile. Specifically, this calls `memset_const_pattern` with u128.
> 
> Note that the u128 path is unusable anyway as 128bit integer are not supported in SPIR-V. So if this is used nowhere
> in any entry point, our post-link DCE will remove the u128 code and the SPIR-V will be valid.
> <https://github.com/RustCrypto/traits/blob/f896a76d4468f1ff855b651124fd5b5378a3417f/crypto-common/src/hazmat.rs#L61-L83> 

Generally, we do accept patches that do enable you to use crates that previously didn't compile, even if you can't use the u128bit path. In similar vein, we've added `black_box` support by just doing a noop, since we can't actually provide a proper black box in SPIR-V.

brody2consult (commented):
A couple minor nits in the test comments

---

Looks like the GitHub web ui did not take my comments ... no worries I will make the update from my laptop. I think there should be a hyphen in "128bit integer":

```diff
-/// Note that the u128 path is unusable anyway as 128bit integer are not supported in SPIR-V. So if this is used nowhere
+/// Note that the u128 path is unusable anyway as 128-bit integer are not supported in SPIR-V. So if this is used nowhere
```

brody2consult:
I just force-pushed another update to update the test comments, use "128-bit" with hyphen in the title, and add you as co-author. I changed the link to link to the code that actually failed to compile. Good to go from my end assuming that the tests continue to succeed.

__I have also updated the PR description to resolve issue #594.__

I do think it would be ideal to show a build warning in case of compiling with non-working functionality, in a similar fashion to "warn on large non-zero fills" as proposed in PR #586. But I would be happy to defer this for now.

I did try the following code mutations which do not trigger any failures in compiletests:

```diff
diff --git a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
index 1bea4fedfe..29667d8ed3 100644
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -309,8 +309,8 @@ fn memset_fill_u64(b: u8) -> u64 {
 }
 
 fn memset_fill_u128(b: u8) -> u128 {
-    let b64 = memset_fill_u64(b) as u128;
-    b64 | b64 >> 64
+    let b64 = memset_fill_u32(b) as u128;
+    b64 >> 20
 }
 
 fn memset_dynamic_scalar(
```

I don't think this should be any issue since this is non-working functionality anyways. But I had wanted to find a way to test my own work on this as discussed in #596.

Thanks for your quick handling of this contribution; I am extremely happy overall.

Firestar99 (approved):

PR #609, opened 2026-06-08
fix glam 0.33 breakage

### LINKED ISSUE #602: spirv-std incompatible with shipped glam version
I have added the `spirv-builder` crate to my project, like described in the book. Then I created a workspace subcrate and added the `spriv-std` as a dependency. Due to limitations I need to work with the 2024 edition of rust, so I chose to use the `0.10.0-alpha.1` version of both. When compiling the minimal shader example from the examples folder I got following errors from the Spirv builder (from multiple source locations):

```
warning: vulkanotest2@0.1.0: error[E0432]: unresolved import `glam::UVec2`
warning: vulkanotest2@0.1.0:   --> ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/spirv-std-0.10.0-alpha.1/src/arch.rs:11:5
warning: vulkanotest2@0.1.0:    |
warning: vulkanotest2@0.1.0: 11 | use glam::UVec2;
warning: vulkanotest2@0.1.0:    |     ^^^^^^^^^^^ no `UVec2` in the root
warning: vulkanotest2@0.1.0:    |
warning: vulkanotest2@0.1.0: help: a similar name exists in the module
warning: vulkanotest2@0.1.0:    |
warning: vulkanotest2@0.1.0: 11 - use glam::UVec2;
warning: vulkanotest2@0.1.0: 11 + use glam::Vec2;
warning: vulkanotest2@0.1.0:    |
warning: vulkanotest2@0.1.0: 
```

my shader `Cargo.toml` has following dependencies:
```toml
[dependencies]
bytemuck = { version = "1.25.0", features = ["derive"] }
spirv-std = { version = "0.10.0-alpha.1", features = ["bytemuck"] }
```

Is this a known error from the alpha version or just me missing something?

Firestar99:
glam 0.33 [introduced a feature for each type of vector to speed up compile times](https://github.com/bitshifter/glam-rs/blob/main/CHANGELOG.md#0330---2026-05-21). This caused [rust-gpu to break when used with glam 0.33](https://github.com/Rust-GPU/rust-gpu/issues/602), which is completely our fault, since we declare [our dependency on glam to be forwards compatible even across breaking changes](https://github.com/Rust-GPU/rust-gpu/blob/36e3348cdc2f824afec64b3b5af5d369d98a4c0d/Cargo.toml#L96):
```toml
glam = { version = ">=0.30.8", default-features = false }
```

The simple fix is to just require glam 0.33, like the second commit does:
```diff
-glam = { version = ">=0.30.8", default-features = false }
+glam = { version = "0.33.0", default-features = false, features = ["u32", "i32", "f64"] }
```

However, there are a few important libraries that still depend on an older glam version, like the most recent release of [bevy_math still depending on `glam v0.30.*`](https://crates.io/crates/bevy_math/0.18.1/dependencies). 

Instead, I would like to propose that we have one feature for each glam version, and default to the most recent one:
```toml
[features]
default = ["glam_0_33"]
glam_0_33 = ["dep:glam_0_33"]
glam_0_32 = ["dep:glam_0_32"]
glam_0_31 = ["dep:glam_0_31"]
glam_0_30 = ["dep:glam_0_30"]
```
And within `lib.rs` we'll switch on the feature gate to select the glam version:
```rust
#[cfg(feature = "glam_0_30")]
pub use ::glam_0_30 as glam;
#[cfg(feature = "glam_0_31")]
pub use ::glam_0_31 as glam;
#[cfg(feature = "glam_0_32")]
pub use ::glam_0_32 as glam;
#[cfg(feature = "glam_0_33")]
pub use ::glam_0_33 as glam;
```

Note that this does *not* support multiple glam versions simultaneously, but I don't think that matters. Maintaining this is cheap, so we can just maintain backwards compatible for whatever N versions that we can be bothered to.

Expect a crates alpha release soon after merging this.

close https://github.com/Rust-GPU/rust-gpu/issues/602

### DIFF
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -116,7 +116,24 @@ pub use scalar_or_vector::*;
 pub use typed_buffer::*;
 pub use vector::*;
 
-pub use glam;
+#[cfg(feature = "glam_0_30")]
+pub use ::glam_0_30 as glam;
+#[cfg(feature = "glam_0_31")]
+pub use ::glam_0_31 as glam;
+#[cfg(feature = "glam_0_32")]
+pub use ::glam_0_32 as glam;
+#[cfg(feature = "glam_0_33")]
+pub use ::glam_0_33 as glam;
+
+#[cfg(not(any(
+    feature = "glam_0_33",
+    feature = "glam_0_32",
+    feature = "glam_0_31",
+    feature = "glam_0_30"
+)))]
+compile_error!(
+    "Crate `spirv-std` needs at least one one of it's `glam*` features enabled to select which glam version to use"
+);
 
 #[cfg(all(not(test), target_arch = "spirv"))]
 #[panic_handler]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1449,9 +1449,9 @@ dependencies = [
 
 [[package]]
 name = "glam"
-version = "0.31.0"
+version = "0.33.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "74a4d85559e2637d3d839438b5b3d75c31e655276f9544d72475c36b92fabbed"
+checksum = "898f5a568a84989b6c0f8caa50a93074b97dbdc58fc6d9543157bb4562758933"
 dependencies = [
  "bytemuck",
  "libm",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,23 +54,21 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "android-activity"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef6978589202a00cd7e118380c448a08b6ed394c3a8df3a430d0898e3a42d046"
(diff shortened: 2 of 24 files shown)


LegNeato (approved):
Agreed this is a good fix.

Can we put raw glam in `cargo-deny` w/ an override for so we can't use it directly w/o CI failing?

nazar-pc:
Docs failed to build with these changes: https://github.com/Rust-GPU/rust-gpu/actions/runs/27157018177/job/80162386469

crates/spirv-std/src/lib.rs
```diff
@@ -116,7 +116,24 @@ pub use scalar_or_vector::*;
...
+
+#[cfg(not(any(
+    feature = "glam_0_33",
+    feature = "glam_0_32",
+    feature = "glam_0_31",
+    feature = "glam_0_30"
+)))]
+compile_error!(
+    "Crate `spirv-std` needs at least one one of it's `glam*` features enabled to select which glam version to use"
+);
```
nazar-pc, line 136: Using [`cfg_select!`](https://doc.rust-lang.org/stable/core/macro.cfg_select.html) macro for selecting the version of glam to import would solve docs issue in much more natural and ergonomic way.

Simply select the newest version first with fallbacks to the older version and the wildcard branch at the end will produce compiler error if user didn't select anything.

skrobchik:
> Expect a crates alpha release soon after merging this.

Could you please publish a new alpha on crates.io @Firestar99 ? 🥺

PR #610, opened 2026-06-08
update to nightly-2026-05-22, rustc 1.97.0

### LINKED ISSUE #605: Update Rust toolchain
Generics in latests nightlies with `generic_const_args` require `const {}` rather than `{}` in places like this:
```rust
atomic_i_increment::<_, { Scope::QueueFamily as u32 }, { Semantics::NONE.bits() }>(..)
```

This makes writing code that compiles both with `rust-gpu`'s old toolchain version and latest nightly very frustrating.

Please consider updating toolchain again some time soon 🙏

Firestar99:
just minor changes were needed:
* target specs removed "allows-weak-linkage" key
* added a call to `.skip_norm_wip()` which will be removed in the future. Seems like rustc is getting some kind of type normalization (by Oli?), which currently is not yet active, but a future toolchain update may need to look into this some more.
* `write_mir_pretty` isn't filterable anymore, replaced it with some custom code, so output of `--dump-mir` may be slightly different

close https://github.com/Rust-GPU/rust-gpu/issues/605

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -253,7 +253,7 @@ jobs:
           #          - rust-gpu-version: bbb61f58b3d24f3f64745050eb214b90bf6dcce9
           #            glam-version: 0.30.7
 
-          # testing rustc 1.5 months later
+          # testing rustc 1.79.0, 1.5 months later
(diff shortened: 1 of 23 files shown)


LegNeato (approved):
🚀

PR #611, opened 2026-06-08
ci: fix docs script

Firestar99:
> Docs failed to build with these changes: https://github.com/Rust-GPU/rust-gpu/actions/runs/27157018177/job/80162386469

https://github.com/Rust-GPU/rust-gpu/pull/609#issuecomment-4651962049

https://github.com/Rust-GPU/rust-gpu/pull/609 added mutually exclusive features (`glam_0_33`, `glam_0_32`, etc.) and the script passes `--all-features`, which ofc fails... so let's just do default features

PR #613, opened 2026-06-09
allow multiple glam features to be active at once, prefer highest glam version

Firestar99:
Alternative to https://github.com/Rust-GPU/rust-gpu/pull/611 thanks to @nazar-pc


LegNeato (approved):
Oooh, the new cfg-select stuff is nice!

PR #619, opened 2026-06-23
Collapse `#[repr(transparent)]` types

Firestar99:
Supersedes https://github.com/Rust-GPU/rust-gpu/pull/566

### DIFF
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -4,7 +4,7 @@
 use crate::attr::{AggregatedSpirvAttributes, IntrinsicType};
 use crate::codegen_cx::CodegenCx;
 use crate::maybe_pqp_cg_ssa::traits::ConstCodegenMethods as _;
-use crate::spirv_type::SpirvType;
+use crate::spirv_type::{SpirvType, name_type_id};
 use itertools::Itertools;
 use rspirv::spirv::{Dim, ImageFormat, StorageClass, Word};
 use rustc_abi::ExternAbi as Abi;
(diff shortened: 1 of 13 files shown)


LegNeato (changes requested):
This looks directionally good to me.

In`sole_structural_newtype_field` it talks about “one non-ZST field”, but `repr(transparent)` is more specific...ignored fields must be both size 0 _and_ align 1. Looks like we are only checking `is_zst()`?

Also a few more projection tests might be good, something like:

```rust
#[repr(transparent)]
struct W(Inner);
```

and then use W through &W, &mut W, raw pointers, arrays/slices, function args/returns, and nested transparent wrappers. Mostly want to make sure that once the wrapper type gets collapsed, accessing .0 still becomes a no-op and we don't accidentally emit an invalid access chain.

Might worth adding a test for an over-aligned ZST too:

```rust
#[repr(align(8))]
struct Blah;

#[repr(transparent)]
struct W {
    inner: u32,
    marker: Blah,
}
```

`marker` is size 0 but it has alignment 8 so it can affect `W`’s layout.

Requesting changes, but feel free to put up again if you are merely extracting and don't want to change anything.

Firestar99:
Rewrote all the tests, seems like all the GEPs are emitted just fine

LegNeato:
@eddyb please take a look.

eddyb (approved):
Oh, this is so easy now, thanks to `SpirvType` not having to match the `rustc`-side layout (other than having fields at the right offsets).

I've wanted to do this in the past, but the existence of e.g. `struct_gep` was blocking it (seems like that has been gone since https://github.com/rust-lang/rust/commit/beed25be9a2a9e1e73ca79210da57d294c27f925 huh).

My only worry now is whether we have code acting on `Ty`/layout, and `SpirvType`, *simultaneously*, and expecting any connection between the two.

Firestar99:
Forge-merging since @LegNeato's review is blocking the merge

nazar-pc:
Interesting, does this suffer from the same issue?:
```rust
#[repr(C)]
pub struct X(u32);
```

If so, then it should probably be treated the same way as `#[repr(transparent)]`, at least for simple cases like this one.

Firestar99:
I wasn't going to do it with `#[repr(C)]` cause removing that struct "layer" will influence inline assembly, as GEPs (aka OpAccessChain) need one index less. Though I honestly don't know if there's a situation where this is actually problematic.

nazar-pc:
I'm sure a general case is more involved, but I'd expect for trivial integers the result should be identical. This PR got me thinking if I'm leaving performance on the table by complicating the code with `#[repr(C)]` new types.

PR #622, opened 2026-06-26
add file lock to coordinate racing cargo gpu installs

### LINKED ISSUE #621: Concurrent SPIR-V codegen building causes issues
When building multiple crates that depend on SPIR-V codegen, compilation can, unfortunately, fail.

The error is not very descriptive, but I think it is caused by some kind of race:
```
  Error: `rustc_codegen_spirv` build failed
```

https://github.com/nazar-pc/abundance/actions/runs/28211711059/job/83574463880

It'd be great if necessary locking was used such that this doesn't cause issues.

The reason I'm now hitting it is because I'm using [parallel steps](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/) in CI with different `target` directories concurrently, so SPIR-V codegen apparently ends up being compiled multiple times unnecessarily (it is normally cached, hence only needs to be compiled once) and that process breaks somewhere along the way.

Firestar99:
* tested locally via cmdline and project with two shader crates competing

close https://github.com/Rust-GPU/rust-gpu/issues/621

### DIFF
--- a/crates/cargo-gpu-install/src/install.rs
+++ b/crates/cargo-gpu-install/src/install.rs
@@ -1,7 +1,7 @@
 //! Install a dedicated per-shader crate that has the `rust-gpu` compiler in it.
 
 use crate::spirv_source::{CrateMetadata, get_channel_from_rustc_codegen_spirv_build_script};
-use crate::{cache_dir, spirv_source::SpirvSource};
+use crate::{cache_dir, spirv_source::SpirvSource, user_output};
 use anyhow::Context as _;
 use spirv_builder::SpirvBuilder;
 use std::path::{Path, PathBuf};
@@ -243,6 +243,7 @@ package = "rustc_codegen_spirv"
             self.spirv_builder_version.as_deref(),
         )?;
         let install_dir = source.install_dir()?;
+        std::fs::create_dir_all(&install_dir)?;
 
         let dylib_filename = format!(
             "{}rustc_codegen_spirv{}",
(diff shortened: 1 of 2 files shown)


LegNeato (changes requested):
File locking in notorious to get right with all the edge cases...perhaps we should use https://crates.io/crates/fd-lock or https://crates.io/crates/gix-lock or https://crates.io/crates/fslock ?

Firestar99:
My read was that many of those crates were made before file locking was [added to the standard library](https://releases.rs/docs/1.89.0/#stabilized-apis) August last year. Since them, even [cargo](https://doc.rust-lang.org/stable/nightly-rustc/src/cargo/util/flock.rs.html#230-244) adopted the functions from std. 

The bit of custom `struct FileLock` I build is merely for pretty printing a message when it'll wait for the lock.

LegNeato (approved):

nazar-pc:
This certainly works better than before, but I think there might still be something not quite handled. I occasionally get this in CI:
```
  Error: ensuring toolchain and components exist

  Caused by:
      Command `rustup component list --toolchain nightly-2026-05-22-aarch64-apple-darwin -q --installed` failed with exit status: 1:
      -- stdout
      
      -- stderr
      error: Missing manifest in toolchain 'nightly-2026-05-22-aarch64-apple-darwin'
```

https://github.com/nazar-pc/abundance/actions/runs/29024559956/job/86141194982?pr=698

It doesn't happen every time, but enough to be annoying.

PR #623, opened 2026-06-28
Add Git information to cargo-gpu's `--version` CLI subcommand

Steveplays28:
Old:

```bash
$ cargo-gpu --version
cargo-gpu 0.10.0-alpha.1
```

New:

```bash
$ cargo-gpu --version
cargo-gpu 0.10.0-alpha.1 installed from crates.io (rev d95ec972)

$ cargo-gpu --version
cargo-gpu 0.10.0-alpha.1 installed from repo at rev d95ec972 (2026-07-16)

$ cargo-gpu --version
cargo-gpu 0.10.0-alpha.1 installed from unknown source
```

### DIFF
--- a/crates/cargo-gpu/build.rs
+++ b/crates/cargo-gpu/build.rs
@@ -1,9 +1,26 @@
 //! cargo-gpu build script.
 
+use serde_json::Value;
 use std::ffi::OsStr;
 use std::path::PathBuf;
 
 fn main() {
+    compute_version_extra();
+}
+
+fn compute_version_extra() {
+    let extra = version_crates_io()
+        .or_else(version_git)
+        .unwrap_or(String::from(" installed from unknown source"));
+
+    std::fs::write(
+        PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("version_extra"),
+        extra,
+    )
+    .unwrap();
+}
+
+fn version_git() -> Option<String> {
     if let Some(git_directory) = invoke_git(["rev-parse", "--git-dir"]) {
         println!("cargo:rerun-if-changed={git_directory}/HEAD");
     }
@@ -17,20 +34,15 @@ fn main() {
         "HEAD",
     ]);
 
-    let extra = if let Some(git_rev) = git_rev
+    if let Some(rev) = git_rev
         && let Some(git_date) = git_date
     {
-        let git_rev = git_rev.get(..8).unwrap_or(&git_rev);
+        let rev = rev.get(..8).unwrap_or(&rev);
         let git_date = git_date.strip_suffix("\n").unwrap_or(&git_date);
-        format!(" ({git_rev} {git_date})")
+        Some(format!(" installed from repo at rev {rev} ({git_date})"))
     } else {
-        " (installed from crates.io)".into()
-    };
-    std::fs::write(
-        PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("version_extra"),
-        extra,
-    )
-    .unwrap();
+        None
+    }
 }
 
 fn invoke_git<I, S>(args: I) -> Option<String>
@@ -43,4 +55,16 @@ where
         .output()
         .ok()
         .and_then(|output| String::from_utf8(output.stdout).ok())
+}
+
+/// `cargo package` adds the `.cargo_vcs_info.json` file on it's own, let's just use that to resolve the rev
+fn version_crates_io() -> Option<String> {
+    let json = std::fs::read_to_string(
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join(".cargo_vcs_info.json"),
+    )
+    .ok()?;
+    let value = serde_json::from_str::<Value>(&json).ok()?;
+    let rev = value.get("git")?.get("sha1")?.as_str()?;
+    let rev = rev.get(..8).unwrap_or(rev);
+    Some(format!(" installed from crates.io (rev {rev})"))
 }
--- a/crates/cargo-gpu/build.rs
+++ b/crates/cargo-gpu/build.rs
@@ -1,12 +1,33 @@
 //! cargo-gpu build script.
 
+use std::ffi::OsStr;
+
 fn main() {
-    let git_hash = std::process::Command::new("git")
-        .args(["rev-parse", "HEAD"])
-        .output()
-        .map_or_else(
-            |_| "unknown".to_owned(),
-            |output| String::from_utf8(output.stdout).unwrap_or_else(|_| "unknown".to_owned()),
-        );
+    let git_directory = invoke_git(["rev-parse", "--git-dir"]);
+    println!("cargo:rerun-if-changed={git_directory}/HEAD");
+
+    let git_hash = invoke_git(["rev-parse", "HEAD"]);
     println!("cargo:rustc-env=GIT_HASH={git_hash}");
+
+    let git_date = invoke_git([
+        "show",
+        "-s",
+        "--format=%cd",
+        "--date=format:%Y-%m-%d",
+        "HEAD",
+    ]);
+    println!("cargo:rustc-env=GIT_DATE={git_date}");
+}
+
+fn invoke_git<I, S>(args: I) -> String
+where
+    I: IntoIterator<Item = S>,
+    S: AsRef<OsStr>,
+{
+    std::process::Command::new("git")
+        .args(args)
+        .output()
+        .ok()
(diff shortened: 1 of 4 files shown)


nazar-pc (changes requested):
It is just my opinion, but it seems really wasteful to add such a massive dependency tree for such a tiny feature.

You essentially added more than 1000 lines of dependencies just to obtain `VERGEN_GIT_COMMIT_DATE` environment variable, which doesn't make sense to me. Everything else was already there.

Adding commit hash is a single line change instead of current +1346 -80, adding date is a bit more than that, but still less than what this PR does.

Steveplays28:
@nazar-pc well okay, fair.
I'll keep the Git CLI invocation that was there before

Steveplays28:
I added a helper function for invoking Git, added `cargo:rerun-if-changed` lines and got the date through the Git CLI. Thanks for checking the PR, Nazar.

My bad for not doing it in this simple way in the first place, I agree that this is better. Less time needed to compile and less code to maintain

crates/cargo-gpu/build.rs
```diff
@@ -1,12 +1,34 @@
...
-            |output| String::from_utf8(output.stdout).unwrap_or_else(|_| "unknown".to_owned()),
+	let git_directory = invoke_git(["rev-parse", "--git-dir"]);
+	println!("cargo:rerun-if-changed={git_directory}/HEAD");
+    println!("cargo:rerun-if-changed={git_directory}/packed-refs");
+    if let Ok(git_head) = std::fs::read_to_string(format!("{git_directory}/HEAD")) && let Some(git_ref) = git_head.strip_prefix("ref: ") {
+        println!(
+            "cargo:rerun-if-changed={git_directory}/{}",
+            git_ref.trim()
         );
+    }
```
nazar-pc: Are you sure these `rerun-if-changed` are actually needed?
If installed from git repository, the revision will have to change, and the whole crate will be recompiled anyway.

I'm not sure what is the situation where these are actually necessary.

nazar-pc (commented):
This looks much better, thanks 🙂

crates/cargo-gpu/build.rs
Steveplays28: from what I read on StackOverflow it is but I haven't tested without those `rerun-if-changed` lines :sweat_smile:

Steveplays28: I just tested it, if I make a commit without those `rerun-if-changed` lines, then re-compile with e.g. `cargo run -p cargo-gpu -- --version`, `COMMIT_HASH` doesn't update
so you could get a situation where you think the program is compiled from an old commit when it isn't

nazar-pc: You already know what version you have locally, so I don't think it matters in that case. And if it is not local, then it is installed from git repo and will be treated as a separate crate anyway.

Steveplays28: I still think it would be a bug to leave out `cargo:rerun-if-changed`. If you were to run `cargo build` and copy the binary for local use, it would contain an incorrect commit hash and date.
It could be confusing to someone in the future.

Firestar99 (commented):
I'm not yet sure if this is a net positive, would love input from others.

Some unordered thoughts:
* I didn't notice `cargo gpu` on main already stored the commit rev, see `cargo gpu show commitsh`. Not very discoverable, definitely prefer putting it into the version.
* a cargo package installed from crates.io won't have a git repo, so the git rev will be "unknown"
  * `cargo nextest --version` also prints the git rev, but also suffers from the same issue when installed via `cargo install`
  * there doesn't seem to be a nice universal way to handle this
* setting a `GIT_HASH` env var will force a larger recompile on each rev change
* `cargo-deny` ci is broken, see https://github.com/Rust-GPU/rust-gpu/pull/624

nazar-pc:
My assumption is that this is primarily useful for identifying what version is used when reporting bugs. Versions published on crates.io are well known, while those installed from git may have the same version, but correspond to any revision.

So it is fine to have `unknown` version for crates.io (I'd probably not set revision at all rather than setting to `unknown` though) and actual hash when installed from git. And for the same reason it is fine to not have anything when installed from local path since it is something developer will hopefully specify explicitly (they're probably testing some local patches in that case).

crates/cargo-gpu/build.rs
```diff
@@ -1,12 +1,39 @@
...
-            |output| String::from_utf8(output.stdout).unwrap_or_else(|_| "unknown".to_owned()),
-        );
+    let git_directory = invoke_git(["rev-parse", "--git-dir"]);
+    println!("cargo:rerun-if-changed={git_directory}/HEAD");
+    println!("cargo:rerun-if-changed={git_directory}/packed-refs");
+    if let Ok(git_head) = std::fs::read_to_string(format!("{git_directory}/HEAD"))
+        && let Some(git_ref) = git_head.strip_prefix("ref: ")
+    {
+        println!("cargo:rerun-if-changed={git_directory}/{}", git_ref.trim());
+    }
```
Firestar99: Wouldn't HEAD be sufficient?

Firestar99 (commented):
I'll merge it once the comment is addressed.

I wouldn't go to the length of writing the git rev to a file on release for `cargo install` from crates.io to include it, seems like a lot of effort without much gain. [nextest](https://github.com/nextest-rs/nextest/blob/main/cargo-nextest/build.rs) actually implements such a system, yet I don't see the git rev so I assume it broke at some point and noone noticed or cared to fix it.

crates/cargo-gpu/build.rs
Steveplays28: as far as I read, that would miss some edge cases

Firestar99: ... Which edge cases? Any examples?

Steveplays28: I looked into it, apparently `packed-refs` is just for compression (makes sense in hindsight), so you're right Firestar, we can get rid of that
It's changed when running `git gc` and `git maintenance`, but we shouldn't care about that because the commit hash and date don't change

and I also checked if the `ref: ` part is necessary, but it's not, even when switching branches. I've removed both

Steveplays28:
> My assumption is that this is primarily useful for identifying what version is used when reporting bugs. Versions published on crates.io are well known, while those installed from git may have the same version, but correspond to any revision.

yeah, I recently brought up an issue on the Discord server that was just due to an out-of-date cargo-gpu. I thought I'd updated it, but I hadn't

> So it is fine to have unknown version for crates.io (I'd probably not set revision at all rather than setting to unknown though) and actual hash when installed from git. And for the same reason it is fine to not have anything when installed from local path since it is something developer will hopefully specify explicitly (they're probably testing some local patches in that case).

I think I agree, maybe the `unknown` commit hash and date shouldn't be set/shown when cargo-gpu is installed from crates.io
I could change that if you want?

eddyb (approved):

Firestar99:
I gave it one more pass made it print this:
```
$ cargo install --path .
$ cargo gpu -V
cargo-gpu 0.10.0-alpha.1 (42a53815 2026-07-13)
$ cargo install cargo-gpu@...
$ cargo gpu -V
cargo-gpu 0.10.0-alpha.1 (installed from crates.io)
```

Changes:
* always print rev, don't require long `--version`
* short rev
* explicitly state when installed from crates.io

Thoughts? (a :+1: if you're happy is fine)

I would also like to make a MSRV-like system but for cargo-gpu, to notify you that you need to upgrade cargo-gpu before proceeding with the build and failing. (In another PR, not here)

Steveplays28:
looks good @Firestar99, thanks

also that cargo-gpu minimum version system would be great

Firestar99:
Small detail I just noticed: When `cargo package` creates a package, it adds a file at `./.cargo_vcs_info.json` with the following contents:
```json
{
  "git": {
    "sha1": "2dd42e970faff3e6f8b1c8c4e8d2daf930c5dd04"
  },
  "path_in_vcs": "crates/cargo-gpu"
}
```
We could read the rev from there, without needing some special release script...

Firestar99:
Made one more iteration:
```
cargo-gpu 0.10.0-alpha.1 installed from crates.io (rev d95ec972)
cargo-gpu 0.10.0-alpha.1 installed from repo at rev d95ec972 (2026-07-16)
cargo-gpu 0.10.0-alpha.1 installed from unknown source
```

Steveplays28:
thanks @Firestar99!

PR #631, opened 2026-07-20
lock rustup across concurrent toolchain installs

### LINKED ISSUE #629: Toolchain locking is still buggy
This certainly works better than before, but I think there might still be something not quite handled. I occasionally get this in CI:
```
  Error: ensuring toolchain and components exist

  Caused by:
      Command `rustup component list --toolchain nightly-2026-05-22-aarch64-apple-darwin -q --installed` failed with exit status: 1:
      -- stdout
      
      -- stderr
      error: Missing manifest in toolchain 'nightly-2026-05-22-aarch64-apple-darwin'
```

https://github.com/nazar-pc/abundance/actions/runs/29024559956/job/86141194982?pr=698

It doesn't happen every time, but enough to be annoying.

_Originally posted by @nazar-pc in https://github.com/Rust-GPU/rust-gpu/issues/622#issuecomment-4927794445_

LegNeato:
Fixes #629

### DIFF
--- a/crates/cargo-gpu-install/src/install.rs
+++ b/crates/cargo-gpu-install/src/install.rs
@@ -385,11 +385,15 @@ pub struct FileLock(std::fs::File);
 
 impl FileLock {
     pub fn lock(path: &Path) -> std::io::Result<Self> {
+        Self::lock_with_message(path, "codegen backend build")
+    }
+
+    pub fn lock_with_message(path: &Path, what: &str) -> std::io::Result<Self> {
         let file = std::fs::File::create(path)?;
         match file.try_lock() {
             Ok(_) => (),
             Err(std::fs::TryLockError::WouldBlock) => {
-                user_output!("Waiting for file lock on codegen backend build...\n");
+                user_output!("Waiting for file lock on {what}...\n");
                 file.lock()?;
             }
             Err(std::fs::TryLockError::Error(e)) => return Err(e),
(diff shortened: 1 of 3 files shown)


nazar-pc (commented):
I've done enough runs with 86c1bd1595bc5ed6b27e80f63bbfa739a804e394 to be fairly confident this PR fixes the issue

LegNeato:
OK, going to merge this without a review as it is causing so much pain. We can fix forward any issues.

crates/cargo-gpu-install/src/install.rs
Firestar99, line 317: I don't see how this PR is improving the situation. Top of this comment range is acquiring the file lock, and afterwards we call `ensure_toolchain_and_components_exist`, within which this PR adds another lock.

@nazar-pc 's CI only uses a single rust-gpu version, so it can't be two rust-gpu installs with separate locks running rustup in parallel either.

nazar-pc: I do use a single version, but I have multiple `target` directories used concurrently, which is most likely how I triggered it. I have not had a single failure since this PR was merged, while previously I was hitting it all the time.

Firestar99: I wonder if rustup races even when just querying about installed components and such

nazar-pc: I believe so, but that is going to change (thankfully) in the future: https://blog.rust-lang.org/inside-rust/2026/07/03/rustup-update-1.30/

PR #632, opened 2026-07-30
rspirv2 disassembly

Firestar99:
We released [rspirv2](https://github.com/Rust-GPU/rspirv2/) :tada: 

* best reviewed commit by commit
* last commit "switch dis preset to default" has a ton of bless changes that don't need to be reviewed, since I'm only changing the disassembly settings for prettier disassembly
* rather, review the blesses in "replace disassembler*" where actual differences between rspirv and rspirv2 are visible
* while you're reviewing this, please also approve https://github.com/Rust-GPU/rust-gpu/pull/533


crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -747,10 +747,29 @@ impl CodegenArgs {
             remap.len() as u32 + 1
         }
 
-        use rspirv::binary::Disassemble;
+        use rspirv::binary::{Assemble, Disassemble};
```
LegNeato: rspirv2 doesn't have its own traits here?

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -747,10 +747,29 @@ impl CodegenArgs {
...
+        fn dis(insts: &(impl Assemble + Disassemble)) -> impl Display {
+            use rspirv2::core::inst_set::CoreInstSet;
+            use rspirv2::dis::DisOptions;
+            use rspirv2::module::Module;
+            use rspirv2::operand::LiteralStringEscape;
+
+            let words = insts.assemble();
+            let module =
+                Module::<CoreInstSet>::from_words_maybe_header(bytemuck::cast_slice(&words))
+                    .unwrap();
```
LegNeato: Return error rather than unwrapping?

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -747,10 +747,29 @@ impl CodegenArgs {
...
+            module
+                .dis(DisOptions {
+                    color: false,
```
LegNeato: Color should be taken from the session context's `--color` value

LegNeato (changes requested):

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
Firestar99: I assume you mean the function arg `impl Assemble`? That just needs to be rspirv's `Assemble` (can even drop `Disassemble`) to accept both rspirv's `Module` (compiletests with `disassemble` and `disassemble-global`) or `Function` (`disassemble-function`) as args. And turn those into bytes that rspirv2 can then "parse" (since it decodes on the fly, it doesn't actually parse it) and disassemble.

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
Firestar99: Nope we can't. compiletest-rs would compare our output with ascii sequences to the `*.stderr` file without any, and fail every test. (Though it does print pretty colors while doing so :D )

LegNeato (approved):
Ah, I see what is going on now.

Firestar99:
Wrong lockfiles *twice*...

Issue #607, opened 2026-05-31
Consider `adt_const_params` for intrinsics


nazar-pc:
Adjacent to https://github.com/Rust-GPU/rust-gpu/issues/605, I'm wondering whether it'd be acceptable to use `adt_const_params` for intrinsics like atomics.

Then this signature:
```rust
fn atomic_i_increment<I: Integer, const SCOPE: u32, const SEMANTICS: u32>(
    ptr: &mut I,
) -> I
```

Will change to something like this:
```rust
fn atomic_i_increment<I: Integer, const SCOPE: Scope, const SEMANTICS: Semantics>(
    ptr: &mut I,
) -> I
```

And usage will be something like this:
```rust
atomic_i_increment::<_, { Scope::QueueFamily }, { Semantics::NONE }>(..)
```

Firestar99:
I wish, but *requiring* nightly-only features breaks building spirv-std on stable, which breaks building it on docs.rs and using it as a dep within your CPU crates that may compile on stable. I'd be fine with having a feature in spirv-std that opts into using some nightly-only feature, but doing so here doesn't really seem possible without hiding the intrinsics completely on non-spirv targets, which will cause a bunch of problems if you plan your CPU crate to depend on your shader crate for type sharing.

nazar-pc:
It would break using it on stable, but it will not break building it on docs.rs since docs.rs is already building things with nightly compiler.

I'm wondering if it even makes sense to expose those intrinsics on non-SPIR-V targets. If such restriction is added (now that it is possible to lint SPIR-V target with Clippy properly directly, etc.), `spirv-std` can expose or not expose things at compile time.

Thus nightly features will be enabled on nightly compiler (that compiling for SPIR-V target will require in the foreseeable future anyway) and they will not be on stable compiler.

---

Also regular standard library has some really weird things going on with intriniscs on other platforms, I suspect for backwards compatibility. For example, [`_mm_aeskeygenassist_si128`](https://doc.rust-lang.org/stable/core/arch/x86/fn._mm_aeskeygenassist_si128.html) can be called both as `_mm_aeskeygenassist_si128(reg, IMM8)` and `_mm_aeskeygenassist_si128::<IMM8>(reg)` (and IntelliJ Rust plugin seems to be quite confused about it).

[...]

PR #635, opened 2026-08-04
Use `adt_const_params` for `Scope` and `Semantics` const generics

### LINKED ISSUE #414: Make `Scope` and `Semantics` normal arguments rather than const generics
The usage of atomics is quite awkward:
```rust
let position_in_bucket = unsafe {
    atomic_i_add::<_, { Scope::QueueFamily as u32 }, { Semantics::NONE.bits() }>(
        bucket_count,
        1,
    )
};
```

Not only one needs to know what each `u32` is, it is not even consistent with the need to cast in one case and calling `.bits()` in another.

Similarly, barriers are awkward too:
```rust
control_barrier::<
    { Scope::Workgroup as u32 },
    { Scope::Workgroup as u32 },
    { Semantics::NONE.bits() },
>();
```

It is even worse when memory semantics is involved (though many of such cases have helper methods):
```rust
control_barrier::<
    { Scope::Workgroup as u32 },
    { Scope::Workgroup as u32 },
    {
        Semantics::WORKGROUP_MEMORY.bits() | Semantics::ACQUIRE_RELEASE.bits()
    },
>();
```

Subjectively, three `u32`s that need to be composed in a very particular way are quite ugly.

Consider changing API to use normal types for these arguments, so we can all enjoy autocomplete in IDE, better formatting and implement ability to combine semantics variants, like this:
```rust
let position_in_bucket = unsafe {
    atomic_i_add(
        bucket_count,
        1,
        Scope::QueueFamily,
        Semantics::None,
    )
};

control_barrier(
    Scope::Workgroup,
    Scope::Workgroup,
    Semantics::WorkgroupMemory | Semantics::AcquireRelease,
);
```

The result of `Semantics::WorkgroupMemory | Semantics::AcquireRelease` can be `Semantics::Combined(u32)` for example, with `Debug` implementation overridden to render human-readable set of enabled flags.

Alternatively, a builder pattern could be used for `Semantics` with `const fn` to have guarantees of sane composition in case some invariants fundamentally do not make sense.

At some point non-integer const generics will become available, but even then I see no benefit from const generics for this particular use case.

nazar-pc:
This implements https://github.com/Rust-GPU/rust-gpu/issues/607 and resolves https://github.com/Rust-GPU/rust-gpu/issues/414 using `adt_const_params`.

Implementation has been on nightly for a long time without major known issues AFAIK. docs.rs already compiles docs with nightly toolchain, so should not be a problem there either.

This makes developer experience better with improved type safety. As a bonus ordering validity is now checked at compile time.

`|` between memory semantics options is technically possible with even more nightly features, but didn't seem worth it for now.

The changes are essentially mechanical.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -380,10 +380,11 @@ jobs:
         run: cargo fmt --check --all --manifest-path tests/difftests/tests/Cargo.toml
       - name: Check docs are valid
         run: RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps
-      - name: Check docs for `spirv-std` and `spirv-builder` on stable (for docs.rs)
-        run: |
-          RUSTDOCFLAGS=-Dwarnings cargo +stable doc --no-deps -p spirv-std
-          RUSTDOCFLAGS=-Dwarnings cargo +stable doc --no-deps -p spirv-builder --no-default-features
+      # TODO: Stable is not supported due to use of nightly features
+      # - name: Check docs for `spirv-std` and `spirv-builder` on stable (for docs.rs)
+      #   run: |
+      #     RUSTDOCFLAGS=-Dwarnings cargo +stable doc --no-deps -p spirv-std
+      #     RUSTDOCFLAGS=-Dwarnings cargo +stable doc --no-deps -p spirv-builder --no-default-features
       - name: cargo clippy
         run: cargo clippy --all-targets -- -D warnings
       - name: custom lints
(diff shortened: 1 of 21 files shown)


Firestar99 (commented):
Not considering this patch requiring a nightly feature, I think the patch looks good. I squashed in a small cleanup to deduplicate
```patch
-    const {
-        assert!(
-            SEMANTICS.has_valid_ordering(),
-            "at most one memory-ordering flag (ACQUIRE, RELEASE, ACQUIRE_RELEASE or \
-             SEQUENTIALLY_CONST) may be set"
-        );
-    }
-
+    const { SEMANTICS.assert_valid() }
```

~~Give me an hour or two, I have an idea I want to toy around with that may solve our stable vs nightly feature problem...~~

See counterproposal in PR https://github.com/Rust-GPU/rust-gpu/pull/636

Firestar99:
Superseded by https://github.com/Rust-GPU/rust-gpu/pull/636

PR #636, opened 2026-08-04
`spirv-std-nightly` crate to enable `adt_const_params` use

### LINKED ISSUE #414: Make `Scope` and `Semantics` normal arguments rather than const generics

### LINKED ISSUE #607: Consider `adt_const_params` for intrinsics

Firestar99:
### Counterproposal to @nazar-pc's PR https://github.com/Rust-GPU/rust-gpu/pull/635

PR https://github.com/Rust-GPU/rust-gpu/pull/635 is modifying `spirv-std` to use `adt_const_params`, which would break for anyone using a CPU-GPU shared crate that depends on `spirv-std` where the CPU target is compiled with stable. It would also be a significant breaking change, with any crate using those functions needing to be manually migrated.

This PR introduces the new crate `spirv-std-nightly`, which reexports the entirety of `spirv-std`, but replaces individual functions with const-generic replacements. It should thus be equivalent to PR https://github.com/Rust-GPU/rust-gpu/pull/635 but not break any existing setups. Instead, it's a drop-in replacement with a simple `use spirv_std_nightly as spirv_std` in any crate. (Though any crate doing so will obviously no longer compile on stable.) You can also mix and match both `spirv-std` crates and it should work just fine.

**Note on publishing:** Our release-plz setup can only update existing crates, I'd need to claim the name and do some setup before we release. But I first want to know if we're claiming that name or if we want to rename the crate beforehand.

close https://github.com/Rust-GPU/rust-gpu/issues/607
close https://github.com/Rust-GPU/rust-gpu/issues/414

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,6 +612,7 @@ name = "compiletests-deps-helper"
 version = "0.0.0"
 dependencies = [
  "spirv-std",
+ "spirv-std-nightly",
 ]
 
 [[package]]
@@ -3678,6 +3679,14 @@ dependencies = [
  "quote",
  "spirv-std-types",
  "syn 3.0.3",
+]
+
+[[package]]
+name = "spirv-std-nightly"
+version = "0.10.0-alpha.1"
+dependencies = [
+ "spirv-std",
+ "spirv-std-macros",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ members = [
     "crates/rustc_codegen_spirv-types",
     "crates/spirv-builder",
     "crates/spirv-std",
+    "crates/spirv-std/nightly",
     "crates/spirv-std/shared",
     "crates/spirv-std/macros",
     "crates/cargo-gpu",
@@ -57,6 +58,7 @@ unexpected_cfgs = { level = "allow", check-cfg = ['cfg(target_arch, values("spir
 [workspace.dependencies]
 spirv-builder = { path = "./crates/spirv-builder", version = "=0.10.0-alpha.1", default-features = false }
 spirv-std = { path = "./crates/spirv-std", version = "=0.10.0-alpha.1" }
+spirv-std-nightly = { path = "./crates/spirv-std/nightly", version = "=0.10.0-alpha.1" }
 spirv-std-types = { path = "./crates/spirv-std/shared", version = "=0.10.0-alpha.1" }
 spirv-std-macros = { path = "./crates/spirv-std/macros", version = "=0.10.0-alpha.1" }
 spirv-tools = { version = "0.13.0", default-features = false }
--- a/crates/spirv-std/Cargo.toml
+++ b/crates/spirv-std/Cargo.toml
@@ -37,8 +37,9 @@ glam_0_30 = { package = "glam", version = "0.30.8", optional = true, default-fea
 
 [features]
 default = ["glam_0_33"]
-bytemuck = ["dep:bytemuck", "bitflags/bytemuck", "glam_0_30?/bytemuck", "glam_0_31?/bytemuck", "glam_0_32?/bytemuck", "glam_0_33?/bytemuck"]
+bytemuck = ["dep:bytemuck", "glam_0_30?/bytemuck", "glam_0_31?/bytemuck", "glam_0_32?/bytemuck", "glam_0_33?/bytemuck"]
 glam_0_33 = ["dep:glam_0_33"]
 glam_0_32 = ["dep:glam_0_32"]
 glam_0_31 = ["dep:glam_0_31"]
 glam_0_30 = ["dep:glam_0_30"]
+adt_const_params = []
--- a/crates/spirv-std/nightly/Cargo.toml
+++ b/crates/spirv-std/nightly/Cargo.toml
@@ -11,7 +11,7 @@ repository.workspace = true
 workspace = true
 
 [dependencies]
-spirv-std.workspace = true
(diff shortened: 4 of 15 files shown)


nazar-pc:
That is a creative workaround, but I guess better than not supporting stable at all

nazar-pc:
The biggest question I have about this is the name. Const generics is one thing, but maybe there will be other nightly features that would be beneficial, then current naming will not scale. Maybe `-nightly` or `-experimental` suffix would be more fitting?

LegNeato:
Can't we just gate code to nightlies with the feature we need? I think serde does something similar...they turn on and off code on different channels and different versions

nazar-pc:
Possible, but then these APIs will have to go into the submodule or something. Right now it is more like "this is what some future version of spirv-std will look like".

Firestar99:
**EDIT:** fixed

Rebased it on main and renamed the crate to `spirv-std-nightly`, only to notice it broke. 
(old functioning rev ecc6db6ffc49c47491c69f0dbbf598d58b72102f in case you need it)

The bitflags v2 update in https://github.com/Rust-GPU/rust-gpu/pull/637 broke this PR. In bitflags v1 we had `Semantics(u32)` but bitflags v2 does `Semantics(SemanticsBits(u32))` and our derive is only attached to the outer one. Bytemuck's traits solve this via the `bitflags/bytemuck` feature deriving Pod on the inner type, which is why you also need to derive it on the outer type to get it functioning. So we'd need to convince bitflags to have a `nightly-adt-const-params` feature of some sort to support this.

nazar-pc:
Hm... how important is it to depend on `bitflags` in the first place? Looks like we're fighting their design decisions and it might be easier to simply write necessary implementation manually in the way `rust-gpu` needs it. Macro expansion could be the starting point.

Firestar99:
Bitflags is *the* standard crate for bitflags, wouldn't really want to replace it. Also I think there's a decent chance they'd take a PR adding that behind a feature gate.

nazar-pc:
That is just perfect derive issue. Simply replace derive with `impl core::marker::ConstParamTy_ for T { }` and it'll work. It is unfortunate that the trait name is for some reason different, but that is exactly what the nightly derive expands to today.

UPD: Hm... maybe not. I really don't like fight external crates for simple things :confused:

Firestar99:
Found a workaround: The `bitflags!` macro supports not just `pub struct Semantics` but also `impl Semantics` to allow you to define the struct yourself.

Firestar99:
@nazar-pc Just rebased this PR. You think it's fine the way it is, and supersedes your PR https://github.com/Rust-GPU/rust-gpu/pull/635? Then I'd merge it as is before I'll do the next release.

nazar-pc:
Yes, looks like it

nazar-pc:
BTW, look was just opened: https://github.com/rust-lang/rust/issues/163261

Firestar99:
So the lint CI is failing cause `cargo doc` is ICE-ing on the const generics in the `spirv-std-nightly` crate. It's fixed in newer nightlies, eg. rebasing on https://github.com/Rust-GPU/rust-gpu/pull/630 makes it pass.

PR #638, opened 2026-08-05
Fix bytemuck feature after bitflags update

nazar-pc:
Docs were not checked in CI, but failed somehow after merging to `main`: https://github.com/Rust-GPU/rust-gpu/actions/runs/30995586739/job/92271767273.

Copy/Clone derived unconditionally, despite only being needed for `bytemuck`. Feel free to update if that is not desirable (I think it is though).

### DIFF
--- a/crates/spirv-std/src/memory.rs
+++ b/crates/spirv-std/src/memory.rs
@@ -30,6 +30,7 @@ bitflags::bitflags! {
     /// Memory semantics to determine how some operations should function - used when calling such
     /// configurable operations.
     #[repr(transparent)]
+    #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
     #[cfg_attr(feature = "bytemuck", derive(bytemuck::Zeroable, bytemuck::Pod))]
     pub struct Semantics: u32 {
         /// No memory semantics.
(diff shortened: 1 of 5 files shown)


nazar-pc:
Will have a bit better version in a moment

nazar-pc:
Okay, better solution is to simply enable `bytemuck` feature on `bitflags` itself. But consider adding docs checks to PRs as well so this can be caught earlier in the future.

Firestar99:
I'm more concerned about [this breaking change from bitflags v1 -> v2](https://github.com/bitflags/bitflags/blob/c3f8d31d59f3ba8c05833fb322ecab8a2cb88060/CHANGELOG.md?plain=1#L352-L358):

> ### ⚠️ Traits
> Generated flags types now derive fewer traits. If you need to maintain backwards compatibility, you can derive the following yourself:
> ```rust
> #[derive(PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Clone, Copy)]
> ```

We should probably derive all that, I'll add it along with a check for the bytemuck feature.

nazar-pc:
Makes sense

crates/spirv-std/src/memory.rs
```diff
@@ -30,7 +30,6 @@ bitflags::bitflags! {
     /// Memory semantics to determine how some operations should function - used when calling such
     /// configurable operations.
     #[repr(transparent)]
-    #[cfg_attr(feature = "bytemuck", derive(bytemuck::Zeroable, bytemuck::Pod))]
```
Firestar99, line 33: FYI Your Semantics doesn't actually impl Pod, you need to both enable the `bitflags/bytemuck` feature *and* derive it.

Firestar99 (approved):

crates/spirv-std/src/memory.rs
nazar-pc: That is not exactly what I understood by "Support deriving bytemuck traits on generated flags types" in their documentation, but fine I guess.

Firestar99: Which is why I've added a test to verify it at compile time :D

PR #640, opened 2026-08-07
support scalar pairs properly

Firestar99:
Supersedes https://github.com/Rust-GPU/rust-gpu/pull/381, see discussions over there.

Removes the "abi readjustment" aka hack to treat ScalarPairs as Scalars and adds proper support for ScalarPairs where necessary. Likely required to update beyond `nightly-2025-07-19`, see discussion in PR https://github.com/Rust-GPU/rust-gpu/pull/630#issuecomment-5214414526.

Reviewers: the fmt args decompiler adjustments are AI written, I have no idea what's going on in that module.

### DIFF
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -92,15 +92,6 @@ pub(crate) fn provide(providers: &mut Providers) {
             // <https://github.com/rust-lang/rust/commit/eaaa03faf77b157907894a4207d8378ecaec7b45>
             arg.make_direct_deprecated();
 
-            // FIXME(eddyb) detect `#[rust_gpu::vector::v1]` more specifically,
-            // to avoid affecting anything should actually be passed as a pair.
-            if let PassMode::Pair(..) = arg.mode {
-                // HACK(eddyb) this avoids breaking e.g. `&[T]` pairs.
-                if let TyKind::Adt(..) = arg.layout.ty.kind() {
-                    arg.mode = PassMode::Direct(ArgAttributes::new());
-                }
-            }
-
             // Avoid pointlessly passing ZSTs, just like the official Rust ABI.
             if arg.layout.is_zst() {
                 arg.mode = PassMode::Ignore;
@@ -490,7 +481,7 @@ pub fn scalar_pair_element_backend_type<'tcx>(
     ty: TyAndLayout<'tcx>,
     index: usize,
 ) -> Word {
-    let [a, b] = match ty.layout.backend_repr() {
+    let [a, b] = match ty.backend_repr {
         BackendRepr::ScalarPair(a, b) => [a, b],
         other => span_bug!(
             span,
--- a/crates/rustc_codegen_spirv/src/builder/format_args_decompiler.rs
+++ b/crates/rustc_codegen_spirv/src/builder/format_args_decompiler.rs
@@ -557,16 +557,19 @@ impl<'tcx> DecodedFormatArgs<'tcx> {
             if let Some((template_id, template_ty_id, rt_args_ptr_id, rt_args_ptr_ty_id)) =
                 split_fmt_args
             {
-                let ctor = if let (Some(template_len), Some(rt_args_count)) = (
+                if let (Some(template_len), Some(rt_args_count)) = (
                     const_ptr_to_composite_len(template_id)
                         .or_else(|| array_len_from_ptr_type(template_ty_id)),
                     const_ptr_to_composite_len(rt_args_ptr_id)
                         .or_else(|| array_len_from_ptr_type(rt_args_ptr_ty_id)),
                 ) {
-                    FmtArgsCtor::NewTemplate {
-                        template_len,
-                        rt_args_count,
-                    }
+                    (
+                        FmtArgsCtor::NewTemplate {
+                            template_len,
+                            rt_args_count,
+                        },
+                        SmallVec::<[Word; 8]>::from_slice(&[template_id, rt_args_ptr_id]),
+                    )
                 } else if let Some(&[Inst::Call(_, callee_id, ref call_args)]) =
                     try_rev_take(-1).as_deref()
                     && call_args.len() == 2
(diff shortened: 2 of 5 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -693,6 +673,34 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                        .align_to(b.primitive().align(self).abi);
+
+                    let elem0_ty = self.scalar_pair_element_backend_type(layout, 0, false);
+                    let elem1_ty = self.scalar_pair_element_backend_type(layout, 1, false);
+
+                    let base_ptr = value_ptr.unwrap();
+                    let ptr1 = bx.inbounds_ptradd(base_ptr, self.const_usize(b_offset.bytes()));
+
+                    let v0 = bx.load(elem0_ty, base_ptr, layout.align.abi);
+                    let v1 = bx.load(elem1_ty, ptr1, layout.align.restrict_for_offset(b_offset));
```
eddyb, line 699: It might be possible to replace this with a call to `load_operand`.

crates/rustc_codegen_spirv/src/builder/format_args_decompiler.rs
```diff
@@ -557,35 +557,60 @@ impl<'tcx> DecodedFormatArgs<'tcx> {
...
+                    // two scalar values passed to the panic entry-point.
+                    //
+                    // The constructor's own arguments (i.e. `pieces`/`template`
+                    // and the `rt::Argument` slice pointers) still carry the
+                    // recoverable const data, so use those, like the aggregate
+                    // (non-split) `Call`+`extract`+`insert` case does below.
+                    let call_args_storage = call_args.iter().copied().collect();
+                    // Consume the matched call + both `OpCompositeExtract`s.
+                    try_rev_take(3).unwrap();
+                    (lookup_fmt_args_ctor(callee_id)?, call_args_storage)
```
eddyb, line 607: The only reason I don't like this has to do with it feeling misplaced, likely a consequence of the `split_fmt_args` changes from months ago, which I might eventually revisit (and shouldn't block this PR).

eddyb (changes requested):
LGTM, modulo trying to simplify `entry.rs` using the higher-level helpers that already exist.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
Firestar99: Pushed a cleanup, definitely a lot cleaner

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -647,8 +626,8 @@ impl<'tcx> CodegenCx<'tcx> {
                 }
             }
 
-            let value_len = if is_pair {
-                // We've already emitted an error, fill in a placeholder value
+            let value_len = if is_pair && is_unsized {
+                // For wide references (e.g., slices), the second component is a length.
```
eddyb, line 630: Losing the old comment makes this look unsound (`undef` is only being used because this situation is illegal).

Also, the variables make this additionally confusing, because `is_...` refer to two different aspects.

`is_pair` is "pass as pair" or "expects pair", whereas `is_unsized` should probably be renamed to `pointee_is_unsized` (it only really makes sense with refs, i.e. "ref to unsized").

Not to mention that the condition for "`value_len` is needed" is `ty::Ref(..)`+`is_pair`, with `is_unsized` being related but not necessarily the same thing.

Also, `self.type_isize()` isn't the right thing to use in the general case, and could lead to panics elsewhere if the unsized type is not a slice (though today that's only `&dyn Trait`, so I'm not sure it's testable at all).

---

Minimal fix: include the error justification in a comment, use `scalar_pair_element_backend_type` w/ `entry_arg_abi.layout`, instead of `type_isize`, maybe gate by `ty::Ref` instead of `is_unsized`.

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -490,7 +481,7 @@ pub fn scalar_pair_element_backend_type<'tcx>(
     ty: TyAndLayout<'tcx>,
```
eddyb, line 481: This reminds me that a longstanding refactor has been replacing all `ty: TyAndLayout` to `layout: TyAndLayout` (because it derefs to the layout, but not the type, so e.g. `layout.ty` and `layout.size`, instead of `ty.ty` and `ty.size` or the redundant `ty.layout.size`).

IIRC I regretted not naming it `LayoutWithTy` or similar, to make clearer the intent.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -693,6 +672,22 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                PassMode::Pair(..) => {
+                    // Load both elements of the scalar pair from the input variable.
+                    assert_eq!(storage_class, Ok(StorageClass::Input));
+                    let OperandRef {
+                        val: OperandValue::Pair(v0, v1),
+                        ..
+                    } = bx.load_operand(PlaceRef::new_sized(
+                        value_ptr.unwrap(),
+                        entry_arg_abi.layout,
+                    ))
```
eddyb, line 684: I think `load_operand` should be used uniformly (in this `else {...}`), when `storage_class` is `Ok`, and then pattern-matching on `(operand.val, entry_arg_abi.mode)`.

This would also force `SpecConstant`s to require `PassMode::Direct` (right now they *technically* allow `PassMode::Ignore`, not sure if anything else catches that).

eddyb (changes requested):

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
Firestar99: This optimization doesn't work: `glam` vecs are now `BackendRepr::Memory` since we removed the ABI patching. Meaning `load_operand` will return an `OperandValue::Ref { .. }` instead of actually `bx.load()`-ing it and returning a `OperandValue::Immediate`, as it does currently.

eddyb (approved):

PR #641, opened 2026-08-10
xtask set-toolchain

Firestar99:
I just couldn't be bothered anymore pasting around the nightly version and it's commit hash every time I want to try a new nightly. (Excuse the xtask crate being a bit messy, it was copied over from cargo-gpu and never cleaned up afterwards.)

Usage:
```bash
# update to today's nightly
cargo xtask set-toolchain
# update to specified nightly
cargo xtask set-toolchain nightly-2026-08-10
cargo xtask set-toolchain 2026-08-10
```

Making this a separate PR so anyone updating toolchains in the future is aware of it @eddyb

@LegNeato I know your [toolchain upgrade action](https://github.com/Rust-GPU/rust-gpu/blob/main/.github/workflows/rust-toolchain-upgrade.yml) uses [this bash script](https://github.com/Rust-GPU/rust-gpu/blob/main/.github/check-rust-toolchain-upgrade.sh) to update the toolchain. But it also contains additional checks specific for the github action, so I'm not sure how these two tools should relate.

### DIFF
--- a/crates/xtask/src/toolchain.rs
+++ b/crates/xtask/src/toolchain.rs
@@ -1,0 +1,87 @@
+use anyhow::Context;
+use chrono::{Datelike, Utc};
+use std::env;
+use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
+
+/// Set the toolchain of rust-gpu to some nightly
+#[derive(Debug, clap::Parser)]
+pub struct SetToolchain {
+    /// nightly version to set the toolchain to
+    ///
+    /// accepts `nightly-1234-56-78` or `1234-56-78`, defaults to today's nightly
+    version: Option<String>,
+}
+
+impl SetToolchain {
+    pub fn run(&self) -> anyhow::Result<()> {
+        // figure out version and commit-hash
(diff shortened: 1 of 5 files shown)


crates/xtask/src/toolchain.rs
```diff
@@ -0,0 +1,81 @@
...
+        update_file(&root.join("crates/rustc_codegen_spirv/build.rs"))?;
+
+        // try to update `jj desc`
+        Command::new("jj")
+            .args(["desc", "-m"])
+            .arg(format!("update to {version}"))
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()
+            .ok();
```
Firestar99: I would love to only update the desc (jj commit message) if the desc is empty, but I haven't found a way to do that. (If you don't have jj cli, it'll skip this entirely)

Firestar99: Found a solution, see https://github.com/jj-vcs/jj/blob/main/docs/templates.md#examples

eddyb (approved):

PR #642, opened 2026-08-10
compiletest `-Zbuild-dir-new-layout` support

Firestar99:
See https://blog.rust-lang.org/2026/03/13/call-for-testing-build-dir-layout-v2/

Will be enabled by default starting `nightly-2026-07-31`.

Only affects our compiletest infra, since `spirv-builder` uses `cargo metadata` to query target dir location to only append `./target/spirv-builder`.


eddyb (approved):
This mechanism was already pretty hacky, whatever it takes to keep it working seems fine to me (long-term the new build dir layout might allow some interesting caching AIUI, unsure how that will interact with anything else we do).

